PR #37, opened 2015-03-20
Make Adapters::Memory#clear conform to expectations

aroben:
An adapter's `#clear` method is supposed to return true according to the specs.

/cc @rsanheim


rsanheim:
:+1: 

I actually would like to change the specs to just expect truthiness, and not an explicit true, but that is a PR for another day.

PR #38, opened 2015-03-25
Speed up Subscriber#strip_trailing_question_mark

aroben:
I noticed this method while profiling `Feature#enabled?`. Turns out we can speed it up without hurting readability by using `chomp` instead of `gsub`.

    $ cat bench.rb
    require 'benchmark'

    Benchmark.bm do |x|
      x.report("gsub ") { 100000.times { %w[enabled enabled?].each { |str| str.to_s.gsub(/\?$/, '') } } }
      x.report("chomp") { 100000.times { %w[enabled enabled?].each { |str| str.to_s.chomp('?') } } }
    end
    $ ruby bench.rb
           user     system      total        real
    gsub   0.270000   0.010000   0.280000 (  0.273812)
    chomp  0.070000   0.000000   0.070000 (  0.071246)

I tried a lot of other variations (`\z` instead of `$`, `gsub!`, `sub`, `sub!`, `slice`, `slice!`, `chomp!`) and this was the fastest.

I doubt this will have a big effect on Flipper's performance overall, but there doesn't seem to be much downside here.

/cc @jnunemaker @rsanheim

### DIFF
--- a/lib/flipper/instrumentation/subscriber.rb
+++ b/lib/flipper/instrumentation/subscriber.rb
@@ -95,7 +95,7 @@ module Flipper
 
       # Private
       def strip_trailing_question_mark(operation)
-        operation.to_s.gsub(/\?$/, '')
+        operation.to_s.chomp('?')
       end
     end
   end


rsanheim:
Related failure? https://travis-ci.org/jnunemaker/flipper/jobs/55820688

aroben:
That spec fails on master for me, so I don't think it's related.

aroben:
Looks like 004edeb0a884da0f18480379d2392db12ff86e52 broke that test.

jnunemaker:
Weird, that spec was failing locally for me so I changed it and it passed. Wonder why it is failing in CI now.

PR #39, opened 2015-03-26
Groups API

jnunemaker:
Renames internal groups to groups_registry and adds public groups and group_names methods.

cc @aroben @rsanheim


Gemfile
```diff
@@ -4,13 +4,6 @@ gemspec
 gem 'rake'
 gem 'metriks', :require => false
 gem 'statsd-ruby', :require => false
-gem 'rspec'
+gem 'rspec', '~> 2.14'
 gem 'rack-test'
 gem 'activesupport', :require => false
-
-group(:guard) do
```
rsanheim: Do you not use it anymore?

Im still a fan.

lib/flipper.rb
```diff
@@ -24,27 +24,39 @@ def self.new(adapter, options = {})
...
     group = Types::Group.new(name, &block)
-    groups.add(group.name, group)
+    groups_registry.add(group.name, group)
     group
   rescue Registry::DuplicateKey
     raise DuplicateGroup, %Q{Group #{name.inspect} has already been registered}
   end
 
+  # Public: Returns an array of registered Types::Group instances.
+  def self.groups
```
rsanheim: :+1:

lib/flipper.rb
```diff
@@ -55,19 +67,19 @@ def self.group_exists?(name)
   # Returns the Flipper::Group if group registered.
   # Raises Flipper::GroupNotRegistered if group is not registered.
   def self.group(name)
-    groups.get(name)
+    groups_registry.get(name)
   rescue Flipper::Registry::KeyNotFound => e
     raise GroupNotRegistered, "Group #{e.key.inspect} has not been registered"
   end
 
-  # Internal: Registry of all groups.
```
rsanheim: Maybe worth making the internal methods `private`?

rsanheim: Oh, that would break the specs tho...

Gemfile
jnunemaker: I haven't been. We can add it back in later if you are doing flipper dev and want it.

lib/flipper.rb
jnunemaker: I always hate this part. Some stuff needs to be tested, but I don't want to support it publicly. I'm open to suggestions.

PR #40, opened 2015-03-26
Remove gate instrumentation

jnunemaker:
This PR removes gate instrumentation. It adds some overhead due to happening per gate with multiple gates per feature and doesn't provide a ton of value currently. We can add something back when we have a better idea of what would be useful. 

cc @aroben @rsanheim

### DIFF
--- a/lib/flipper/feature.rb
+++ b/lib/flipper/feature.rb
@@ -175,7 +175,7 @@ module Flipper
 
     # Private
     def conditional_gates(gate_values)
-      @conditional_gates ||= non_boolean_gates.select { |gate|
+      non_boolean_gates.select { |gate|
         value = gate_values[gate.key]
         gate.enabled?(value)
       }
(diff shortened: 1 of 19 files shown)


aroben:
As we discussed elsewhere, just doing `ActiveSupport::Notifications.subscribe /(?<!gate_operation)\.flipper$/` gives most of the same performance benefits as removing the instrumentation as was done in this PR. So if the instrumentation is still useful (e.g., for profiling Flipper itself) I'm OK with it remaining.

jnunemaker:
The only thing the instrumentation is useful for is determining which gate is slow if a feature enabled check is slow. If the issue was just the subscription, I'll close this and leave it.

aroben:
I found that 90% of the time spent was in the subscription. The remaining 10% seems OK to me since the instrumentation can be useful in some circumstances.

PR #44, opened 2015-04-03
Rename percentage of random to percentage of time

### LINKED ISSUE #43: Rename percentage of random to percentage of time
@rsanheim @aroben @dewski any thoughts on this? I'm getting sick of random. Percentage of time seems more accurate.

jnunemaker:
Title says it all. I'm finding that percentage of time just flows better and is more immediately obvious as to what it does. Enable this feature for a percentage of the time. It isn't really based on time, which is why I named it random originally, but I'm starting to not care about that as much.

Thoughts @rsanheim @dewski @aroben?

Fixes #43


jnunemaker:
Since I'm maintaining this long term, I'm going to go with this. If anyone does have any issues let me know and we can discuss more.

PR #45, opened 2015-04-03
Feature and Gate Tweaks

jnunemaker:
Mostly internal changes but a few public ones.

* moves typecasting to module and out of gate values
* moves gate instrumentation to feature, will make it easier to allow tweaking gates per feature in the future
* adds Feature#disabled_groups
* Flipper.groups and .group_names now return a set to be consistent with intention and the rest of flipper. Perhaps I should change everything to arrays and drop the Set idea as it is kind of just annoying in ruby.

fyi @rsanheim @dewski @aroben


lib/flipper/feature.rb
```diff
@@ -219,44 +224,64 @@ def gate_values
...
+    #
+    # Returns Integer greater than or equal to 0 and less than or equal to 100.
     def percentage_of_time_value
       gate_values.percentage_of_time
     end
 
     # Public: Returns the string representation of the feature.
     def to_s
-      @to_s ||= name.to_s
+      name.to_s
```
rsanheim: :+1:

rsanheim:
> Perhaps I should change everything to arrays and drop the Set idea as it is kind of just annoying in ruby.

Agree with this, sets just don't feel needed with what we are doing here.

PR #47, opened 2015-04-08
Use the instrumented block's payload, not the original payload

aroben:
Some instrumentation services don't pass the original payload straight through to the block (e.g., they might copy it and add their own information to it). With those instrumenters it's important to use the payload passed to the block rather than the original payload or information added by the block might get lost.

/cc @jnunemaker @rsanheim

### DIFF
--- a/lib/flipper/instrumenters/memory.rb
+++ b/lib/flipper/instrumenters/memory.rb
@@ -12,6 +12,11 @@ module Flipper
       end
 
       def instrument(name, payload = {})
+        # Copy the payload to guard against later modifications to it, and to
+        # ensure that all instrumentation code uses the payload passed to the
+        # block rather than the one passed to #instrument.
+        payload = payload.dup
+
         result = if block_given?
           yield payload
         else
--- a/lib/flipper/feature.rb
+++ b/lib/flipper/feature.rb
@@ -354,7 +354,7 @@ module Flipper
         :thing => thing,
       }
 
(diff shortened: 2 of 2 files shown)


lib/flipper/instrumenters/memory.rb
```diff
@@ -12,6 +12,11 @@ def initialize
       end
 
       def instrument(name, payload = {})
+        # Copy the payload to guard against later modifications to it, and to
+        # ensure that all instrumentation code uses the payload passed to the
+        # block rather than the one passed to #instrument.
+        payload = payload.dup
```
aroben: This causes a number of Feature specs to fail (without the fix to feature.rb).

rsanheim:
Related rails code: https://github.com/rails/rails/commit/7f4d8e3fbd704cf1005a46309e3795192b114013

rsanheim:
Going to go ahead and merge this, and confirm it fixes the application issue we are seeing.

lib/flipper/instrumenters/memory.rb
jnunemaker: Hmm. Seems like a lot of new objects to create duping every payload.

aroben: This is only for tests right?

jnunemaker: Sorry. Sleep deprived.

Issue #50, opened 2015-04-09
Would like support for custom/dynamic groups


aroben:
It would be nice to have a way to define a custom group type that corresponds to, e.g., a GitHub team. Something like this:

```ruby
class GitHubTeam
  def flipper_id
    "GitHubTeam:#{id}"
  end

  def self.find_by_flipper_id(flipper_id)
    find flipper_id.split(":")[1]
  end
end

Flipper.register_custom_group(:github_team) do |group_id, actor|
  GitHubTeam.find_by_flipper_id(group_id).include? actor
end

$flipper[:my_feature].enable $flipper.custom_group(:github_team, GitHubTeam.find("rails/owners"))
$flipper[:my_feature].enabled? some_user #=> true because some_user is a member of @rails/owners
```

Basically the only difference from normal groups is that the group's ID is passed to the block as well as the actor, so we can look up the group (team) at runtime.

/cc @jnunemaker @rsanheim

aroben:
The adapter would need to store both the group name (`:github_team`) and the group object's `flipper_id`. We'd either need to combine these into a single string somehow, or adapters will have to be changed to be able to store multiple values for a single set member.

aroben:
Combining into a single string seems pretty simple. We'd just have to pick a separator character (perhaps  ASCII's `\x1E` record separator) and then disallow that character from appearing in the group name. I'll take a crack at this.

jnunemaker:
I don't have time to look at this right unfortunately but it doesn't feel like the right abstraction. I'd be more inclined to have a name to group mapping and just iterate them and create normal groups. 


{
rails_owners: "rails/owners",
}.each do |flipper_group, group|
  Flipper.register...
end

Or I would just define a flipper group that was enabled for a bunch of teams. 

Does any of that make sense? Custom group as you have above doesn't feel different enough from group and my fear is it will confuse people down the road as to which should be used.

aroben:
The approach I took in #51 is just to support passing an extra parameter to group blocks. Another possibility I was considering is having a way to register custom logic to handle groups that haven't been registered:

```ruby
Flipper.group_fallback_handler = do |group_name, thing|
  Team.find_by_name(group_name).include? thing
end
$flipper[:my_feature].enable_group "rails/owners"
$flipper[:my_feature].enabled? some_user
```

Since `rails/owners` hasn't been registered as a group using `Flipper.register`, the fallback handler block will be called, passing "rails/owners" as the `group_name`.

[...]

PR #51, opened 2015-04-10
Allow an extra parameter to be passed to group blocks

aroben:
Methods that took a group name now take an optional block parameter, which is passed through to the group's block. This can be used, e.g., to implement dynamic groups based on a database. Here's how it looks:

```ruby
Flipper.register(:team) do |actor, team_id|
  Team.find(team_id).members.include?(actor)
end

$flipper[:my_feature].enable $flipper.group(:team, Team.by_name("admins").id.to_s)
```

The block parameter must be convertible to a String via `#to_str`.

Fixes https://github.com/jnunemaker/flipper/issues/50

/cc @rsanheim @jnunemaker

### DIFF
--- a/lib/flipper/group.rb
+++ b/lib/flipper/group.rb
@@ -3,7 +3,12 @@ module Flipper
     attr_reader :name
 
     def initialize(name, &block)
-      @name = name.to_sym
+      name = name.to_sym
+      if name.to_s.include?(Types::Group::SEPARATOR)
+        raise ArgumentError.new("#{name.inspect} must not include #{Types::Group::SEPARATOR.inspect} characters")
+      end
+
+      @name = name
       @block = block
     end
 
--- a/lib/flipper/group.rb
+++ b/lib/flipper/group.rb
@@ -1,0 +1,14 @@
+module Flipper
+  class Group
+    attr_reader :name
+
+    def initialize(name, &block)
+      @name = name.to_sym
+      @block = block
+    end
+
+    def match?(*args)
+      @block.call(*args) == true
+    end
+  end
+end
--- a/spec/flipper/feature_spec.rb
+++ b/spec/flipper/feature_spec.rb
@@ -270,9 +270,9 @@ describe Flipper::Feature do
         @preview_features = Flipper.register(:preview_features) { |thing| true }
         @not_enabled = Flipper.register(:not_enabled) { |thing| true }
         @disabled = Flipper.register(:disabled) { |thing| true }
-        subject.enable @staff
-        subject.enable @preview_features
-        subject.disable @disabled
+        subject.enable Flipper::Types::Group.new(:staff)
+        subject.enable Flipper::Types::Group.new(:preview_features)
+        subject.disable Flipper::Types::Group.new(:disabled)
       end
 
       it "returns set of enabled groups" do
@@ -309,9 +309,9 @@ describe Flipper::Feature do
         @preview_features = Flipper.register(:preview_features) { |thing| true }
         @not_enabled = Flipper.register(:not_enabled) { |thing| true }
         @disabled = Flipper.register(:disabled) { |thing| true }
-        subject.enable @staff
-        subject.enable @preview_features
-        subject.disable @disabled
+        subject.enable Flipper::Types::Group.new(:staff)
+        subject.enable Flipper::Types::Group.new(:preview_features)
+        subject.disable Flipper::Types::Group.new(:disabled)
       end
 
       it "returns set of groups that are not enabled" do
@@ -336,9 +336,9 @@ describe Flipper::Feature do
         @preview_features = Flipper.register(:preview_features) { |thing| true }
         @not_enabled = Flipper.register(:not_enabled) { |thing| true }
         @disabled = Flipper.register(:disabled) { |thing| true }
-        subject.enable @staff
-        subject.enable @preview_features
-        subject.disable @disabled
+        subject.enable Flipper::Types::Group.new(:staff)
+        subject.enable Flipper::Types::Group.new(:preview_features)
+        subject.disable Flipper::Types::Group.new(:disabled)
       end
 
       it "returns set of enabled groups" do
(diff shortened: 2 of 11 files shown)


spec/flipper/feature_spec.rb
```diff
@@ -270,9 +270,9 @@
         @preview_features = Flipper.register(:preview_features) { |thing| true }
         @not_enabled = Flipper.register(:not_enabled) { |thing| true }
         @disabled = Flipper.register(:disabled) { |thing| true }
-        subject.enable @staff
-        subject.enable @preview_features
-        subject.disable @disabled
+        subject.enable Flipper::Types::Group.new(:staff)
```
rsanheim: Why did you need these changes?

rsanheim:
This seems like a good solution to me.  It also seems non-invasive to me, and I can't see any areas were existing users of flipper would be hurt by it unless they are depending on the specific class of `Flipper::Types::Group`.

lib/flipper/group.rb
```diff
@@ -0,0 +1,19 @@
+module Flipper
+  class Group
+    attr_reader :name
+
+    def initialize(name, &block)
+      name = name.to_sym
+      if name.to_s.include?(Types::Group::SEPARATOR)
```
rsanheim: Hrm, I guess this could break old clients, if they happen to use this separator in group names.  That would be pretty surprising, though.

spec/flipper/feature_spec.rb
aroben: https://github.com/aroben/flipper/commit/95ca7f43af4327dffb99b20aa0f001010c67f40e made `Flipper.register` return a `Flipper::Group` instead of a `Flipper::Types::Group`. This code was then passing that `Flipper::Group` into `Feature#enable`.

We could handle this case, but it seemed like it would just encourage mistakes to be so lax about the types. I think in most cases client code shouldn't be dealing with `Flipper::Group` objects at all; they're mostly an implementation detail.

lib/flipper/group.rb
aroben: Yup, it would certainly be surprising. And the clients will notice it right away, as soon as `Flipper.register` is called.

aroben:
I guess I should write some docs too.

rsanheim:
This would definitely require a `0.0.7.beta` bump.

I also noticed that there is a `CHANGELOG` that noone has touched since 0.0.6...we should either kill that or automate it somehow so we don't forget it.

jnunemaker:
I just don't think I like the complexity this adds. I feel like the code below, which uses the stock flipper features we have now, is more explicit and simple. 

```ruby
Flipper.register(:admins_team) do |actor|
  Team.by_name("admins").members.include?(actor)
end

flipper[:my_feature].enable flipper.group(:admins_team)
```

I keep looking at the changes and I just don't get the use case driving the block changes and the dynamic group issue/pr. I'm open to change, but this feels like trading explicit simple code for a bit magical code. Thoughts?

aroben:
@jnunemaker The hope is to better support the way that we at GitHub ship new features.

We have a bunch of teams at GitHub and we like to turn new features on for them before turning them on more widely. For example, someone working on a new organization-related feature might turn the feature on for just the `@github/orgs` team. Or someone working on a new pull request feature might turn it on for just the `@github/pull-requests` team. Or someone working on a new billing feature might turn it on for just the `@github/billing` team.

As you said, we could just define groups for each of these teams:

```ruby
%w[orgs pull-requests billing].each do |team|
  Flipper.register("#{team}_team") do |actor|
    Team.by_name(team).members.include?(actor)
  end
end
```

But requiring a code change each time we decide there's a new team that it would be useful to scope a feature to seems heavyweight. The goal of this PR was to make it so these teams could be added dynamically without making a code change/deploy. In our web UI for modifying Flipper features, we'd just have a "enable for team" field that let you type in any GitHub team name.

There is definitely added complexity for supporting this functionality. The hope was to make it super easy for someone to turn on a feature for a team without having to make a commit/deploy.

jnunemaker:
I've always considered using teams at GitHub as a backwards way of doing features. The first backwards thing is that it leads to an explosion of teams which is its own issue. The second backwards part is that most people then use that team for mentions and more. This means when you join the team to try the new feature, you also get a bunch of notifications. Because I've never wanted to receive the notifications, I often don't join the new teams and try out new features. **I think this style of enabling features at GitHub exists simply because we didn't have flipper in the past**.

Now that we have flipper, I would lean toward using teams on github.com for mentions and notifications as we do (to get work done) and use flipper for joining/leaving the use of a new feature.

Let's pretend I was going to move notifications to a new cluster. Instead of making a new team in the github organization like notifications-new-cluster and using those people as guinea pigs, I'd make a feature that was named notifications_new_cluster and I'd tell people to add themselves as an actor or I would add them myself using our flipper interface.

This solves the issue of needing to commit/deploy as the person can simply add themselves in the UI as an actor that is enabled for the feature. It also moves all feature related stuff into flipper completely which means no chance of N+1 database calls for feature checks. The Team.by_name(...).members check requires a database call and would not (without work) be memoized per request as flipper actor checks are.

Any thoughts on that? I realize this is different from how we do things now, but I think it is a good thing, an improvement. Teams are for getting work done and communication and features are for toggling.

aroben:
> The second backwards part is that most people then use that team for mentions and more.

This is definitely a problem (unless you create two teams, one for the people working on the feature and the other for people testing the feature, but then you have an even bigger explosion of teams).

> Now that we have flipper, I would lean toward using teams on github.com for mentions and notifications as we do (to get work done) and use flipper for joining/leaving the use of a new feature.

That seems like a reasonable approach for now. If we find there are groups of people that we typically want to enable features for (e.g., all engineers working on github.com) we can create a group for them.

Thanks for taking the time to think about this and write up your thoughts. I agree we should try to lean into Flipper a little harder before trying to shoehorn our old workflow into Flipper. Maybe eventually we'll decide that team-based shipping is a good idea, but let's see how far we get without it.

jnunemaker:
> we can create a group for them

Sounds good.

> Thanks for taking the time to think about this and write up your thoughts.

No problem. Thanks for proposing a working solution. Makes it easier to think through.

> Maybe eventually we'll decide that team-based shipping is a good idea, but let's see how far we get without it.

Definitely. If we get to a point where we are feeling pain, we can come back to the drawing board.

PR #58, opened 2015-04-24
Ax description

### LINKED ISSUE #56: Ax description
It is kind of terrible. Instead, inspect could use enabled gates or include the gate values. This also makes description available as an actual field that users fill out for each feature.

jnunemaker:
I found it confusing. Sometimes it would say enabled for 15% of actors, 15% of the time which meant 15% of actors and 15% of time but looked like 15% of actors 15% of the time.

Additionally, I’d like to make description available as a user defined thing describing what the feature is used for.

Fixes #56


rsanheim:
:+1:

PR #68, opened 2015-05-05
Always instrument flipper type

jnunemaker:
It was split before based on what method you called and what you passed in. Now it always instruments a flipper type for consistency sake.


jnunemaker:
I also added support for enabling with a symbol as it just made sense.

PR #77, opened 2015-07-15
Add optional env vars to connect with mongodb

alabeduarte:
Hi!

I was struggling to run the specs due my mongodb is running in a specific host, so I did something similar to the Redis adapter spec.

So to run the specs you can execute:
```
BOXEN_MONGODB_HOST=my_mongodb_host bundle exec rspec
```

Thanks!


jnunemaker:
Thank you! Let me know how it works out for you.

PR #87, opened 2015-11-16
Rspec3 syntax

listrophy:
Anyone who's writing their own adapter and using flipper's `shared_examples` to test it are likely to get deprecation warnings.

This PR includes 2 unrelated things... sorry!

**First** (and most importantly), it switches the `should` syntax to `expect`. Nearly all the conversion was done with http://xinminlabs.github.io/synvert/, so there should be less chance for user error. The only manually converted file is the shared_spec file in lib.

**Second**, I got a bit frustrated running specs on a clean repo the first time. Turns out, the mongo adapter waits 60 (or so) seconds (per spec!) before failing if it can't connect to the mongo server. This PR shortens that to 1 second and outputs a helpful message about maybe mongo not running.

### DIFF
--- a/spec/flipper/adapters/mongo_spec.rb
+++ b/spec/flipper/adapters/mongo_spec.rb
@@ -1,4 +1,5 @@
 require 'helper'
+require 'timeout'
 require 'flipper/adapters/mongo'
 require 'flipper/spec/shared_adapter_specs'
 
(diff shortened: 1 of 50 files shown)


spec/flipper/adapters/mongo_spec.rb
```diff
@@ -16,8 +17,19 @@
 
   before do
     begin
-      collection.drop
-    rescue Mongo::Error::OperationFailure
+      Timeout::timeout(1) do
```
jnunemaker: Can we just set the timeout in the mongo client initialization and call it good? A rescue would be good, but I'd rather just use their timeout than a timeout like this.

jnunemaker:
Looks good other than one comment I left.

spec/flipper/adapters/mongo_spec.rb
listrophy: Since it truly is a separate concern, I just reverted the commit related to this. Setting the timeout in the mongo client initialization can be a separate PR.

PR #89, opened 2015-11-17
Add a link back to the app via a button.

delynn:
Very similar to [Sidekiq #1875](https://github.com/mperham/sidekiq/pull/1875).

### DIFF
--- a/lib/flipper/ui.rb
+++ b/lib/flipper/ui.rb
@@ -11,6 +11,10 @@ require 'flipper/ui/middleware'
 
 module Flipper
   module UI
+    class << self
+      attr_accessor :app_url
+    end
+
     def self.root
       @root ||= Pathname(__FILE__).dirname.expand_path.join('ui')
     end
--- a/lib/flipper/ui/views/layout.erb
+++ b/lib/flipper/ui/views/layout.erb
@@ -18,6 +18,9 @@
   <body>
     <div class="container header">
       <a class="btn btn-sm right" href="<%= script_name %>/features/new">Add Feature</a>
+      <% if Flipper::UI.app_url %>
+        <a class="btn btn-sm right" href="<%= Flipper::UI.app_url %>">Back to App</a>
+      <% end %>
 
       <ol class="breadcrumb">
         <% @breadcrumbs.each do |breadcrumb| %>
--- a/spec/flipper/ui/actions/features_spec.rb
+++ b/spec/flipper/ui/actions/features_spec.rb
@@ -16,6 +16,22 @@ RSpec.describe Flipper::UI::Actions::Features do
       expect(last_response.body).to include("stats")
       expect(last_response.body).to include("search")
     end
+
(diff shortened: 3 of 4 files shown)


jnunemaker:
Neat. Seems useful. Could you drop a screenshot of it in action? 

This does make me wonder if we should start with a config object of some sort. Without that we could end up with a lot of these options and they would all be separate. With a config object, we can localize the config stuff. I guess we can always deprecate someday and have it write through to the config object if/when we need it. 

It also makes me wonder if there is any way we can infer it by default perhaps based on the rack env. Any thoughts on that?

delynn:
@jnunemaker Sure here's a screenshot:

screen shot 2015-11-18 at 5 52 57 pm

I will look at the rack env--there might be something that gets set. Or yeah, maybe we can just remove the path prefix for the mounted app and use that as a sensible default?

I agree that it's probably a good idea to create a config object. My goal was to create an MVP of this. I'd be happy to wrap it in a formal config object if you'd like, just let me know.

jnunemaker:
> Or yeah, maybe we can just remove the path prefix for the mounted app and use that as a sensible default?

Yeah, something like that is what I was thinking.

> My goal was to create an MVP of this. I'd be happy to wrap it in a formal config object if you'd like, just let me know.

Totally understood. Don't go that far yet. Let's see if we can do something smart from rack. If not, we can roll with what you have.

Seeing the pic I wonder if it would be better to make it a part of the breadcrumb. App > Flipper Home > Features or something. What do you think?

delynn:
I actually thought about doing that originally as well (and think it's probably a more appropriate spot). When I started implementing it though, it required quite a bit to change with how breadcrumbs are recorded and rendered, so I again went with the least invasive change.

What if I create a different branch that has the changes necessary for the breadcrumb solution? We can then evaluate which one we like more?

jnunemaker:
> What if I create a different branch that has the changes necessary for the breadcrumb solution? We can then evaluate which one we like more?

That sounds awesome.

jnunemaker:
Could we just add a breadcrumb to where we initialize to an empty array? 

https://github.com/jnunemaker/flipper/blob/9ea8a7df5aad222bd3332f09c082071a1898c835/lib/flipper/ui/action.rb#L59

delynn:
Yeah, that's part one for sure. The issue I ran into was actually related to the rendering. Breadcrumbs currently always have the `script_name` prepended to the path: https://github.com/jnunemaker/flipper/blob/9ea8a7df5aad222bd3332f09c082071a1898c835/lib/flipper/ui/views/layout.erb#L28

jnunemaker:
Oh, right. So maybe we remove the script name there and move it in to where we append the breadcrumb in the helper? That would allow first to not include script name but the rest would.

delynn:
Yeah, that _should_ work. I will try to find some today to make this happen!

delynn:
Closing thanks to #93

PR #93, opened 2015-12-05
Add a link back to the app via a breadcrumb.

delynn:
An alternate approach that uses the existing breadcrumbs and also has a really sensible default.

screen shot 2015-12-04 at 5 04 13 pm

### DIFF
--- a/lib/flipper/ui/action.rb
+++ b/lib/flipper/ui/action.rb
@@ -57,6 +57,9 @@ module Flipper
         @code = 200
         @headers = {"Content-Type" => "text/plain"}
         @breadcrumbs = []
+        if app_path
+          @breadcrumbs << Breadcrumb.new("App", app_path)
+        end
       end
 
       # Public: Runs the request method for the provided request.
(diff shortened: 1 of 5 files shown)


delynn:
@jnunemaker Sorry this took so long, but life got a little hectic for me.

I think I prefer the way this turned out vs. #89.

lib/flipper/ui/action.rb
```diff
@@ -159,6 +162,10 @@ def active?
       # href - The String href for the anchor tag (optional). If nil, breadcrumb
       #        is assumed to be the end of the trail.
       def breadcrumb(text, href = nil)
+        if href != nil
```
jnunemaker: I would do the more idiomatic `unless href.nil?`.

jnunemaker:
Tweaked a few things: https://github.com/jnunemaker/flipper/compare/48b073006604a0671bbfcfa95a8d9e632b660c41...0c5b88232e60d9da5598c6fb11a806af89087f29

delynn:
Looks great. Thanks!

Issue #101, opened 2016-02-09
Distinguish production environment from others


jetaggart:
One useful trick we use in production is to have a banner, or some sort of distinguishing color/banner/icon/popup that indicates that you are making changes to production. What do you all think about adding something similar to Flipper? It's definitely helpful to prevent mistakes from happening.

jetaggart:
IRT, Flipper UI.

jnunemaker:
We do the same at GitHub. Seems like a good idea. I haven't thought through how it would work or where it would go, but a PR would definitely activate my spidey sense. :)

blunderdome:
We'd be happy to see this

PR #102, opened 2016-02-10
Confirm feature enablement in production environment

jetaggart:
One idea on how to warn. Just a simple confirm only in production. 

#101

### DIFF
--- a/lib/flipper/ui/public/js/production.js
+++ b/lib/flipper/ui/public/js/production.js
@@ -1,0 +1,5 @@
+$(function() {
+  $("[data-warn-on-click]").on("click", function() {
+    return confirm("You are in PRODUCTION. Are you sure?");
+  });
+});
--- a/lib/flipper/ui/views/feature.erb
+++ b/lib/flipper/ui/views/feature.erb
@@ -24,10 +24,10 @@
       <%== csrf_input_tag %>
 
       <% unless @feature.on? %>
-        <button type="submit" name="action" value="Enable" class="btn btn-danger tooltipped tooltipped-n" aria-label="Enable for everyone">Enable</button>
+        <button type="submit" name="action" value="Enable" class="btn btn-danger tooltipped tooltipped-n" data-warn-on-click aria-label="Enable for everyone">Enable</button>
       <% end %>
 
-      <button type="submit" name="action" value="Disable" class="btn tooltipped tooltipped-n" <% if @feature.off? %>disabled aria-label="Already disabled for everyone"<% else %>aria-label="Disable for everyone by clearing all percentages, groups and actors."<% end %>>Disable</button>
+      <button type="submit" name="action" value="Disable" class="btn tooltipped tooltipped-n" data-warn-on-click <% if @feature.off? %>disabled aria-label="Already disabled for everyone"<% else %>aria-label="Disable for everyone by clearing all percentages, groups and actors."<% end %>>Disable</button>
     </form>
   </div>
 </div>
--- a/lib/flipper/ui/views/layout.erb
+++ b/lib/flipper/ui/views/layout.erb
@@ -39,5 +39,8 @@
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.2/jquery.min.js"></script>
     <script src="<%= script_name %>/js/bootstrap.min.js"></script>
     <script src="<%= script_name %>/js/application.js"></script>
+    <% if ENV.values_at("RACK_ENV", "RAILS_ENV").include? "production" %>
+      <script src="<%= script_name %>/js/production.js"></script>
+    <% end %>
   </body>
 </html>


jnunemaker:
I appreciate the idea, but I personally don't want to confirm every click. I wonder if we could allow setting the environment in config somewhere and then just add environment to the breadcrumb (ie: Production > Home > Features > stats? That feels a little terrible too. 

How we do it at GitHub is a bar across the top that is a different color, but that bar contains other useful information, which we don't have to present here in the flipper ui.

Another idea would be perhaps a css color config option that changes the color of the breadcrumb/add feature section's outline? That would allow people to set color based on whatever environment. 

Any other ideas?

jetaggart:
Yeah we do a similar thing with a bar/color. What if the nav has the environment name centered in it (to the right of the breadcrumbs)?

jnunemaker:
That could work.

jnunemaker:
Closing as stale. Feel free to bring this back to life at some point.

krnjn:
Thanks @jnunemaker for this gem, really loving the implementation / cleanliness and ease with it so far! I saw this issue as this was something that we had thought would be nice to have when disabling for a specific actor. 

The reason I think it's _slightly_ different than the use case described above is that it's pretty easy to disable a feature for a specific actor and pretty hard (nigh impossible) to see which actor you **just** disabled the feature for. Our concern is that someone from our team may accidentally click disable for a specific actor (and without confirmation, it would go through) and if we needed to know who they "accidentally" removed, we'd have to audit all of our clients who have access to that given feature and compare it against the remaining actors who it's enabled for (right now numbers in the 100s).

I can totally get why it's annoying to have a confirm on other enable / disable states for entire groups or entire features (there isn't much of a downside to either enable or disable for a group altogether), but it can be harder to "undo" the disable for a specific actor given the scenario I described above (especially as there isn't a flash message stating something like "You've disabled X feature for user:56"). The place where I was thinking of such a change is mainly on this form here: https://github.com/jnunemaker/flipper/blob/master/lib/flipper/ui/views/feature.erb#L214

Just a thought -- happy to hear thoughts / workarounds either way.

jnunemaker:
@krnjn I think if enough people want this, then we can add it. I'm game and would be open to a pull request. I'd add a [config option to Flipper::UI](https://github.com/jnunemaker/flipper/blob/c58dd5245b2236ba95d76889b6d8324160019cea/lib/flipper/ui/configuration.rb) (maybe named confirm?) and default it to `false` for backwards compatibility. Then, in `Flipper::UI` I'd add the confirm if that config option is `true`.

That said, have you considered adding an audit log? If so and rejected, why? If not, that would definitely help more. Even with confirm you can errantly click ok. To implement an audit log, you can provide an instrumenter (like `ActiveSupport::Notifications`) and subscribe to adapter enable/disable (event is `adapter_operation.flipper`). 

Here is where the adapter events get instrumented (shows payload): https://github.com/jnunemaker/flipper/blob/c58dd5245b2236ba95d76889b6d8324160019cea/lib/flipper/adapters/instrumented.rb#L121-L148

Here is an example of setting up an instrumenter and subscriber:
https://github.com/jnunemaker/flipper/blob/c58dd5245b2236ba95d76889b6d8324160019cea/examples/instrumentation.rb

If you had a table with columns like operation, feature, gate, and thing you'd be set. That alone wouldn't give you who performed the operation, but would make reversing it very easy. Probably wouldn't be hard to add something like this to flipper that would only take a few lines to configure. We could always show it in the UI for a feature if it was configured as well. Just having it in console though would help for restoration in situations like this. Honestly, most of the time the data stored per feature is small enough you could use jsonb in postgres and serialize all the gate values on each enable/disable to make rollback really easy. 

We do pretty close to what I just said on https://www.flippercloud.io. 

## Audit log for entire environment:

Screen Shot 2020-04-12 at 9 14 46 PM

## Audit log for individual feature:

Screen Shot 2020-04-12 at 9 15 08 PM

Our schema is different than what I suggest above, but that is for a variety of reasons. Here it is if you are curious though:

```ruby
  create_table "feature_audits", id: :serial, force: :cascade do |t|
    t.integer "feature_id", null: false
    t.integer "actor_id", null: false # actor id/type is User or Token in our system
    t.string "actor_type", null: false
    t.integer "action", null: false # enable/disable/enable_actor/disable_actor/etc
    t.string "value" # percentage for % of actors, flipper_id for actors gate and group name for groups gate
    t.datetime "created_at", null: false
    t.integer "environment_id", null: false
    t.index ["environment_id"], name: "index_feature_audits_on_environment_id"
    t.index ["feature_id"], name: "index_feature_audits_on_feature_id"
  end
```

Any of that helpful?

krnjn:
Hey @jnunemaker – thanks for the super detailed response. Really grateful for additional pointers for how we can make our implementation even more secure. 

We hadn't decided on the audit log one way or the other -- but given the complexity involved, I think we'd want to have a more certain set of uses for it before we invest the time to build the needed infra for it (right now we're using Redis as our adapter and had wanted to avoid integrating our feature flag needs with our database for now). 

WRT to the confirmation component:

> Even with confirm you can errantly click ok.

While I agree a user can definitely still make a mistake and remove an actor that they were not supposed to (for which case the audit log does seem like the ideal / best solution to resolve and undo that change), what I'm mostly concerned about is the [fat-finger error](https://en.wikipedia.org/wiki/Fat-finger_error) in which you accidentally clicked right when you're hovering over the trashcan icon in the actors section. 

I think in those cases the confirm is pretty useful as a user is able to cancel an accidental click of the remove (e.g. truly an accidental click, not an erroneous removal). I'll concede that with that logic you could argue that you should add confirms everywhere, but I agree with your earlier points that it can be annoying to do that for actions that are not scoped to a single actor and its easy to disable / re-enable a feature entirely in case you accidentally clicked elsewhere in the dashboard.

No worries if this feels too specific -- we'll see if anyone else cares for this and it merits the change (or if you're up for adding it, we'd be much obliged!).

barnett:
Hey 👋 

Just ran into a similar issue last week with my team where the enable/disabled was fat-fingered (by myself 🙃)  and was quite painful to put together the history. Didn't realize the audit log functionality so we'll look to implement such but think a confirm for disabling where data is deleted and thus harder to resurrect would be a great help to prevent such issues again.

Issue #104, opened 2016-02-24
Disable a boolean gate via the UI would auto-remove it


raysegantii:
It's cause: https://github.com/jnunemaker/flipper/commit/ec711815 -- the feature, boolean gate, is gone once I press Disable and then leave the page. I would have to re-add the feature although I simply want to disable it for awhile for experiment.

What was the reason to introduce that commit? Would it be reverted?

jnunemaker:
Adapters should only remove the gate values, not the entire feature. I'm guessing this is based on the adapter you are using. As I look, I don't have an explicit test in the shared specs for that and some of the core adapters would fail if so. I'll get a test added and update the core adapters to correct that. 

Which adapter were you using?

raysegantii:
I'm using the ActiveRecord, is there something wrong with it for handling boolean flag?

jnunemaker:
Hmm, no the AR adapter explicitly only clears the gate values, not the feature.
https://github.com/jnunemaker/flipper/blob/4c6a9978940e0f7b097e51df153c4cac49f141c8/lib/flipper/adapters/active_record.rb#L66

When you say AR adapter, do you mean the core one I support or one of the community ones that exist?

jnunemaker:
It looks like flipper activerecord handles the gate clearing incorrectly, if you are using that.
https://github.com/bgentry/flipper-activerecord/blob/cbdc8ab4ec57799343a6c8a9c2192f6bd87bc3d8/lib/flipper/adapters/activerecord.rb#L38

They delete the entire feature.

PR #105, opened 2016-02-24
Not to clear boolean gate

raycheung:
To address #104


jnunemaker:
Thanks for the PR! 

I don't think this quite solves the issue. If the adapter calls clear for the disable of a boolean, then it fixes the issue, but if it doesn't, the problem could remian. I think the root issue is that I don't have an explicit test in the shared specs for this. It is important that clear only clears gate values, not the entire feature. I've opened https://github.com/jnunemaker/flipper/pull/106 to address #104 so I'll be closing this. Let me know if that makes sense or not.

PR #107, opened 2016-02-25
Add support for Rack 2.

lucasmazza:
Require for folks trying out Rails 5. I can tweak the build matrix to run the tests against rack `2.0.0.alpha` in the same way that the repository supports Rails 3.2 + 4.2 if we want to go that road.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,13 @@ services:
   - redis-server
   - mongodb
 env:
+  - RAILS_VERSION=5.0.0.beta3
   - RAILS_VERSION=4.2.5
   - RAILS_VERSION=3.2.21
+matrix:
+  # don't run rails 5 on ruby versions that can't install rack 2
+  exclude:
+    - rvm: 2.0.0
+      env: RAILS_VERSION=5.0.0.beta3
+    - rvm: 2.1.8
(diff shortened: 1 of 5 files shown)


jnunemaker:
I'd love a build matrix tweak to ensure that this is tested. Happy to merge ahead of rails 5 release if we at least have rails 5 matrix in here.

lucasmazza:
@jnunemaker updated :+1: there's also a new commit to fix a deprecation warning from Rails 5 but I can squash them together if you want.

jnunemaker:
> Gem::InstallError: rack requires Ruby version >= 2.2.2.

Looks like rack 2 doesn't like < ruby 2.2. I can fix that up after the CI run is done. 

```
matrix:
  # don't run rails 5 on ruby versions that can't install rack 2
  exclude:
    - rvm: 2.0.0
      env: RAILS_VERSION=5.0.0.beta3
    - rvm: 2.1.8
      env: RAILS_VERSION=5.0.0.beta3
```

I think I just need to add that. Noting it here so I don't forget.

lucasmazza:
@jnunemaker thanks :heart:

jnunemaker:
:heart: for the PR!

PR #111, opened 2016-03-04
Add read only adapter

### LINKED ISSUE #91: Add read-only mode
I appreciate the fact that `flipper-ui` is a separate gem from the core `flipper` gem, as it means we can have a separate lightweight management application (using flipper-ui) for enabling and disabling feature flags.  In fact, we intend to use a single lightweight rack application to manage feature flags for a few smaller ruby applications (and one large one).  However, with this kind of setup it would be ideal to prevent accidental changes to state except from the management application.  A 'read_only' setting would be a simple way to address this.

jnunemaker:
This adds read only adapter for situations where the application uses flipper but the UI runs separate and handles the updating.

fyi @dangerp

closes #91
closes #92

### DIFF
--- a/lib/flipper/adapters/read_only.rb
+++ b/lib/flipper/adapters/read_only.rb
@@ -1,0 +1,51 @@
+module Flipper
+  module Adapters
+    # Public: Adapter that wraps another adapter and raises for any writes.
+    class ReadOnly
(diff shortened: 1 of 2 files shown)


dangerp:
Awesome, thanks!

PR #113, opened 2016-03-18
Allow truthy and falsey values to be enough in group match blocks

### LINKED ISSUE #110: `Flipper.register` should match any truthy value, not just `true`
We recently wrote something like the following, hoping to set a generic group for flipper usage.

    Flipper.register(:reviewers) do |actor|
      actor.try(:email) =~ /@example.com\z/
    end

It didn't work, because `String#=~` returns an integer (the position of the match in the string) rather than true/false. Flipper groups only respond if the `register` block [explicitly returns `true`](https://github.com/jnunemaker/flipper/blob/master/lib/flipper/types/group.rb#L18).

This seems a bit unrubyish, so I wondered if there was a specific reason that only `true` was allowed rather than (IMO) the more usual convention of any truthy value?

jnunemaker:
## Problem

Ruby general is ok with truthy and falsey values over explicitly true and false. Flipper was different and expected exactly true to be returned by the flipper group match block. 

## Solution

Allow truthy to work. This means that blocks that return truthy values which previously would have returned false will now return true.

Fixes #110

### DIFF
--- a/spec/integration_spec.rb
+++ b/spec/integration_spec.rb
@@ -9,8 +9,8 @@ RSpec.describe Flipper do
 
   let(:actor_class) { Struct.new(:flipper_id) }
 
-  let(:admin_group)  { flipper.group(:admins) }
-  let(:dev_group)    { flipper.group(:devs) }
+  let(:admin_group) { flipper.group(:admins) }
+  let(:dev_group)   { flipper.group(:devs) }
 
   let(:admin_thing) { double 'Non Flipper Thing', :flipper_id => 1,  :admin? => true, :dev? => false }
   let(:dev_thing)   { double 'Non Flipper Thing', :flipper_id => 10, :admin? => false, :dev? => true }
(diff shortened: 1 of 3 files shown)


spec/integration_spec.rb
```diff
@@ -9,12 +9,17 @@
 
   let(:actor_class) { Struct.new(:flipper_id) }
 
-  let(:admin_group) { flipper.group(:admins) }
-  let(:dev_group)   { flipper.group(:devs) }
+  let(:admin_group)  { flipper.group(:admins) }
+  let(:dev_group)    { flipper.group(:devs) }
+  let(:truthy_group) { flipper.group(:truthy) }
+  let(:falsey_group) { flipper.group(:falsey) }
```
gareth: Maybe I don't know how your tests work, but are these two groups used anywhere?

jnunemaker: I think I was going to use them and then realized I could just use admin?. I'll remove them. Good catch.

Issue #103, opened 2016-02-17
Conflict between Time & Actors percentages


olivierlacan:
The interface appears to suggest that it's possible to enable, for example, a feature for 50% of actors *and* 25% of the time. 

![image](https://cloud.githubusercontent.com/assets/65950/13122524/3bcb85f4-d585-11e5-96d0-22a9e64bad37.png)

In my experience when the actors percentage gate is enabled, it means that whoever does have the feature enabled will have it enabled 100% of the time. In that case, the interface should not allow the other gate to be enabled simultaneously, should it?

jnunemaker:
Flipper is all about enabling. If you enable for 50% of actors and 25% of the time, 50%-ish of actors will be enabled and it should also randomly be enabled 25% of the time. It checks the gates in order (https://github.com/jnunemaker/flipper/blob/4c6a9978940e0f7b097e51df153c4cac49f141c8/lib/flipper/feature.rb#L323-L327). If any return true, it stops. If they return false, flipper will keep going until one is true. 

I say all that to say that each gate is about true/false individually. The gates are not used in combination (aka enabled for 50% of actors 25% of the time which I think is what you are saying?). Think of each gate as OR'd. If any are true, things are enabled. If all are false, it is disabled. 

Am I understanding what you are asking? Does that explain it?

olivierlacan:
Ok, it's not clear at all from the interface that it's an `OR` situation because they co-exist on the exact same line. That's what's so confusing about this. I honestly think that flipping one gate on should disable the other if there's an order dependency, otherwise you may have people using Flipper thinking they do indeed have 50% of actors seeing a feature 25% of the time.

Does that make sense?

jnunemaker:
There is no order dependency. All gates are OR'd. The order in which they return true doesn't matter. Either one or more of them returns true or none of them do. Using one gate at a time is actually hindering. At GitHub, we often enable features for specific users or repositories. Then eventually we enable for a percentage of actors or some other option. Any of those that return true let the user in to the feature and that is very useful. 

[...]

PR #114, opened 2016-03-18
Clarify percentage of time/actor functionality being independent in UI

jnunemaker:
## Problem

Sometimes people get confused and think that percentage of actors and time work in tandem. They don't. They work completely independently of each other.

## Solution

A small help note below them to explain the difference. I don't love repeating basically the same text in two spots, but I tried a few things and went with this. Open to better ideas if anyone has any.

screen shot 2016-03-18 at 4 36 14 pm

## Alternatives

The original suggestion took too much focus from the actual form controls.

screen shot 2016-03-18 at 4 27 54 pm

I also tried a tooltip, but it felt too hidden and didn't allow for much text. 

screen shot 2016-03-18 at 4 25 50 pm

cc @olivierlacan for thoughts as he brought it up here https://github.com/jnunemaker/flipper/issues/103

### DIFF
--- a/lib/flipper/ui/assets/stylesheets/application.scss
+++ b/lib/flipper/ui/assets/stylesheets/application.scss
@@ -52,3 +52,7 @@ body {
 .feature-enabled-gates {
   width:380px;
 }
+
+p.help {
+  color: #999;
+}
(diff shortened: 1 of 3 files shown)


olivierlacan:
This is fantastic @jnunemaker. I really appreciate you taking the time to make a patch for this. Sorry it should have been my initiative since I proposed it.

PR #115, opened 2016-04-01
Initial Flipper#API

AlexWheeler:
Initial layout of Flipper#API.  Not ready to merge, but wanted to see if you think this is headed in the right direction.

The API will allow applications other than the rack application including Flipper to communicate with it.
This PR
  * creates gem dependent on Flipper in order to access the flipper DSL and classes.
  * ports many of the basic classes from Flipper#UI i.e. action, action_class, etc. but removes anything having to do with a UI and keeps it related to just the API
  * namespaces base classes for the API under api/ and everything else that could change in the future (i.e. decorators, actions, etc) under api/__version__

One of bigger things to think about is how to allow clients to authenticate.  Possibly add a service that generates secure credentials and clients could use that, or allow flipper users to define their own key/secret and use something like basic auth?


jnunemaker:
Super excited to look this over. Missed it initially. Sorry. I'll try to give it a look tomorrow.

lib/flipper/api/v1/actions/features.rb
```diff
@@ -0,0 +1,22 @@
...
+    module V1
+      module Actions
+        class Features < Api::Action
+          route %r{api/v1/features/}
+
+          def get
+            features = flipper.features.map do |feature|
+              V1::Decorators::Feature.new(feature)
+            end
+            json_response({features: features})
```
jnunemaker: YES. Always use hashes for responses. Way more flexible for the future. Good stuff!

spec/flipper/api/v1/actions/features_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'helper'
+
+RSpec.describe Flipper::Api::V1::Actions::Features do
+  describe 'GET /alex' do
```
jnunemaker: alex? :) Perhaps /api/v1/features.

flipper-api.gemspec
```diff
@@ -0,0 +1,25 @@
...
+  gem.license       = "MIT"
+  gem.homepage      = "https://github.com/jnunemaker/flipper"
+  gem.files         = `git ls-files`.split("\n").select(&flipper_api_files) + ["lib/flipper/version.rb"]
+  gem.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n").select(&flipper_api_files)
+  gem.name          = "flipper-api"
+  gem.require_paths = ["lib"]
+  gem.version       = Flipper::VERSION
+
+  gem.add_dependency 'rack', '>= 1.4', '< 3'
+  gem.add_dependency 'rack-protection', '~> 1.5.3'
```
jnunemaker: No need for this.

flipper-api.gemspec
```diff
@@ -0,0 +1,25 @@
...
+  gem.add_dependency 'flipper', "~> #{Flipper::VERSION}"
+  gem.add_dependency 'erubis', '~> 2.7.0'
```
jnunemaker: No need for this.

jnunemaker:
> creates gem dependent on Flipper in order to access the flipper DSL and classes.

👍 

> ports many of the basic classes from Flipper#UI i.e. action, action_class, etc. but removes anything having to do with a UI and keeps it related to just the API

👍  At some point farther down the road it may make sense to move some of the shared stuff into one place, but I'm not concerned at all. Let's duplicate for now until it hurts. I suspect it won't.

> namespaces base classes for the API under api/ and everything else that could change in the future (i.e. decorators, actions, etc) under api/version

👍  Works for me. I am fine with versioning in the route.

> One of bigger things to think about is how to allow clients to authenticate. Possibly add a service that generates secure credentials and clients could use that, or allow flipper users to define their own key/secret and use something like basic auth?

Since this is just rack, I'd leave out authentication, unless I'm overlooking something. That is what I did for the flipper UI. It is relatively easy to just put some rack middleware or a rails routing constraint in front of flipper-ui to force authentication. We can do the same for the API. I like going that route as it doesn't force any particular auth on people. Thoughts?

flipper-api.gemspec
```diff
@@ -0,0 +1,25 @@
...
+  gem.description   = "Rack middleware that provides an  API for the flipper gem."
+  gem.license       = "MIT"
+  gem.homepage      = "https://github.com/jnunemaker/flipper"
+  gem.files         = `git ls-files`.split("\n").select(&flipper_api_files) + ["lib/flipper/version.rb"]
+  gem.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n").select(&flipper_api_files)
+  gem.name          = "flipper-api"
+  gem.require_paths = ["lib"]
+  gem.version       = Flipper::VERSION
+
+  gem.add_dependency 'rack', '>= 1.4', '< 3'
```
jnunemaker: Thinking long term, but not necessarily for now we may want to move these dependency requirements to a constant somewhere in the versions file so that API and UI can share the same.

lib/flipper/api.rb
```diff
@@ -0,0 +1,22 @@
+require 'rack'
+require 'flipper'
+
+require 'rack/methodoverride'
+require 'rack/protection'
+require 'flipper/api/middleware'
+
+module Flipper
+  module Api
+    class << self
```
jnunemaker: This can die.

lib/flipper/api.rb
```diff
@@ -0,0 +1,22 @@
+require 'rack'
+require 'flipper'
+
+require 'rack/methodoverride'
+require 'rack/protection'
```
jnunemaker: 💀

AlexWheeler:
Thanks so much for the feedback @jnunemaker.  Definitely agree down the road we can move shared stuff (i.e. ActionCollection, Action, etc.) into one place but really like the idea of waiting until it hurts because there's nothing worse than cornering yourself with a bad abstraction.

re: authentication ,you're totally right we should just leave that to the users as with Flipper::UI

This week I'm going to clean this up, remove any unnecessary dependencies and submit a real PR.  Going to close for now and reopen once I've made the changes.

jnunemaker:
@AlexWheeler sounds great!

Issue #62, opened 2015-04-27
Add Flipper::API module


jnunemaker:
Now that we have Flipper::UI it would be nice to have Flipper::API. Flipper::API would provide the same functionality as the UI but all json responses instead of html/css views. Clients could then be created for chatops/hubot to enable/disable features and all that.

digidigo:
Would this be required to allows an angular app to check whether a feature should be on or off for a user.

jnunemaker:
No, for that I would do whatever you need to on your end. This is so other systems can enable/disable and do other things through an API instead of a web UI. For example, if your company ran https://hubot.github.com/, hubot could talk to the API to turn features on/off and do more. 

For angular, I would just make routes on your own that do so and make angular talk to them.

AlexWheeler:
@jnunemaker is anyone working on this?  I'd be stoked to start and/or contribute if someone has a branch going

jnunemaker:
@AlexWheeler not currently. Go for it!

jnunemaker:
Perhaps start with one or two routes and we can settle on some stuff rather than doing a bunch of work and possibly changing directions? Perhaps GET /features, GET /features/:feature-key, POST /features/:feature-key/boolean?

AlexWheeler:
@jnunemaker sounds great!

jnunemaker:
Created a [project for the api](https://github.com/jnunemaker/flipper/projects/2) and added the notes I could think of that still need to be done. Also tried to track all the completed work for those that are curious.

AlexWheeler:
Closing since this successfully shipped a while back! 🎉

PR #116, opened 2016-04-21
Create Flipper#Api

AlexWheeler:
* Sets up Flipper#Api module
* version namespaced
* moves action_collection out of UI namespace and into flipper since imho abstract enough 
- api gemspec
- api middleware
- api action
- api boolean gate
- api boolean gate spec

creates Flipper::Api::App so that can return status code other than 200 when action class is not found for requested URI.

Only thing I'm not sure about is what to do about decorators.  UI::Feature / UI::Gate have the logic we need and nothing really extra.  We could copy/paste those and namespace under API::V1, but feels crappy doing that, at same time don't want to rely too much on them since their logic could change.  With that said the logic is pretty minimal and mostly display logic.  Thinking of making a base class Decorators::Feature / Decorators::Gate and then Api versions could inherit and just override the as_json method and UI::Decorators::Feature/Gate could just not override and let that base decorator respond?  Interested in others opinions?

Once decide on this can finish api#boolean_gate spec, which will probably just respond w/ the serialized feature


lib/flipper/api/v1/actions/boolean_gate.rb
```diff
@@ -0,0 +1,21 @@
...
+    module V1
+      module Actions
+        class BooleanGate < Api::Action
+          route %r{api/v1/features/[^/]*/(enable|disable)/?\Z}
+          
+          def put
+            feature_name = Rack::Utils.unescape(request.path.split("/")[-2])
+            feature = flipper[feature_name.to_sym]
+            action = Rack::Utils.unescape(request.path.split("/").last)
+            feature.send(action)
```
jnunemaker: Let's verify that the action is one in a valid set of actions.

```ruby
VALID_ACTIONS = Set.new(["enable", "disable"]).freeze

if VALID_ACTIONS.include?(action)
  feature.send(action)
  json_response ...
else
  ... some response that says it is not valid action
end
```

Or:

case action
when "enable"
  feature.enable
when "disable"
  feature.disable
else
  ... some json response that says action is not valid
end
```

I don't want someone to be able to call a ruby method based on the url.

File: lib/flipper/api/v1/actions/boolean_gate.rb [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+  module Api
+    module V1
+      module Actions
+        class BooleanGate < Api::Action
+          route %r{api/v1/features/[^/]*/(enable|disable)/?\Z}
+          
+          def put
+            feature_name = Rack::Utils.unescape(request.path.split("/")[-2])
+            feature = flipper[feature_name.to_sym]
+            action = Rack::Utils.unescape(request.path.split("/").last)
```
jnunemaker: route_parts = request.path.split("/") and then we can re-use that for feature_name and action.

lib/flipper/ui/action_collection.rb
```diff
@@ -1,20 +0,0 @@
-module Flipper
```
jnunemaker: I'd actually rather just duplicate what we need for the API gem for now and we can extract things later. Let's completely duplicate and worry about tidying up at the end. Seem ok? Any strong feelings against that?

lib/flipper/api/v1/actions/boolean_gate.rb
jnunemaker: OIC. You only match enable/disable in route so it is safe by those means. I guess that seems ok.

spec/flipper/api/v1/actions/boolean_gate_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      put '/api/v1/features/my_feature/disable'
+    end
+
+    it 'disables feature' do
+      expect(last_response.status).to eq(200)
+      expect(flipper[:my_feature].off?).to be_truthy
+    end
+  end
+
+  describe 'invalid paremeter' do
```
jnunemaker: Thanks for this test. :D

jnunemaker:
> We could copy/paste those and namespace under API::V1, but feels crappy doing that, at same time don't want to rely too much on them since their logic could change.

Duplicate away! I'm totally fine with duplication like this. We can always extract common parts down the road. Duplication is easy to get rid of, sharing things and ensuring that we always want to share and know all the interactions of that sharing is harder. I usually duplicate at least two times before I extract. I am totally fine with a wholesale copy paste of anything you need from the UI gem.

lib/flipper/ui/action_collection.rb
AlexWheeler: sounds great, going to make api/action_collection again, and put this back under ui::

lib/flipper/api/v1/actions/boolean_gate.rb
AlexWheeler: great idea will make that change and avoid the duplication. Could see some idea of this being moved into Api::Action at some point since just about every action uses some form of splitting on "/" and accessing an index, but for now I like how its very obvious right here
```

jnunemaker: > Could see some idea of this being moved into Api::Action at some point since just about every action uses some form of splitting on "/" and accessing an index

Yeah that would be totally fine by me. Good call.

jnunemaker:
Just ping me again whenever you are ready or if you lose interest for any reason (so I can pick it up).

AlexWheeler:
@jnunemaker sounds good, will ping you, just got to work but will finish this initial PR tonight, and then get /feature & /features done  this weekend or tonight depending how much coffee I have today

AlexWheeler:
@jnunemaker I think its in a pretty good spot for the initial api setup and boolean_gate.  Let me know what you think.  Was thinking we could break up feature and features into a new PR as well as the rest of the endpoints and documentation

jnunemaker:
Works for me. Merged.

jnunemaker:
refs #62

PR #119, opened 2016-05-05
Api /feature endpoint

AlexWheeler:
Add /feature endpoint

adds:
GET, DELETE actions
* Feature Decorator
* Gate Decorator
* specs

---------------------------------------------------------------------------------------------------------------------
Have two questions (answered - thank you):

1. lib/flipper/gate.rb:17
states that ```key``` is a private method.  Should json from ```get``` only return name and value, and omit key?

2. I want to return a 404 if a get request fails to find a feature, however flipper/dsl defines a public function ```[]``` and aliases it to ```#feature```  which creates an new feature if it doesn't already exist.  What do you think about changing this functionality so that the ```feature``` still creates a new feature, but ```[]``` will return nil if no feature exists for a given key?  I know this is public so we might not want to do that.  We could also add a module like Flipper:Api::Dsl and monkey-patch it, but this is bit messy.  Really curious of your thoughts, thanks!

### DIFF
--- a/lib/flipper/api/v1/decorators/feature.rb
+++ b/lib/flipper/api/v1/decorators/feature.rb
@@ -1,0 +1,33 @@
+require 'delegate'
+require 'flipper/api/v1/decorators/gate'
+
+module Flipper
+  module Api
+    module V1
+      module Decorators
+        class Feature < SimpleDelegator
+
+          # Public: The feature being decorated.
+          alias_method :feature, :__getobj__
+          
+          # Public: Returns instance as hash that is ready to be json dumped.
+          def as_json
+            {
+              'id' => name.to_s,
+              'state' => state.to_s,
+              'gates' => gates.map { |gate|
+                Decorators::Gate.new(gate, gate_values[gate.key]).as_json
+              }
+            }
+          end
+
+          private
+
+          def gate_values
+            feature.gate_values
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/flipper/api/v1/decorators/gate.rb
+++ b/lib/flipper/api/v1/decorators/gate.rb
@@ -1,0 +1,35 @@
+module Flipper
+  module Api
+    module V1
+      module Decorators
+        class Gate < SimpleDelegator
+          # Public the gate being decorated
+          alias_method :gate, :__getobj__
+
+          # Public: the value for the gate from the adapter.
+          attr_reader :value
+
+          def initialize(gate, value = nil)
+            super gate
+            @value = value
+          end
+
+          def as_json
+            {
+              'key' => gate.key.to_s,
+              'name' => gate.name.to_s,
+              'value' => value_as_json
+            }
+          end
+
+          private
+
+          # json doesn't like sets
+          def value_as_json
+            data_type == :set ? value.to_a : value
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/flipper/api/v1/actions/feature_spec.rb
+++ b/spec/flipper/api/v1/actions/feature_spec.rb
@@ -1,0 +1,120 @@
+require 'helper'
+
+RSpec.describe Flipper::Api::V1::Actions::Feature do
+  let(:app) { build_api(flipper) }
+  let(:feature) { build_feature }
+  let(:gate) { feature.gate(:boolean) }
+
+  describe 'get' do
+    context 'enabled feature' do
+
+      before do
+        flipper[:my_feature].enable
+        get 'api/v1/features/my_feature'
+      end
+
+      it 'responds with correct attributes' do
+        response_body = {
+          'id' => 'my_feature',
+          'state' => 'on',
+          'gates' => [
+            {
+              'key' => 'boolean',
+              'name' => 'boolean',
+              'value' => true
+            },
+            {
+              'key' => 'groups',
+              'name' => 'group',
+              'value' =>[]
+            },
+            {
+              'key' => 'actors',
+              'name' => 'actor',
+              'value' => []
+            },
+            {
+              'key' => 'percentage_of_actors',
+              'name' => 'percentage_of_actors',
+              'value' => 0
+            },
+            {
+              'key' => 'percentage_of_time',
+              'name' => 'percentage_of_time',
+              'value' => 0
+            }
+          ]
+        }
+
+        expect(last_response.status).to eq(200)
+        expect(json_response).to eq(response_body)
+      end
+    end
+
+    context 'disabled feature' do
+      before do
+        flipper[:my_feature].disable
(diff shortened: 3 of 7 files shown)


jnunemaker:
1. that is incorrect. It should be internal (ie: safe for ui and api but probably not best outside of stuff like that). Sorry! Go ahead and use it however you need here.

2. Returning nil would break everyone who is using this as most people use bracket notation from what I've seen. I checked and the web UI just shows that the feature is disabled. Perhaps you could just check if the feature is in the features set (exposed by DSL and adapters)? I think returning 404 makes sense for  GET.

jnunemaker:
Also, would you be interested in being a collaborator on the repo? You'd still work in pull requests like I do, but then you wouldn't have to work in a fork. Totally up to you. Let me know if you are interested in helping longer term or if you just want to stick short term on the API stuff. You've been doing great work so I'd be happy to add you.

AlexWheeler:
@jnunemaker thanks for the quick responses to those questions, will get this PR finished up soon! Being a collaborator on the repo sounds great to me.  I would definitely like to contribute longer term and have been having a lot of fun with the API PRs - has been a great way to learn the codebase, thanks

AlexWheeler:
@jnunemaker Believe we're at a good point with the Get/Delete actions for `/feature` endpoint.  Let me know what you think

lib/flipper/api/v1/decorators/feature.rb
```diff
@@ -0,0 +1,33 @@
...
+          # Public: The feature being decorated.
+          alias_method :feature, :__getobj__
+          
+          # Public: Returns instance as hash that is ready to be json dumped.
+          def as_json
+            {
+              'id' => name.to_s,
+              'state' => state.to_s,
+              'gates' => gates.map { |gate|
+                Decorators::Gate.new(gate, gate_values[gate.key]).as_json
```
jnunemaker: This will cause gate_values to be called for each gate which means an N+1 call against the adapter. That is why [I assign it to a temp var in the UI](https://github.com/jnunemaker/flipper/blob/c25930a6e176d535ce9b21faf4cb5b6c34b098d0/lib/flipper/ui/decorators/feature.rb#L21-L29) and then re-use the temp var.

lib/flipper/api/v1/decorators/feature.rb
```diff
@@ -0,0 +1,33 @@
...
+    module V1
+      module Decorators
+        class Feature < SimpleDelegator
+
+          # Public: The feature being decorated.
+          alias_method :feature, :__getobj__
+          
+          # Public: Returns instance as hash that is ready to be json dumped.
+          def as_json
+            {
```
jnunemaker: What are your thoughts on including name and/or key in this hash?

lib/flipper/api/v1/decorators/gate.rb
```diff
@@ -0,0 +1,35 @@
...
+          def initialize(gate, value = nil)
+            super gate
+            @value = value
+          end
+
+          def as_json
+            {
+              'key' => gate.key.to_s,
+              'name' => gate.name.to_s,
+              'value' => value_as_json
```
jnunemaker: I usually drop trailing commas here as ruby doesn't care and then I never have to remember to add a comma to drop in a new key/value. If you think about it when you tweak other stuff go for it. If not, no worries.

spec/flipper/api/v1/actions/feature_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+          'gates' => [
+            {
+              'key' => 'boolean',
+              'name' => 'boolean',
+              'value' => false
+            },
+            {
+              'key'=> 'groups',
+              'name'=> 'group',
+              'value'=>[]
```
jnunemaker: Minor: a space between => and [].

spec/flipper/api/v1/actions/feature_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+            },
+            {
+              'key' => 'actors',
+              'name' => 'actor',
+              'value' => []
+            },
+            {
+              'key' => 'percentage_of_actors',
+              'name' => 'percentage_of_actors',
+              'value'=>0
```
jnunemaker: Minor: a space between => and 0.

spec/flipper/api/v1/actions/feature_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+          'gates' => [
+            {
+              'key' => 'boolean',
+              'name' => 'boolean',
+              'value' => true
+            },
+            {
+              'key' => 'groups',
+              'name' => 'group',
+              'value' =>[]
```
jnunemaker: Same as below.

jnunemaker:
Looks great. I left some comments, but I'm going to go ahead and merge. This is close enough. I am still curious about your thoughts on the questions I left whenever you get a chance.

jnunemaker:
refs #62

lib/flipper/api/v1/decorators/feature.rb
AlexWheeler: @jnunemaker I think we should include key since:
- it will always be a string (feature#initialize  calls key = name.to_s).  That way the returned JSON value will always reflect the correct value.  If we returned name in JSON and it was a symbol as object attribute, it would serialize to JSON and be turned into a string, which it isn't since its a ruby symbol.  Could see that being confusing to a client expecting the true value.  A user can pass in a string key to flipper.feature() and it will always work since it calls to_sym

I don't think we need to include name as pretty_name like in UI because a client should just get their data and be able to make any decisions around formatting they'd like. With this change we would remove `id` from the hash.  What do you think?
```
{
  'key' => 'some_key'
  'state' => 'some_state',
  'gates' => [# some gates]
}
```

jnunemaker: That is what I was thinking too. I'll drop key in.

jnunemaker: Oh, actually I just noticed you included id as name.to_s. I missed that. I think I'd rather see key => key than id. What was your reasoning for using id for key?

AlexWheeler: @jnunemaker totally mistake  on my end, forgot to change from when was first building it and referencing some of the ::UI responses.  We should definitely have key => key!

jnunemaker: Will do.

jnunemaker: Fixed by https://github.com/jnunemaker/flipper/commit/4864077a1daeeaac516b4a2fc9f15851aca41d82

PR #122, opened 2016-05-12
Allow changing of actor or percentage gates in active_record adapter

vegantech:
Choosing a second (or additional) percentage creates an additional row in the feature_gates table.  The one that is used is arbitrary, typically the first one created.

Additionally, if a user chooses a percentage for a feature that was previously chosen, this will create an exception.

The fix should work in rails 3.2 and 4.x.

### DIFF
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -105,11 +105,11 @@ module Flipper
       def enable(feature, gate, thing)
         case gate.data_type
         when :boolean, :integer
-          @gate_class.create!({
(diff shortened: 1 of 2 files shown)


jnunemaker:
Well that is less than ideal. :( Thanks for the patch. I suspect that disable suffers the same problem. I can fix that up after I merge this.

jnunemaker:
I tweaked your approach a bit https://github.com/jnunemaker/flipper/pull/124. Can't believe I missed that. Thanks for finding and sorry for any pain it caused you and the lost time tracking it down. :)

PR #126, opened 2016-05-27
add #remove method to the Feature class

svarks:
I thought it might make sense for it to be available on the feature level:

```rb
flipper[:feature].remove
```

### DIFF
--- a/lib/flipper/api/v1/actions/feature.rb
+++ b/lib/flipper/api/v1/actions/feature.rb
@@ -20,7 +20,7 @@ module Flipper
 
           def delete
             feature = flipper[feature_name]
-            flipper.adapter.remove(feature)
+            feature.remove
             json_response({}, 204)
           end
 
--- a/lib/flipper/api/v1/actions/feature.rb
+++ b/lib/flipper/api/v1/actions/feature.rb
@@ -19,13 +19,17 @@ module Flipper
           end
 
           def delete
-            feature = flipper[feature_name]
-            feature.remove
-            json_response({}, 204)
+            if feature_names.include?(feature_name)
+              flipper.remove(feature_name)
+
+              json_response({}, 204)
+            else
+              json_response({}, 404)
+            end
           end
 
           private
-          
+
           def feature_name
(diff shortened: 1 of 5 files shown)


jnunemaker:
This brings of up the question of what does remove mean. Adapter#remove is for removing the feature from the set of known features, but I would assume Feature#remove would also clear the gate values and such. Also, we currently don't have Feature#add, which would be the reciprocal method of #remove. I think for these reason I'm inclined to just let the api use the adapter remove method and not add remove to feature. What are your thoughts on this?

svarks:
Hmm that's a good point. I've been mostly using this api:

```rb
flipper.class                  # => Flipper::DSL

flipper[:feature]              # calls Feature.new(name, @adapter, ...)
flipper[:feature].enable       # calls adapter.enable(self)
flipper[:feature].disable      # calls adapter.disable(self)
flipper[:feature].enable(user)
```

So it seems that almost every adapter method is exposed through DSL, except for `#remove`.
And I just noticed that it wasn't obvious to me how to remove redis data when you stop using the feature.
I don't have a strong preference either way though.

jnunemaker:
I lied. Adapter#remove is for removing feature from set of known features and clearing gate values. Totally forgot. My bad. I think the only thing this is missing is adding the remove method to the DSL with an accompanying spec and changing the web action to use feature#remove instead of adapter#remove. If you have time to knock those out, go for it. If not, I'll try to get it done in the coming weeks.

svarks:
Thanks John! I've just added the `DSL#remove` method. Please let me know if I missed anything, I'm happy to make any other changes.

lib/flipper/api/v1/actions/feature.rb
```diff
@@ -19,13 +19,17 @@ def get
...
           def delete
-            feature = flipper[feature_name]
-            flipper.adapter.remove(feature)
-            json_response({}, 204)
+            if feature_names.include?(feature_name)
+              flipper.remove(feature_name)
+
+              json_response({}, 204)
+            else
+              json_response({}, 404)
```
jnunemaker: 🎉

PR #127, opened 2016-06-01
minitest shared tests

### LINKED ISSUE #55: Add shared spec for use in minitest
Same as rspec one but works for minitest.

AlexWheeler:
fixes #55 

First shot at minitest shared tests.  Definitely let me know if you had something else in mind.  Thought through some different ways to handle this, including making a sort of DSL like rspec's, but then decided that the whole reason of using minitest is to not do something like that.  This way we keep to the spirit of minitest with simple Ruby, as opposed to defining functions like `it_behaves_like`, etc.

This PR adds:
* shared test directory and shared_adapter_test
* redis adapter test
* memory adapter test
* mongo adapter test
*  pstore adapter test
* active_record adapter test

To make a shared test:
1. create test cases in file under `lib/flipper/test` (will automatically be loaded by test_helper)
2. prepend module to set up any state that test depends, must include `@adapter` instance variable that is an instance of the adapter


jnunemaker:
This is what I was thinking. Totally agree on avoiding DSL. 👍 

Why use prepend when we could use include? Include should work the same here and I've not really used prepend so I'm just curious.

> create test cases in file under test/shared/ (will automatically be loaded by test_helper)

I'd like this to live under lib/ as the shared adapter specs do. lib/ gets packaged with gems which means other gems that depend on flipper can use the shared specs. Putting this in lib/test/shared_adapter_tests.rb or whatever matches the shared spec path means that external adapters can use this file to ensure that they match the adapter behavior. Make sense?

test/adapters/redis_test.rb
```diff
@@ -0,0 +1,15 @@
+require 'test_helper'
+require 'flipper/adapters/redis'
+
+class RedisTest < MiniTest::Test
+  prepend SharedAdapterTests
+
+  def setup
+   options = {}
+   if ENV['BOXEN_REDIS_URL']
```
jnunemaker: I actually don't use boxen anymore so I'm ok with leaving this part out if you want. We can also leave it and just do a remove boxen artifacts PR later. Either way is fine by me!

AlexWheeler: *I actually don't use boxen anymore*
great, will remove it in this PR

*I'd like this to live under lib/ as the shared adapter specs do*
Awesome makes sense, will change that, thanks

*Why use prepend when we could use include? Include should work the same here and I've not really used prepend so I'm just curious.*

It comes down to where we want to call super.  I think it's nice to just expect someone using the shared_adapter_tests to only have to define a setup method that initializes an @adapter and effectively hide some of the implementation details.  I'll give an example of difference between `prepend` and `include`.

`include`
class RedisTest < MiniTest::Test
  include SharedAdapterTests

  def setup
   client = Redis.new({}).tap { |c| c.flushdb }
   @adapter = Flipper::Adapters::Redis.new(client)
    super
  end
end

using `include` will make RedisTest#setup take precedence over the module's `setup` so in this case minitest calls RedisTest.new.setup, which initializes @adapter and then calls super to look up the inheritance chain and call the module's setup.


`prepend` makes the prepended module's `setup` take precedence over RedisTest#setup, so when minitest calls RedisTest.new.setup, the module's `setup` will be called first, calling super which calls RedisTest#setup and initializes @adapter then hands back control to the module to run all the tests.

jnunemaker:
> tl;dr using prepend allows anyone using shared_adapter_tests not have to type super in setup

Ah :cool:. Fine by me then. I think ruby 1.9 support is ending anyway.

Gemfile
```diff
@@ -14,6 +14,7 @@ gem 'rspec', '~> 3.0'
 gem 'rack-test', '~> 0.6.3'
 gem 'sqlite3', '~> 1.3.11'
 gem 'rails', "~> #{ENV["RAILS_VERSION"] || '4.2.5'}"
+gem 'minitest', '~> 5.8.0'
```
AlexWheeler: test::unit is default testing framework for rails versions < 4.0, so tests were failing on those versions.  Is adding it to the Gemfile best way to handle this?

jnunemaker: Fine by me.

AlexWheeler:
@jnunemaker ported all the shared_example_specs to shared_adapter_test (MiniTest) and added Mongo, Redis, Memory, ActiveRecord, PStore tests that run against the shared_adapter_test

lib/flipper/test/shared_adapter_test.rb
```diff
@@ -0,0 +1,239 @@
...
+      :boolean => nil,
+      :groups => Set.new,
+      :actors => Set.new,
+      :percentage_of_actors => nil,
+      :percentage_of_time => nil,
+    }
+    assert_equal expected, @adapter.get(@feature)
+  end
+
+  def test_can_enable_and_get_value_for_boolean_gate
```
jnunemaker: Why the split on this test and the following one when compared to the shared_adapter_specs.rb? I think of these as more integration type specs so I like to see that each side of the coin works following each other more than that they work just in isolation.

lib/flipper/test/shared_adapter_test.rb
```diff
@@ -0,0 +1,239 @@
...
+    assert_equal true, @adapter.disable(@feature, @group_gate, @flipper.group(:early_access))
+    result = @adapter.get(@feature)
+    assert_equal Set['admins'], result[:groups]
+
+    assert_equal true, @adapter.disable(@feature, @group_gate, @flipper.group(:admins))
+    result = @adapter.get(@feature)
+    assert_equal Set.new, result[:groups]
+  end
+
+  def test_can_enable_disable_and_get_value_for_an_actor_gate
```
jnunemaker: This is missing the disable of asdf at the end as well.

lib/flipper/test/shared_adapter_test.rb
```diff
@@ -0,0 +1,239 @@
...
+    assert_equal '0', result[:percentage_of_actors]
+  end
+
+  def test_can_enable_percentage_of_actors_gate_many_times_and_consistently_return_values
+    (1..100).each do |percentage|
+      assert_equal true, @adapter.enable(@feature, @actors_gate, @flipper.actors(percentage))
+      result = @adapter.get(@feature)
+      assert_equal percentage.to_s, result[:percentage_of_actors]
+    end
+  end
```
jnunemaker: Missing the disabled spec that matches this named "can disable percentage of actors gate many times and consistently return values".

test/adapters/mongo_test.rb
```diff
@@ -0,0 +1,17 @@
...
+  prepend SharedAdapterTests
+
+  def setup
+    host = '127.0.0.1'
+    port = '27017'
+    collection = Mongo::Client.new(["#{host}:#{port}"], server_selection_timeout: 1, database: 'testing')['testing']
+    collection.drop
+    collection.create
+    @adapter = Flipper::Adapters::Mongo.new(collection)
+    collection.create
```
jnunemaker: This is extra and can be removed.

test/adapters/mongo_test.rb
```diff
@@ -0,0 +1,17 @@
...
+  rescue Mongo::Error::OperationFailure
```
jnunemaker: I'd prefer to just wrap this around the drop and/or the create. The main reason, which I might be wrong on, is if mongo isn't running then I think you'll end up with @adapter not being assigned which would cause no method errors instead of mongo errors (which point you at mongo not running). Maybe just try stopping mongo and running this spec and seeing what happens? If it fails with a decent error that would make me or someone else realize mongo isn't running then I'm fine with this. If it doesn't, then perhaps moving the rescue to where it was or something better should be done. Any thoughts or did you run into anything weird here?

jnunemaker:
Left a few notes. Overall this looks great. Thanks so much for the work. Just a few minor changes between the spec version and this one that I'd like to see fixed and a question on the mongo stuff. Let me know if you have any questions or strong feelings about any of my comments.

lib/flipper/test/shared_adapter_test.rb
AlexWheeler: great catch sorry for missing that.  Will double check that I've covered all of the tests

lib/flipper/test/shared_adapter_test.rb
AlexWheeler: Totally agree, will keep them in the same test

AlexWheeler:
Thanks for the great notes @jnunemaker.  I've addressed those and made sure both shared_specs and shared_tests match 1:1

Issue #96, opened 2015-12-21
Add core memcache adapter


jnunemaker:

nruth:
Would this provide data storage in memcached (as opposed to redis, active_record, etc), or a way to cache lookups across requests?

jnunemaker:
This would wrap another adapter and purely be a cache. We do this for github.com. It would be something similar to the memoize adapter, but for memcache.

nruth:
That would be really cool to have!

I've been thinking about trying something like that for an app we've added flipper active_record to. It seems like different gates and "things" could have different cache configurations (e.g. expiry) but I'm not sure if it'd be useful in practice. 

Seeing what you're using for such a busy site would be quite illuminating, or would probably just solve our problem, like the rest of the gem does. :+1:

jnunemaker:
As I mentioned, ours looks really similar to the memoizable one here:
https://github.com/jnunemaker/flipper/blob/52471278683a75633d4af29e3f952d3d0934e58b/lib/flipper/adapters/memoizable.rb

Just remove the if memoizing checks and you are pretty darn close.

nruth:
Thanks for the nudge in the right direction. I misunderstood similar as similar concept or functionality.

AlexWheeler:
This is great and I've been thinking a bit about it.  A wrapper for an adapter would work well, but looking at `Flipper::Adapters::Memoizable`'s initialize function it takes a `cache` argument.  What do you think of instead of making a wrapper for an adapter we made a Memcached  class that satisfies the cache interface.  Down the line if you want to use different caches you could still use the same memoizable middleware and class, but just make caches satisfy the cache interface.  Could even make shared cache tests

jnunemaker:
Memoizable should always be in memory and should be used regardless of any other outside caching you use. The goal of memoizable is to avoid repeated network calls for the same data. The goal of a memcached adapter is the same concept, but I don't think we want to merge them as a memcached adapter would be more about repeated network calls to a downstream storage system (aka mysql or whatever). I see it more like layers of caches. You want one in memory to prevent duplicate calls during a request and another do prevent duplicate calls (ie: memcached) in a time window to a slower, more complex data store (ie: mysql).

I don't think it makes sense to introduce a new concept like a cache. I actually really like that there is just one concept (adapter) and that you can adapt it to whatever you need from a performance standpoint. 

Thoughts?

jnunemaker:
[...]

PR #132, opened 2016-06-21
Dalli (memcached) adapter

AlexWheeler:
This is just initial thoughts on the memcached adapter #96 .  Wanted to get something up and see if this similar to what you were thinking.  Let me know, thanks!  I don't think we would need to make a middleware for this like the memoization middleware.  The user could just initialize their adapter with this adapter for example in a rails initializer when using rails.  Or is there a reason we would want another middleware?

for example
```

active_record_adapter = Flipper::Adapters::ActiveRecord.new
adapter = Flipper::Adapters::Memcached.new(active_record_adapter)
$flipper = Flipper.new(adapter)
```


lib/flipper/adapters/memcached.rb
```diff
@@ -0,0 +1,74 @@
+require 'dalli'
+
+module Flipper
+  module Adapters
+    class Memcached
```
jnunemaker: Since this is dalli and some apps use the memcached gem, I'd prefer to call this adapter dalli I think. Helps avoid confusing and allows for a memcached adapter down the road.

lib/flipper/adapters/memcached.rb
```diff
@@ -0,0 +1,74 @@
...
+      include Flipper::Adapter
+      FeaturesKey =  :flipper_features
+
+      attr_accessor :cache
+      attr_accessor :name
+
+      def initialize(adapter, address = 'localhost:11211', options = {})
```
jnunemaker: I'm kind of averse to taking parameters and then initializing an instance for people. The main problem I've ran into in the past is that these initialization options change and then I'm constantly updating or behind what is possible. Additionally, people usually have an instance already they are using in their app, so this forces multiple connections to be created. Instead I usually do something like this:

```ruby
def initialize(adapter, cache)
  @adapter = adapter
  @cache = cache
end
```

test/adapters/memcached_test.rb
```diff
@@ -0,0 +1,16 @@
+require 'test_helper'
+require 'flipper/adapters/memory'
+require 'flipper/adapters/memcached'
+
+class MemcachedTest < MiniTest::Test
```
jnunemaker: 👍  Looks good here.

lib/flipper/adapters/memcached.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      def disable(feature, gate, thing)
+        @adapter.disable(feature, gate, thing).tap do
+          @cache.delete(feature)
+        end
+      end
+
+      private
+
+      def cached(value)
```
jnunemaker: I'd drop this method and use Client#fetch in dalli directly in the methods above. get/set can be racey as last write wins whereas get/add is not.

lib/flipper/adapters/memcached.rb
```diff
@@ -0,0 +1,74 @@
...
+        end
+      end
+
+      def clear(feature)
+        @adapter.clear(feature).tap do
+          @cache.delete(feature)
+        end
+      end
+
+      def flush
```
jnunemaker: I'd drop this method. Since we can pass in an instance, the instance can be used to flush in tests. New methods per adapter for adapters that are used in production can cause tricky issues (like using the method in middleware, then removing that adapter and then that method is gone and errors).

lib/flipper/adapters/memcached.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+
+      def features
+        cached(FeaturesKey) do
+          @adapter.features
+        end
+      end
+
+      def add(feature)
+        @adapter.add(feature).tap do
```
jnunemaker: Probably just personal preference, but I'm not a fan of tap really. I think my thought is there is no way a block/yield is faster than a temporary variable like this:

```ruby
result = @adapter.add(feature)
@cache.delete(FeaturesKey)
result
```

I also think that it reads more explicitly for any programmer in any language as that is a common idiom, whereas tap might be known just by rubyists and what not. Not a huge concern there, but if things are more common, I don't mind if they are a bit less aesthetically beautiful. 

Any passionate thoughts or feelings on keeping tap?

lib/flipper/adapters/memcached.rb
jnunemaker: We'll probably want to provide a ttl option/argument for initialize that can be passed to fetch and perhaps default it to 0 or something.

jnunemaker:
> The user could just initialize their adapter with this adapter for example in a rails initializer when using rails. 

Correct. This is what we do at GitHub for our memcached adapter.

Left some notes, but it looks super close!

AlexWheeler:
Thanks a ton for the comments @jnunemaker.  They all were great and have been addressed.

lib/flipper/adapters/dalli.rb
```diff
@@ -0,0 +1,75 @@
...
+      #Internal
+      attr_accessor :cache
+
+      #Public: The name of the adapter.
+      attr_accessor :name
+
+      # Public
+      def initialize(adapter, cache, ttl = 0)
+        @adapter = adapter
+        @name = adapter.name
```
jnunemaker: I'm inclined to leave this as :dalli or :cached if we go that route. I think there is value in this adapter's name being different than the adapter it is wrapping so you can instrument them separately to see cached and uncached performance.

jnunemaker:
This looks great! So clean.

lib/flipper/adapters/dalli.rb
jnunemaker: Fixed by 8fd1a7d337abaeaa2c6bfcb14b4f97a5d60a8ccf

lib/flipper/adapters/dalli.rb
```diff
@@ -0,0 +1,75 @@
...
+module Flipper
+  module Adapters
+    # Internal: Adapter that wraps another adapter with the ability to cache
+    # adapter calls in Memcached using the Dalli gem.
+    class Dalli
+      include Flipper::Adapter
+      FeaturesKey = :flipper_features
+
+      #Internal
+      attr_accessor :cache
```
jnunemaker: I also made these readers instead of accessors.

Issue #140, opened 2016-06-25
Dalli Adapter documentation


AlexWheeler:
Add documentation for Dalli adapter 

`lib/flipper/adapters/dalli.rb`

PR #141, opened 2016-06-26
Dalli cache adapter optimization documentation

AlexWheeler:
#140 add documentation for Dalli adapter.  I think it makes most sense to put this under Optimization as opposed to adapters since it's really an optimization and meant to wrap a primary database adapter.  I decided to use the header `Cache Adapters` because we might have more adapters used just for caching in the future.  Curious to hear thoughts and if the example / wording could be improved

screen shot 2016-06-26 at 12 31 33 pm


AlexWheeler:
@jnunemaker let me know what you think about these docs when you get a chance

jnunemaker:
Looks good. Thanks!

Issue #129, opened 2016-06-08
Support for ActiveRecord 5


md5:
It looks like Rails 5.0.0 will be released relatively soon, so I wanted to open an issue regarding support for ActiveRecord 5.

Since the version range for `activerecord` is set to `">= 3.2", "< 5"`, Bundler will happily allow it to be installed with `5.0.0.rc1`, which is what I have in my application (`5.0.0.rc1` is considered `< 5` by RubyGems). So far, I haven't noticed any incompatibilities, but I haven't done extensive testing beyond basic feature flag changes.

md5:
I just ran across an issue that seems to be related to `flipper-active_record`. I had an initializer that looked like this to set up the memorizer middleware:

```ruby
# config/initializers/flipper.rb
module MyApplication
  def self.flipper
    @flipper ||= begin
      adapter = Flipper::Adapters::ActiveRecord.new

      Flipper.new(adapter)
    end
  end
end

require 'flipper/middleware/memoizer'
Rails.application.config.middleware.use Flipper::Middleware::Memoizer, -> { MyApplication.flipper }
```

I also have this initializer generated by `rails new`:

```ruby
# config/initializers/active_record_belongs_to_required_by_default.rb 

 Be sure to restart your server when you modify this file.

# Require `belongs_to` associations by default. This is a new Rails 5.0
# default, so it is introduced as a configuration option to ensure that apps
# made on earlier versions of Rails are not affected when upgrading.
Rails.application.config.active_record.belongs_to_required_by_default = true
```

At some point, I noticed that `ActiveRecord::Base.belongs_to_required_by_default` was no longer set to `true`. I did a `git bisect` and found that it started when I added `flipper-active_record` to my project.

After I changed my `Gemfile` to add `require: false` to my `gem 'flipper-active_record'` line and added an explicit `require 'flipper-active_record'` to my initializer above, I once again had `ActiveRecord::Base.belongs_to_required_by_default == true` as expected.

jnunemaker:
Thanks for opening the issue and testing. I don't think there should be anything big that would be incompatible, but we should definitely start testing. Step 1 is probably to get a rails 5 gemfile going that runs in travis ci. It doesn't make sense that flipper would affect initializers like you've mentioned. Seems like something else complex is going on there, but let me know if it is flipper.

md5:
[...]

PR #144, opened 2016-07-07
Adjust gemspec and test/ci for Rails 5

jnunemaker:
Now that Rails 5 is out the gemspec for the AR adapter needed to be adjusted. Additionally, I updated all the test/ci stuff to run against rails 5.


jnunemaker:
refs https://github.com/jnunemaker/flipper/issues/129

Issue #130, opened 2016-06-16
Add Flipper::Ap::V1 /features route


AlexWheeler:
Need to add `api/v1/features` to get all features

Some files to look at for reference to how everything works:
`lib/flipper/api/middleware.rb`
`lib/flipper/api/v1/actions/feature.rb`
`lib/flipper/api/v1/decorators/feature.rb`

PR #147, opened 2016-07-22
Api features endpoint

AlexWheeler:
addresses #130 

Having thoughts that might be better to return something like:

```
features = flipper.features.map { |feature|
  { feature: Decorators::Feature.new(feature).as_json }
}
json_response({ features: features })
```
so could add some non-breaking updates to response if wanted to (and add it to /feature), but then I also thought this works and not sure what else would want to add

i.e.
```
features = flipper.features.map {|feature|
  { feature: feature: Decorators::Feature.new(feature).as_json, some_other_data: "..." }
}
```

### DIFF
--- a/lib/flipper/api/middleware.rb
+++ b/lib/flipper/api/middleware.rb
@@ -7,7 +7,7 @@ Pathname(__FILE__).dirname.join('v1/actions').each_child(false) do |name|
 end
 
 module Flipper
-  module Api 
+  module Api
     class Middleware
       # Public: Initializes an instance of the API middleware.
       #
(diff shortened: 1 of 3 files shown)


jnunemaker:
I am a firm believer in hash responses. They are the most flexible for future additions. 

```ruby
features = flipper.features.map { |feature|
  Decorators::Feature.new(feature).as_json
}

json_response({
  features: features,
})
```

Doing this means we can add other keys and not break older clients (as we would with an array). Looks good to me other than that one tweak.

AlexWheeler:
great, should be good now, thanks @jnunemaker

jnunemaker:
Looks good. My only thought is someday it might be nice to have slimmed down version or slim by default and option to add gate information and stuff.

PR #148, opened 2016-07-22
override inspect on Rack::Builder for pretty rake routes output

### LINKED ISSUE #117: Lacking pretty output in `rake routes`
Hey All,

When I run ```rake routes``` the line for ```/flipper``` looks like this (you may need to scroll to the right)
```
                                                      /flipper                                                                  #<Rack::Builder:0x007fbbe44200c0 @warmup=nil, @run=#<Proc:0x007fbbe443b9d8@/Users/rob/.rvm/gems/ruby-2.2.3@borg/gems/flipper-ui-0.7.5/lib/flipper/ui.rb:31 (lambda)>, @map=nil, @use=[#<Proc:0x007fbbe4421d08@/Users/rob/.rvm/gems/ruby-2.2.3@borg/gems/rack-1.6.4/lib/rack/builder.rb:86>, #<Proc:0x007fbbe4400040@/Users/rob/.rvm/gems/ruby-2.2.3@borg/gems/rack-1.6.4/lib/rack/builder.rb:86>, #<Proc:0x007fbbe43f3f70@/Users/rob/.rvm/gems/ruby-2.2.3@borg/gems/rack-1.6.4/lib/rack/builder.rb:86>, #<Proc:0x007fbbe43f3ea8@/Users/rob/.rvm/gems/ruby-2.2.3@borg/gems/rack-1.6.4/lib/rack/builder.rb:86>, #<Proc:0x007fbbe43f3de0@/Users/rob/.rvm/gems/ruby-2.2.3@borg/gems/rack-1.6.4/lib/rack/builder.rb:86>]>
```

The route works fine when I hit it in my browser, so I'm wondering if anyone knows why this may be happening.

Thanks!

AlexWheeler:
Will fix #117 .  Not sure if this is the _best_ way, but works for now.  Reference #117 for conversation about why I think this works for now. Curious to hear what everyone thinks

With this change:

screen shot 2016-07-22 at 4 48 30 pm

### DIFF
--- a/lib/flipper/ui.rb
+++ b/lib/flipper/ui.rb
@@ -44,6 +44,8 @@ module Flipper
       builder.use Flipper::Middleware::Memoizer, flipper
       builder.use Middleware, flipper
       builder.run app
+      klass = self
+      builder.define_singleton_method(:inspect) { klass } # pretty rake routes output
       builder
     end
   end


lib/flipper/ui.rb
```diff
@@ -44,6 +44,8 @@ def self.app(flipper, options = {})
       builder.use Flipper::Middleware::Memoizer, flipper
       builder.use Middleware, flipper
       builder.run app
+      klass = self
+      builder.define_singleton_method(:inspect) { klass } # pretty rake routes output
```
AlexWheeler: rake routes calls `inspect`.  Line 47 just keeps a reference to the class `Flipper::UI` in case for some reason it changes in the future will still reference correct class and not just hard code 'Flipper::UI', then we just override inspect on the builder instance, which is the object rake routes sends the `inspect` message to

jnunemaker:
Not a huge fan of it, but I don't have a better idea right now. One thought was to use klass.inspect instead of just klass in for the inspect method. Another was to just return "Flipper::UI" instead of holding a reference to self in the klass temp var. I'm cool with merging this. We can always tweak later if we come up with a better idea.

PR #151, opened 2016-08-05
Make add_actor input placeholder lowercase

tommyp:
Today I was sitting with one of our QA people as they added an actor for their own user account. Due to the placeholder being `User` she initially add `User:123456` which didn't work. They then added `user:123456` which did. 

If I'm mistaken, feel free to close this but I think it does need to be lowercase and figured this was an easy win rather than calling `downcase` when this string is turned into an actual record.

### DIFF
--- a/lib/flipper/ui/views/add_actor.erb
+++ b/lib/flipper/ui/views/add_actor.erb
@@ -15,7 +15,7 @@
     <form action="<%= script_name %>/features/<%= @feature.key %>/actors" method="post">
       <%== csrf_input_tag %>
       <input type="hidden" name="operation" value="enable">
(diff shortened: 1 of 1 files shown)


jnunemaker:
Lowercase or uppercase is totally dependent on what you choose as your flipper id. I use class name and id, which means User:6 or something like that (with capitalized U) so I chose it as the example. Are you intentionally downcasing flipper id's? No problem with that (other than perhaps confusion in the UI). I'm just curious as to why if so.

jnunemaker:
Closing as stale, but happy to talk more when/if you have time to get back to this.

andyw8:
I opened #416 to suggest a change to this.

PR #152, opened 2016-08-16
POST /api/v1/features - add feature endpoint

AlexWheeler:
Looks like master is failing, #154 should fix it. Once merged I can rebase

* endpoint to add a feature

I decided to make the parameter 'name' since that matches the signature for `Flipper::DSL#feature`, which takes a name as an argument.

I decided to use 422 because RFC 4918 describes it as:

   The 422 (Unprocessable Entity) status code means the server
   understands the content type of the request entity (hence a
   415(Unsupported Media Type) status code is inappropriate), and the
   syntax of the request entity is correct (thus a 400 (Bad Request)
   status code is inappropriate) but was unable to process the contained
   instructions.  For example, this error condition may occur if an XML
   request body contains well-formed (i.e., syntactically correct), but
   semantically erroneous, XML instructions.

I am curious if you think we should return something in the response body for 200s.  Maybe the decorated feature? 

I like the idea of having errors be an array in case we want to add multiple errors.  I'm not sure what thoughts are on the shape of the error response, but having looked at Github's and Twitter's I like a few things from each.  I think including a link to an explanation in the documentation is really rad (as Github does), and I think the possibility of custom error codes that map to some type of error is cool and can be really helpful.  For now I figured we'll just keep it a hash with the message key, and down the line we could possibly add more keys like those if we think its a good idea without breaking anything.

I'm thinking in the next PR I can make the 422 error a child of Flipper::Api::Error and define a to_s or to_json method or something so we can reuse this in other endpoints where we expect a 422

https://developer.github.com/v3/#client-errors
https://dev.twitter.com/overview/api/response-codes

### DIFF
--- a/lib/flipper/api/v1/actions/features.rb
+++ b/lib/flipper/api/v1/actions/features.rb
@@ -19,6 +19,19 @@ module Flipper
               features: features
             })
           end
+
+          def post
+            feature_name = params.fetch('name') do
+              json_response({
+                errors: [{
+                  message: 'Missing post parameter: name',
+                }]
+              }, 422)
+            end
+
+            flipper.adapter.add(flipper[feature_name])
+            json_response({}, 200)
+          end
         end
       end
     end
--- a/spec/flipper/api/v1/actions/features_spec.rb
+++ b/spec/flipper/api/v1/actions/features_spec.rb
@@ -25,26 +25,26 @@ RSpec.describe Flipper::Api::V1::Actions::Features do
                   "name"=> "boolean",
                   "value" => true},
                   {
-                    "key" =>"groups",
-                    "name" => "group",
-                    "value" =>[],
-                  },
-                  {
-                    "key" => "actors",
-                    "name"=>"actor",
-                    "value"=>["10"],
-                  },
-                  {
-                    "key" => "percentage_of_actors",
-                    "name" => "percentage_of_actors",
-                    "value" => 0,
-                  },
-                  {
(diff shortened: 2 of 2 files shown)


lib/flipper/api/v1/actions/features.rb
```diff
@@ -19,6 +19,19 @@ def get
...
+            feature_name = params.fetch('name') do
+              json_response({
+                errors: [{
+                  message: 'Missing post parameter: name',
+                }]
+              }, 422)
+            end
+
+            flipper.adapter.add(flipper[feature_name])
+            json_response({}, 200)
```
jnunemaker: I think I like your idea of returning the decorated feature.

lib/flipper/api/v1/actions/features.rb
```diff
@@ -19,6 +19,19 @@ def get
               features: features
             })
           end
+
+          def post
+            feature_name = params.fetch('name') do
```
jnunemaker: Name works for me.

jnunemaker:
Totally forgot that this should probably not work if feature_creation_enabled is disabled. See the UI for more. https://github.com/jnunemaker/flipper/blob/27b59518849a0216cef9e93997ad65d7c7d7559e/lib/flipper/ui/actions/add_feature.rb#L11

jnunemaker:
Scratch that. That is a UI feature only. I suppose down the road we can have an API config option like that as well. Carry on!

PR #158, opened 2016-09-24
Feature Check Context

jnunemaker:
Sometimes when checking if a group should be enabled or not you just want a little extra context about the current check. This allows the group check to know which feature is being checked and what the gate values are for that feature. It allows for more advanced/crazy things. Should be backwards compatible, so if you don't need the context, just don't use it. I included [an example](https://github.com/jnunemaker/flipper/blob/ac2cb1f67f8442470ced00a8480a2a09d6021636/examples/group_dynamic_lookup.rb#L11-L16) for those that are curious.


lib/flipper/feature.rb
```diff
@@ -86,23 +86,17 @@ def remove
     def enabled?(thing = nil)
       instrument(:enabled?) { |payload|
         values = gate_values
-
         payload[:thing] = gate(:actor).wrap(thing) unless thing.nil?
-
         context = FeatureCheckContext.new(
           feature_name: @name,
           values: values
```
dakache: open_gate = gates.detect { |gate|
 -          gate.open?(thing, context)
 -        }
 -
 -        if open_gate.nil?
 -          false
 -        else
 +        if open_gate = gates.detect { |gate| gate.open?(thing, context) }
            payload[:gate_name] = open_gate.name
            true
 +        else
 +          false
          end
        }
      end

PR #160, opened 2016-09-28
display support for active_record adapter Rails 5

AlexWheeler:
Looks like its safe to say active_record adapter is working with Rails 5 since active_record.gemspec now has dependency at ">= 3.2", "< 6" and Travis is testing Rails5 and changelog says support as of Flipper-0.9.1 🎉


jnunemaker:
Ah yes. I missed that. Thanks!

Issue #165, opened 2016-10-05
Api V1 Group Gate


AlexWheeler:
Endpoint for enabling/disabling feature for a given group. 

`flipper[:feature].enable_group :group`

jnunemaker:
Thanks for adding this and the other API endpoint features. I dropped them on the [API project](https://github.com/jnunemaker/flipper/projects/2) as well to help keep track of stuff.

PR #166, opened 2016-10-05
Add Api V1 groups gate endpoint

AlexWheeler:
References #165 
Endpoint for enabling/disabling feature for a given group. I'm thinking we follow the naming convention with Boolean gate of using *route*/(enable | disable).

*api/v1/features/:feature_name/groups/:group_name/enable*
*api/v1/features/:feature_name/groups/:group_name/disable*

Given a feature *logging* and a group *admins*

`PUT api/v1/features/logging/groups/admins/enable # enables logging feature for admins group`
`PUT api/v1/features/logging/groups/admins/disable # disables logging feature for admins group`

I'm trying to think about how to handle errors:

it would be nice to know if you requested an unknown feature or non-registered group rather than just a generic 404.  Still haven't decided on an exception api for the api, but maybe:

{status: *http_code*, message: *error_mssage*}

i.e. {"status": 404,  message: "feature not found"}
      {"status": 404,  message: "group not registered"}

Also I'm checking if the group name is in `Flipper.groups` over rescuing *Flipper::GroupNotRegistered* since exception handling is slow. But it could be nice to rescue it since we get access to the Flipper Error messages such as "Group :group_name has not been registered", what do you think?

Also maybe this should be a post over put since its actually creating a Gate resource?

### DIFF
--- a/lib/flipper/api/v1/actions/groups_gate.rb
+++ b/lib/flipper/api/v1/actions/groups_gate.rb
@@ -1,0 +1,49 @@
+require 'flipper/api/action'
+require 'flipper/api/v1/decorators/feature'
+
+module Flipper
+  module Api
+    module V1
+      module Actions
+        class GroupsGate < Api::Action
+          route %r{api/v1/features/[^/]*/groups/?\Z}
+
+          def post
+            ensure_valid_params
+            feature = flipper[feature_name]
+            feature.enable_group(group_name)
+            decorated_feature = Decorators::Feature.new(feature)
+            json_response(decorated_feature.as_json, 200)
+          end
+
+          def delete
+            ensure_valid_params
+            feature = flipper[feature_name]
+            feature.disable_group(group_name)
+            decorated_feature = Decorators::Feature.new(feature)
+            json_response(decorated_feature.as_json, 200)
+          end
+
+          private
+
+          def ensure_valid_params
+            json_response({ code: 1, message: 'Feature not found.', more_info: '' }, 404) unless feature_names.include?(feature_name)
+            json_response({ code: 2, message: 'Group not registered.', more_info: '' }, 404) unless Flipper.group_exists?(group_name)
+          end
+
+          def feature_name
+            @feature_name ||= Rack::Utils.unescape(path_parts[-2])
+          end
+
+          def group_name
+            @group_name ||= params['name']
+          end
+
+          def feature_names
+            @feature_names ||= flipper.adapter.features
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/flipper/api/v1/actions/groups_gate_spec.rb
+++ b/spec/flipper/api/v1/actions/groups_gate_spec.rb
@@ -1,0 +1,75 @@
+require 'helper'
+
+RSpec.describe Flipper::Api::V1::Actions::GroupsGate do
+  let(:app) { build_api(flipper) }
+
+  describe 'enable' do
+    before do
+      flipper[:my_feature].disable
+      Flipper.register(:admins) do |actor|
+        actor.respond_to?(:admin?) && actor.admin?
+      end
+      post '/api/v1/features/my_feature/groups', { name: 'admins' }
+    end
+
+    it 'enables feature for group' do
+      person = double
+      allow(person).to receive(:flipper_id).and_return(1)
+      allow(person).to receive(:admin?).and_return(true)
+      expect(last_response.status).to eq(200)
+      expect(flipper[:my_feature].enabled?(person)).to be_truthy
+    end
+
+    it 'returns decorated feature with group enabled' do
+      group_gate = json_response['gates'].find { |m| m['name'] == 'group' }
+      expect(group_gate['value']).to eq(['admins'])
+    end
+  end
+
+  describe 'disable' do
+    before do
+      flipper[:my_feature].disable
+      Flipper.register(:admins) do |actor|
+        actor.respond_to?(:admin?) && actor.admin?
+      end
+      flipper[:my_feature].enable_group(:admins)
+      delete '/api/v1/features/my_feature/groups', { name: 'admins' }
+    end
+
+    it 'disables feature for group' do
+      person = double
+      allow(person).to receive(:flipper_id).and_return(1)
+      allow(person).to receive(:admin?).and_return(true)
+      expect(last_response.status).to eq(200)
+      expect(flipper[:my_feature].enabled?(person)).to be_falsey
+    end
+
+    it 'returns decorated feature with group disabled' do
+      group_gate = json_response['gates'].find { |m| m['name'] == 'group' }
+      expect(group_gate['value']).to eq([])
+    end
+  end
+
+  describe 'non-existent feature' do
+    before do
+      delete '/api/v1/features/my_feature/groups', { name:  'admins' }
+    end
+
+    it  '404s with correct error response when feature does not exist' do
+      expect(last_response.status).to eq(404)
+      expect(json_response).to eq({ 'code' => 1, 'message' => 'Feature not found.', 'more_info' => '' })
+    end
+  end
+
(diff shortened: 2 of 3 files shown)


lib/flipper/api/v1/actions/groups_gate.rb
```diff
@@ -0,0 +1,41 @@
...
+
+module Flipper
+  module Api
+    module V1
+      module Actions
+        class GroupsGate < Api::Action
+          route %r{api/v1/features/[^/]*/groups/[^/]*/(enable|disable)/?\Z}
+
+          def put
+            if feature_names.include?(feature_name) && Flipper.group_names.include?(group_name)
```
jnunemaker: Perhaps we could use `Flipper.group_exists?` instead of checking the collecting in `group_names`?

AlexWheeler:
@jnunemaker that's a great blog post thanks for sharing,

*like the approach suggested in this article. Perhaps we have an ERROR_CODES constant somewhere with the key being a unique code (integer) and the value being a message (string)?*

agreed, will play with this idea

*What do you think api/v1/features/:feature_name/groups/:group_name and api/v1/features/:feature_name/groups should return? *

I was originally thinking that we would add GET routes as you mention that would map `:feature_name/groups` to `feature.enabled_groups` and `groups/group_name`  to some data about that group, but you're totally right this is for sure unnecessary.

I really like option 1:
`POST api/v1/features/:feature_name/groups`
`DELETE api/v1/features/:feature_name/groups`

For gates we can use the gate name as the resource name and the http verb to enable/disbale and any request parameters that are needed i.. name or a percentage / time.

`POST/DELETE api/v1/features/:feature_name/:gate_name`

Boolean gate would become

`POST api/v1/features/:feature_name/boolean  enable`
`DELETE api/v1/features/:feature_name/boolean # disable`

boolean, percentage_of_time are singular, while percentage_of_actors, actors, and groups are plural, but I think this is okay and pretty intuitive.

Does that sound good to you?

lib/flipper/api/v1/actions/groups_gate.rb
AlexWheeler: definitely can, missed that method thanks

jnunemaker:
> Does that sound good to you?

🤘  That sounds great.

lib/flipper/api/v1/actions/groups_gate.rb
```diff
@@ -0,0 +1,49 @@
...
+            decorated_feature = Decorators::Feature.new(feature)
+            json_response(decorated_feature.as_json, 200)
+          end
+
+          private
+
+          def ensure_valid_params
+            json_response({ code: 100, message: 'Feature not found.', more_info: '' }, 404) unless feature_names.include?(feature_name)
+            json_response({ code: 200, message: 'Gate not registered.', more_info: '' }, 404) unless Flipper.group_exists?(group_name)
+          end
```
AlexWheeler: I figure for this PR this is the most straightforward way to handle this for now since its pretty obvious what the code is doing

AlexWheeler:
@jnunemaker I added the error responses for when
1. feature is not found
2. gate is not registered

I also update the POST and DELETE  (for enabling/disabling) to return the decorated feature and a 200 status.  I think this is a good convention going forward.  I'll submit PR for boolean gate to make sure we're doing this there as well if you're into that idea.

I think it makes sense to submit a separate PR for standardizing the error responses, and I like the ideas in that article of having:

code, message, and more_info
more_info will be a url to the documentation on that error.  Some apis like Twilio do this and I find it super useful as a consumer of the api.

lib/flipper/api/v1/actions/groups_gate.rb
```diff
@@ -0,0 +1,49 @@
...
+require 'flipper/api/v1/decorators/feature'
+
+module Flipper
+  module Api
+    module V1
+      module Actions
+        class GroupsGate < Api::Action
+          route %r{api/v1/features/[^/]*/groups/?\Z}
+
+          def post
```
jnunemaker: This reads really nice. Great work.

lib/flipper/api/v1/actions/groups_gate.rb
```diff
@@ -0,0 +1,49 @@
...
+            feature = flipper[feature_name]
+            feature.disable_group(group_name)
+            decorated_feature = Decorators::Feature.new(feature)
+            json_response(decorated_feature.as_json, 200)
+          end
+
+          private
+
+          def ensure_valid_params
+            json_response({ code: 100, message: 'Feature not found.', more_info: '' }, 404) unless feature_names.include?(feature_name)
```
jnunemaker: Any particular reason for 100 and 200 as the codes? I ask because I would perhaps lean toward just starting with 1 and 2. Having a 3 digit number feels to close to an http status perhaps.

lib/flipper/api/v1/actions/groups_gate.rb
```diff
@@ -0,0 +1,49 @@
...
+            json_response({ code: 200, message: 'Gate not registered.', more_info: '' }, 404) unless Flipper.group_exists?(group_name)
```
jnunemaker: I believe you mean "Group" here, not "Gate" since you are checking if the group does not exist.

spec/flipper/api/v1/actions/groups_gate_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    it 'enables feature for group' do
+      person = double
+      allow(person).to receive(:flipper_id).and_return(1)
+      allow(person).to receive(:admin?).and_return(true)
+      expect(last_response.status).to eq(200)
+      expect(flipper[:my_feature].enabled?(person)).to be_truthy
+    end
+
+    it ' returns decorated feature with group enabled' do
```
jnunemaker: Extra space here between `'` and `returns`.

spec/flipper/api/v1/actions/groups_gate_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  describe 'group not registered' do
+    before do
+      flipper[:my_feature].disable
+      delete '/api/v1/features/my_feature/groups', { name:  'admins' }
+    end
+
+    it '404s with correct error response when group not registered' do
+      expect(last_response.status).to eq(404)
+      expect(json_response).to eq({ 'code' => 200, 'message' => 'Gate not registered.', 'more_info' => '' })
```
jnunemaker: Group here as well I think.

jnunemaker:
Oh, I was also going to say, and you certainly don't need to do this for this PR, but it might be nice to have something like this to allow for re-using the error code errors amongst API classes.

```ruby
# somewhere in flipper/api in its own file?
class ErrorCode
  attr_reader :code, :message, :info

  def initialize(code, message, info = "")
    @code = code
    @message = message
    @info = info
  end

  def as_json
    {
      code: @code,
      message: @message,
      more_info: @more_info,
    }
  end
end

# somewhere in Flipper::Api?
ERROR_CODES = {
  feature_not_found: ErrorCode.new(1, "Feature not found.", "... link to docs ..."),
  gate_not_registered: ErrorCode.new(1, "Group not registered.", "... link to docs ..."),
}

# usage in this spot would look like this maybe?
json_response(ERROR_CODES[:feature_not_found].as_json, 404)
```

lib/flipper/api/v1/actions/groups_gate.rb
AlexWheeler: you're totally right I had same thoughts about it looking like an http status code. I was thinking it would be nice to have a similar concept to http status code meaning where 100 level codes could be for problem with feature, 200 level error could be problem with gate, etc. but I like the idea of keeping it simple

AlexWheeler:
Thanks for the feedback, yes I'm definitely thinking of the best way to make the errors reusable. I'll get a PR up for an idea i have asap

PR #167, opened 2016-10-05
Updates active_record adaptor generator docs

cosmocochrane1:
*Clearly defines what tables are generated by default, making it more obvious what the generator is actually doing which I think leads to an easier onboarding process for new flipper users


jnunemaker:
Thanks! Always appreciate people taking the time to make flipper better for new users.

PR #171, opened 2016-10-07
do not use mass assignment anymore for active record adapter

### LINKED ISSUE #161: Can't mass-assign protected attributes: key
```Can't mass-assign protected attributes: key``` when creating a new feature with flipper ui and Flipper::Adapters::ActiveRecord. Adding ```attr_accessor :key``` to the adapter workaround the problem.

### PR DESCRIPTION
**Pull request by AlexWheeler**
fixes #161 
* rails 3 does not allow mass assignment.  This makes sure the active record adapter does not mass assign attributes

### DIFF
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -45,10 +45,11 @@ module Flipper
 
       # Public: Adds a feature to the set of known features.
       def add(feature)
-        attributes = {key: feature.key}
         # race condition, but add is only used by enable/disable which happen
         # super rarely, so it shouldn't matter in practice
-        @feature_class.where(attributes).first || @feature_class.create!(attributes)
+        unless @feature_class.where(key: feature.key).first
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/flipper/adapters/active_record.rb [hunk 1]
```diff
@@ -45,10 +45,11 @@ def features
 
       # Public: Adds a feature to the set of known features.
       def add(feature)
-        attributes = {key: feature.key}
         # race condition, but add is only used by enable/disable which happen
         # super rarely, so it shouldn't matter in practice
-        @feature_class.where(attributes).first || @feature_class.create!(attributes)
+        unless @feature_class.find_by_key(feature.key)
```
jnunemaker: I think dynamic finders are [deprecated](https://github.com/rails/activerecord-deprecated_finders) `find_by_key` which is why I went with the less aesthetically pleasing `where(...).first`. Would you mind swapping this back to the where so it is more future proof?

PR #173, opened 2016-10-11
add flipper-api documentation

stve:
I was looking at using the `flipper-api` but found that it lacked documentation so while documenting it for my own purposes, I decided it'd be better to contribute upstream. I've documented all of the API endpoints that currently exist.

Note: I wasn't sure if a `Usage` section was necessary or not as it'd be virtually the same as that of `flipper-ui` in terms of mounting, etc. so I left that out for now. But it is probably worth noting in some way as the `/flipper-api` path used in the documentation examples assumes a mount point so I'm curious what your thoughts are on that.

Thanks!

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -1,0 +1,190 @@
+# Flipper::Api
+
+API for the [Flipper](https://github.com/jnunemaker/flipper) gem.
+
+## Installation
+
+Add this line to your application's Gemfile:
+
+    gem 'flipper-api'
+
+And then execute:
+
+    $ bundle
+
+Or install it yourself as:
+
+    $ gem install flipper-api
+
+## Endpoints
+
+### Get all features
+
+`GET /flipper-api/api/v1/features`
+
+**Response**
+
+Returns an array of feature objects:
+
+```json
+{
+  "features": [
+    {
+      "key": "search",
+      "state": "on",
+      "gates": [
+        {
+          "key": "boolean",
+          "name": "boolean",
+          "value": false
+        },
+        {
+          "key": "groups",
+          "name": "group",
+          "value": []
+        },
+        {
+          "key": "actors",
+          "name": "actor",
+          "value": []
+        },
+        {
+          "key": "percentage_of_actors",
+          "name": "percentage_of_actors",
+          "value": 0
+        },
+        {
+          "key": "percentage_of_time",
+          "name": "percentage_of_time",
+          "value": 0
+        }
+      ]
+    },
+    {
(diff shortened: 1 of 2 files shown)


docs/api/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+Or install it yourself as:
+
+    $ gem install flipper-api
+
+## Endpoints
+
+### Get all features
+
+`GET /flipper-api/api/v1/features`
```
jnunemaker: I see what you mean about the mount point. I think I would remove the mount point here and just document `/api/v1/features`. Perhaps a section on how to mount it and how that affects these paths between the installation and endpoints? What do you think?

jnunemaker: One thing I love to do when documenting API's is include a curl command. For example, here is what I did for Gauges way back in the day: http://get.gaug.es/documentation/reference-listing/your-information/.

gaugesdocs

What do you think about that? Path, Request (as curl) and Response? Does that seem like it would work here or does it spur any other ideas?

stve: 👍  I really like this structure. I left the `URL` section and added cURL examples for each endpoint.

stve:
@jnunemaker thanks for your feedback. I've added some documentation on mounting and pointed to the `Flipper::UI` usage section as it's really comprehensive. I also modified the URLs and examples.

As more endpoints are added, do you think having this on a single page will be acceptable or is it worth splitting out to separate pages per endpoint?

jnunemaker:
> As more endpoints are added, do you think having this on a single page will be acceptable or is it worth splitting out to separate pages per endpoint?

I usually lean toward a single page until it hurts.

docs/api/README.md
```diff
@@ -0,0 +1,253 @@
...
+`Flipper::Api` is a mountable application that can be included in your Rails/Ruby apps. In a Rails application, you can mount `Flipper::Api` to a route of your choice:
+
+```ruby
+# config/routes.rb
+YourRailsApp::Application.routes.draw do
+  mount Flipper::Api.app(flipper) => '/flipper-api'
+end
+```
+
+For more advanced mounting techniques and for suggestions on how to mount in a non-Rails application, it is recommend that you review the [`Flipper::UI` usage documentation](https://github.com/jnunemaker/flipper/blob/master/docs/ui/README.md#usage) as the same approaches apply to `Flipper::Api`.
```
jnunemaker: I like linking to UI. Someday maybe we can just have a mounting.md file that docs both of them, but this works for me!

PR #175, opened 2016-10-12
changes boolean gate route to api/v1/features/boolean

AlexWheeler:
* this is convention we're going for with api gates

`POST api/v1/features/:feature_name/:gate_name # enable`

`DELETE api/v1/features/:feature_name/:gate_name # disable`

discussion about this can be found at #166


jnunemaker:
:metal: Really nice. I think this is going to work well.

PR #178, opened 2016-10-21
Makes Api V1 error responses reusable and standardized

AlexWheeler:
* moves error codes / as_json logic into descriptive module and provides `json_error_response` method to all actions

* `json_error_response` is very clear you're rendering an error as opposed to a successful response

* adds specs

* updates Groups Gate endpoint to use this error response format

we get to take advantage of Ruby's `fetch` so a KeyError will be raised if pass in an invalid error key, and also I like this approach because we get an api on top of directly accessing the hash, just use json_error_response(:error_key, status) and everything just works

### DIFF
--- a/lib/flipper/api/action.rb
+++ b/lib/flipper/api/action.rb
@@ -1,5 +1,6 @@
 require 'forwardable'
 require 'flipper/api/error'
+require 'flipper/api/error_response'
 require 'json'
 
 module Flipper
@@ -88,11 +89,27 @@ module Flipper
         throw :halt, response
       end
 
+      # Public: Call this with a json serializable object (i.e. Hash)
+      # to serialize object and respond to request
+      #
+      # object - json serializable object
+      # status - http status code
+
       def json_response(object, status = 200)
         header 'Content-Type', Api::CONTENT_TYPE
         status(status)
         body = JSON.dump(object)
         halt [@code, @headers, [body]]
+      end
+
+      # Public: Call this with an ErrorResponse::ERRORS key to respond
+      # with the serialized error object as response body
+      #
+      # error_key - key to lookup error object
+
+      def json_error_response(error_key)
+        error = ErrorResponse::ERRORS.fetch(error_key.to_sym)
+        json_response(error.as_json, error.http_status)
       end
 
       # Public: Set the status code for the response.
--- a/spec/flipper/api/action_spec.rb
+++ b/spec/flipper/api/action_spec.rb
@@ -25,35 +25,84 @@ RSpec.describe Flipper::Api::Action do
     end
   }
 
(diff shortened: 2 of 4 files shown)


spec/flipper/api/action_spec.rb
```diff
@@ -25,35 +25,72 @@ def delete
...
+  describe '#json_error_response' do
+    describe ":feature_not_found" do
+
+      it 'locates and serializes error correctly' do
+        action = action_subclass.new({}, {})
+        response = catch(:halt) do
+          action.json_error_response(:feature_not_found, 404)
+        end
+        expect(response).to eq([404, {"Content-Type"=>"application/json"}, ["{\"code\":1,\"message\":\"Feature not found.\",\"more_info\":\"\"}"]])
+      end
```
AlexWheeler: wasn't quite sure if this is good way to test these, but I wanted to make sure we could test all the expected functionality of render_error_json, let me know if you'd rather test this another way

lib/flipper/api/action.rb
```diff
@@ -88,13 +89,30 @@ def halt(response)
...
         halt [@code, @headers, [body]]
       end
 
+      # Public: Call this with an ErrorResponse::ERRORS key to respond
+      # with the serialized error object as response body
+      #
+      # error_key - key to lookup error object
+      # status - http status code
+
+      def json_error_response(error_key, status)
```
jnunemaker: Makes me wonder if the same error key will always have the same http status and if we could/should move that to ErrorResponse. 🤔  Curious what your thoughts are. I'm ok with not moving.

lib/flipper/api/action.rb
```diff
@@ -88,13 +89,30 @@ def halt(response)
...
+        error = ErrorResponse::ERRORS.fetch(error_key.to_sym)
```
jnunemaker: 😍  I love fetch.

spec/flipper/api/action_spec.rb
jnunemaker: I am ok with this. Thanks for adding the tests! 

If I were to change anything, I might assert each part separately. I find separate assertions easier as they more directly point to the assertion failure. Instead of searching through a big object for a differences you can just look at simple assertions.

```ruby
status, headers, body = [...]
body_hash = JSON.parse(body[0])
assert_equal 404, status
assert_equal "application/json", headers["Content-Type"]
assert_equal 1, body_hash["code"]
assert_equal "Feature not found.", body_hash["message"]
assert_equal "", body_hash["more_info"]
```

lib/flipper/api/action.rb
AlexWheeler: You're right, let's keep it simple for now, not really sure of a situation where it wouldn't be consistent anyways

PR #179, opened 2016-10-22
Api V1 percentage_of_actors gate endpoint

AlexWheeler:
* POST / DELETE request to enable / disable

* *send parameter 'percentage' to request percentage to enable 

* only POST requires a percentage parameter to be passed with request

* percentage_invalid raised and returned if percentage param is missing or not <= 100.  I like the name because it keeps with the convention:

:{ thing } _ { error_description }
i.e.
:{feature} _ {not_found}
:{group} _ {not_registered}
:{percentage } _ {invalid}

* uses regex to make sure percentage parameter is a number <= 100.  This is powerful because it handles all the edge cases, and we can keep it in the ensure_valid_post_params method

1. parameter is missing
2. parameter is empty string or nil
3. parameter is out of range

* adds specs

### DIFF
--- a/lib/flipper/api/error_response.rb
+++ b/lib/flipper/api/error_response.rb
@@ -23,6 +23,7 @@ module Flipper
       ERRORS = {
         feature_not_found: Error.new(1, "Feature not found.", "", 404),
         group_not_registered: Error.new(2, "Group not registered.", "", 404),
+        percentage_invalid: Error.new(3, "Percentage must be a positive number less than or equal to 100.", "", 400),
       }
     end
   end
--- a/lib/flipper/api/v1/actions/percentage_of_actors_gate.rb
+++ b/lib/flipper/api/v1/actions/percentage_of_actors_gate.rb
@@ -1,0 +1,57 @@
+require 'flipper/api/action'
+require 'flipper/api/v1/decorators/feature'
+
+module Flipper
+  module Api
+    module V1
+      module Actions
+        class PercentageOfActorsGate < Api::Action
+          route %r{api/v1/features/[^/]*/percentage_of_actors/?\Z}
+
+          def post
+            ensure_valid_enable_params
+            feature = flipper[feature_name]
+            feature.enable_percentage_of_actors(percentage)
+            decorated_feature = Decorators::Feature.new(feature)
+            json_response(decorated_feature.as_json, 200)
+          end
+
+          def delete
+            ensure_valid_disable_params
+            feature = flipper[feature_name]
+            feature.disable_percentage_of_actors
+            decorated_feature = Decorators::Feature.new(feature)
+            json_response(decorated_feature.as_json, 200)
+          end
+
+          private
+
+          def ensure_valid_enable_params
+            json_error_response(:feature_not_found) unless feature_names.include?(feature_name)
+            json_error_response(:percentage_invalid) unless percentage >= 0 && percentage <= 100
+          end
+
+          def ensure_valid_disable_params
+            json_error_response(:feature_not_found) unless feature_names.include?(feature_name)
+          end
+
+          def feature_name
+            @feature_name ||= Rack::Utils.unescape(path_parts[-2])
+          end
+
+          def percentage
+            @percentage ||= begin
+                              Integer(params['percentage'])
+                            rescue ArgumentError, TypeError
+                              -1
+                            end
+          end
+
+            def feature_names
+              @feature_names ||= flipper.adapter.features
+            end
+          end
+        end
+      end
+    end
+  end
--- a/spec/flipper/api/v1/actions/percentage_of_actors_gate_spec.rb
+++ b/spec/flipper/api/v1/actions/percentage_of_actors_gate_spec.rb
@@ -1,0 +1,85 @@
+require 'helper'
+
+RSpec.describe Flipper::Api::V1::Actions::PercentageOfActorsGate do
+  let(:app) { build_api(flipper) }
+
+  describe 'enable' do
+    before do
+      flipper[:my_feature].disable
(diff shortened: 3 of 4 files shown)


lib/flipper/api/v1/actions/percentage_of_actors_gate.rb
```diff
@@ -0,0 +1,53 @@
...
+          def ensure_valid_disable_params
+            json_error_response(:feature_not_found) unless feature_names.include?(feature_name)
+          end
+
+          def feature_name
+            @feature_name ||= Rack::Utils.unescape(path_parts[-2])
+          end
+
+          def percentage
+            @percentage ||= params.fetch('percentage') { json_error_response(:percentage_invalid) }.to_i
```
AlexWheeler: perhaps we want to pass the percentage into a regex to make sure it is an integer?

since 

"percentage": "alex" would also end up with 0

lib/flipper/api/v1/actions/percentage_of_actors_gate.rb
```diff
@@ -0,0 +1,57 @@
...
+          def percentage
+            @percentage ||= params['percentage'].to_i
+          end
+
+          def feature_names
+            @feature_names ||= flipper.adapter.features
+          end
+
+          def number_less_than_or_equal_100?(val)
+            !!/^(100|[1-9][0-9]?)$/.match(val)
```
jnunemaker: `^` and `$` only match [beginning of line](https://ruby-doc.org/core-2.1.1/Regexp.html#class-Regexp-label-Anchors). What do you think about using `\A` and `\Z` to match beginning and end of string? 

Alternatively, did you consider using the stricter integer conversion ruby has (`Integer(val)`)? I've never used a regex for this, so I am just curious what the difference would be between `Integer(val)` with a range check and a rescue for `ArgumentError` vs the regex. 

```
irb(main):002:0> Integer("1")
=> 1
irb(main):004:0> Integer("asdf")
ArgumentError: invalid value for Integer(): "asdf"
	from (irb):4:in `Integer'
	from (irb):4
	from /Users/jnunemaker/.rbenv/versions/2.3.1/bin/irb:11:in `<main>'
```

Then, once converted to an integer, I check the if the value is in a valid range. What do you think

spec/flipper/api/v1/actions/percentage_of_actors_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    it  '404s with correct error response when feature does not exist' do
+      expect(last_response.status).to eq(404)
+      expect(json_response).to eq({ 'code' => 1, 'message' => 'Feature not found.', 'more_info' => '' })
+    end
+  end
+
+  describe 'invalid percentage parameter' do
+    before do
+      flipper[:my_feature].disable
+      post '/api/v1/features/my_feature/percentage_of_actors', { percentage: '300' }
```
jnunemaker: This is a good out of range test. Does it feel worthwhile to add an invalid value as well, like "asdf"?

lib/flipper/api/v1/actions/percentage_of_actors_gate.rb
jnunemaker: Technically you can enable for 0 percent using flipper and the flipper-ui. Do you feel we should support the same semantics for the API or just point people to disable?

AlexWheeler: correct, I think it makes sense to support same semantics especially if users are used to that

lib/flipper/api/v1/actions/percentage_of_actors_gate.rb
```diff
@@ -0,0 +1,57 @@
...
+          end
+
+          def feature_name
+            @feature_name ||= Rack::Utils.unescape(path_parts[-2])
+          end
+
+          def percentage
+            @percentage ||= begin
+                              Integer(params['percentage'])
+                            rescue ArgumentError, TypeError
```
AlexWheeler: I like this much better great suggestion.  Here we rescue TypeError as well for nil cases Integer(nil)

spec/flipper/api/v1/actions/percentage_of_actors_gate_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  describe 'percentage parameter not an integer' do
+    before do
+      flipper[:my_feature].disable
+      post '/api/v1/features/my_feature/percentage_of_actors', { percentage: 'foo' }
+    end
+
+    it '400s with correct error response when percentage parameter is invalid' do
+      expect(last_response.status).to eq(400)
+      expect(json_response).to eq({ 'code' => 3, 'message' => 'Percentage must be a positive number less than or equal to 100.', 'more_info' => '' })
+    end
```
AlexWheeler: added test for invalid value that is not an integer or nil

lib/flipper/api/error_response.rb
```diff
@@ -23,6 +23,7 @@ def as_json
       ERRORS = {
         feature_not_found: Error.new(1, "Feature not found.", "", 404),
         group_not_registered: Error.new(2, "Group not registered.", "", 404),
+        percentage_invalid: Error.new(3, "Percentage must be a positive number less than or equal to 100.", "", 400),
```
jnunemaker: Missed this in my review. Should this be 422? 400 is bad request and that isn't accurate as if it gets here it was a well formed request, it just wasn't valid. I think 422 is what rails typically uses for this type of error. Thoughts?

spec/flipper/api/v1/actions/percentage_of_actors_gate_spec.rb
jnunemaker: Good stuff!

jnunemaker:
🎉

Issue #168, opened 2016-10-05
Api V1 percentage_of_actors gate


AlexWheeler:
endpoint to enable/disable percentage_of_actors for a feature following the convention:
`POST api/v1/features/:feature_name/:gate_name # enable`
`DELETE api/v1/features/:feature_name/:gate_name # disable`

api/v1/features/logging/percentage_of_actors
`flipper[:feature].enable_percentage_of_actors 10`

jnunemaker:
Closed by https://github.com/jnunemaker/flipper/pull/179 I believe.

PR #180, opened 2016-10-25
add api v1 percentage_of_time endpoint

AlexWheeler:
* POST / to enable  with percentage parameter
* DELETE to disable

re: discussion in #168, after this I'll get a PR up for updating status code for percentage_invalid from 400 -> 422 for the ErrorResponse and the endpoints specs because it is more fitting

### DIFF
--- a/spec/flipper/api/v1/actions/percentage_of_time_gate_spec.rb
+++ b/spec/flipper/api/v1/actions/percentage_of_time_gate_spec.rb
@@ -1,29 +1,28 @@
 require 'helper'
 
-RSpec.describe Flipper::Api::V1::Actions::PercentageOfActorsGate do
+RSpec.describe Flipper::Api::V1::Actions::PercentageOfTimeGate do
   let(:app) { build_api(flipper) }
 
   describe 'enable' do
     before do
       flipper[:my_feature].disable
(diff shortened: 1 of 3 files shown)


spec/flipper/api/v1/actions/percentage_of_time_gate_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      expect(flipper[:my_feature].enabled_gate_names).to include(:percentage_of_time)
+    end
+
+    it 'returns decorated feature with gate enabled for 5% of time' do
+      gate = json_response['gates'].find { |gate| gate['name'] == 'percentage_of_time' }
+      expect(gate['value']).to eq(10)
+    end
+  end
+
+  describe 'disable' do before do
```
jnunemaker: Odd formatting here. I'd drop before down on the next line.

spec/flipper/api/v1/actions/percentage_of_time_gate_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      flipper[:my_feature].disable
```
jnunemaker: Should this be enable since we are testing disable?

Issue #169, opened 2016-10-05
Api V1 actor gate


AlexWheeler:
endpoint to enable/disable feature for given actor following the convention
`POST api/v1/features/:feature_name/:gate_name # enable`
`DELETE api/v1/features/:feature_name/:gate_name # disable`

i.e.
api/v1/features/logging/actors
`flipper[:feature].enable_actor actor`

PR #181, opened 2016-10-28
add api v1 actors gate endpoint

AlexWheeler:
completes #169 
* POST / DELETE to enable disable with 'flipper_id' parameter
* adds specs
* adds Fliper::Api::Actor shim just like UI

Returns error if flipper_id parameter is missing or nil.  Flipper allows enabling an actor with a nil flipper_id, and I thought about allowing this and only rendering error on missing flipper_id parameter, but the more I think about it I don't think its ideal for a user to do this.  A flipper id should ideally be unique.  It opens the door for accidentally enabling a feature for anything with a nil flipper_id and users wanting to do this are better off using the group gate right?  Let me know your thoughts

Figured its easier to review small PRs so will update invalid error responses to 422, or can do it in this PR as a separate commit if you'd like

### DIFF
--- a/lib/flipper/api/error_response.rb
+++ b/lib/flipper/api/error_response.rb
@@ -24,6 +24,7 @@ module Flipper
         feature_not_found: Error.new(1, "Feature not found.", "", 404),
         group_not_registered: Error.new(2, "Group not registered.", "", 404),
         percentage_invalid: Error.new(3, "Percentage must be a positive number less than or equal to 100.", "", 400),
+        flipper_id_invalid: Error.new(4, "Required parameter flipper_id is missing.", "", 400),
       }
     end
   end
--- a/lib/flipper/api.rb
+++ b/lib/flipper/api.rb
@@ -1,6 +1,7 @@
 require 'rack'
 require 'flipper'
 require 'flipper/api/middleware'
+require 'flipper/api/actor'
 
 module Flipper
   module Api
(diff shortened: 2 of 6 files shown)


jnunemaker:
>  I don't think its ideal for a user to do this

Agreed. We should probably plug that in flipper as well if it does actually allow that. It was not intentional.

> Figured its easier to review small PRs so will update invalid error responses to 422, or can do it in this PR as a separate commit if you'd like

Sounds good to me. I'll look this over now quick.

lib/flipper/api/error_response.rb
```diff
@@ -24,6 +24,7 @@ def as_json
         feature_not_found: Error.new(1, "Feature not found.", "", 404),
         group_not_registered: Error.new(2, "Group not registered.", "", 404),
         percentage_invalid: Error.new(3, "Percentage must be a positive number less than or equal to 100.", "", 400),
+        flipper_id_invalid: Error.new(4, "Required parameter flipper_id cannot be nil.", "", 400),
```
jnunemaker: What do you think about "Required parameter flipper_id is missing."? I'm not sure that we need to mention nil.

AlexWheeler: yes, I've gone back and forth on this. I was thinking that saying its missing sounds like the key is missing, but definitely see how nil doesn't really translate to JSON.  I like your suggestion as it being missing could also be interpreted as the key or value is missing. Will update thanks!

PR #183, opened 2016-11-01
use json_error_response for /feature /features

AlexWheeler:
/feature, /features endpoints were built before ErrorResponse so now instead of `json_response({}, 404)` we can use `json_error_response(:feature_not_found)`.  With this change any api error handling code uses json_error_response.

POST /features will now return a 200 response (was 204) with the decorated feature as the body.  I think this is more useful for any clients and is similar to what we're doing for enabling / disabling gates

DELETE /features still returns a 204, was thinking it might be nice to return the same response as GET /features, but it might be unnecessary to query/serialize all that every time you just want to delete something

Adds new ErrorResponse `:name_invalid`.  Thought about calling it `:feature_name_invalid`, but I figure `:name_invalid` is generic enough where it can be used for some other endpoints in the future if have an invalid name param.

*reformats expected_response in features_spec

### DIFF
--- a/spec/flipper/api/v1/actions/features_spec.rb
+++ b/spec/flipper/api/v1/actions/features_spec.rb
@@ -17,22 +17,23 @@ RSpec.describe Flipper::Api::V1::Actions::Features do
         expected_response = {
           "features" => [
             {
-              "key" =>"my_feature",
+              "key" => "my_feature",
               "state" => "on",
               "gates" => [
                 {
                   "key"=> "boolean",
                   "name"=> "boolean",
-                  "value" => true},
-                  {
-                  "key" =>"groups",
+                  "value" => true
+                },
+                {
+                  "key" => "groups",
                   "name" => "group",
-                  "value" =>[],
+                  "value" => [],
                 },
                 {
                   "key" => "actors",
-                  "name"=>"actor",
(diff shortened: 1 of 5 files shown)


spec/flipper/api/v1/actions/features_spec.rb
```diff
@@ -17,22 +17,23 @@
...
-                  "value" => true},
-                  {
-                  "key" =>"groups",
+                  "value" => true
+                },
+                {
+                  "key" => "groups",
                   "name" => "group",
-                  "value" =>[],
+                  "value" => [],
```
jnunemaker: Thanks for tweaking these!

spec/flipper/api/v1/actions/features_spec.rb
```diff
@@ -74,9 +75,44 @@
         post 'api/v1/features', { name: 'my_feature' }
       end
 
-      it 'responds 200 on success' do
+      it 'responds 200 ' do
         expect(last_response.status).to eq(200)
-        expect(json_response).to eq({})
+      end
+
+      it 'returns decorated feature' do
```
jnunemaker: I like returning the decorated feature as you are now. 👍

PR #188, opened 2016-11-04
Make shared_adapter_specs work with disable_monkey_patching setting

ddonahue99:
RSpec ships with disable_monkey_patching! set to true these days, so it's a good practice to prefix top-level blocks with the RSpec namespace. In my case, I have a custom adapter that tries to pull these shared examples in, but because of my RSpec config, it blows up.

See https://www.relishapp.com/rspec/rspec-core/v/3-0/docs/configuration/zero-monkey-patching-mode


AlexWheeler:
@ddonahue99 agreed this is a great idea.  Just to make sure I understand.  You have RSpec monkey patching disabled and you're getting an undefined method "shared_examples_for" since it is no longer defined on Main or Module

PR #189, opened 2016-11-07
disables RSpec monkey patching

AlexWheeler:
As pointed out in #188 RSpec >= 3.0 wants to avoid monkey patching.  This will help guide filpper spec code in the right direction by not allowing monkey patched methods


jnunemaker:
I wanted @ddonahue99 to get credit for the original PR so I merged it and then yours over it. :) Hope that is cool. Thanks to both!

PR #191, opened 2016-11-10
api v1 actors gate docs

AlexWheeler:


docs/api/README.md
```diff
@@ -446,3 +446,113 @@ Successful disabling of the group will return a 200 HTTP status and the feature
...
+**Parameters**
+
+* `feature_name` - The name of the feature
+
+* `flipper_id` - The flipper_id of actor to enable
+
+**Request**
+
+```
+curl -X POST -d "flipper_id=user@user.com" http://example.com/flipper-api/api/v1/features/reports/actors
```
jnunemaker: I'd love to see this flipper_id (and the ones below) be something like class:id (e.g. User:1). Using an email could be problematic as people change email addresses. I recommend using database ids as they don't change and I don't want to potentially mislead people.

PR #192, opened 2016-11-10
Wait for ActiveSupport to tell us when active_record is loaded

johnreilly:
This prevents overwriting the `belongs_to_required_by_default` setting in Rails 5, as described in #129


jnunemaker:
Works for me and it passes the gambit of CI.

PR #197, opened 2016-11-17
Add sequel adapter

lazebny:
Hey, 
First commit allows not install mongo, memcached, redis, gems in local environment. All tests suite is spining up in docker containers and this could be reached with 1 command. Please check in README.md file.
Second commit add sequel adapter.

### DIFF
--- a/flipper-sequel.gemspec
+++ b/flipper-sequel.gemspec
@@ -1,0 +1,25 @@
+# -*- encoding: utf-8 -*-
+require File.expand_path('../lib/flipper/version', __FILE__)
+
+flipper_sequel_files = lambda { |file| file =~ /sequel/ }
+
+Gem::Specification.new do |gem|
+  gem.authors       = ["John Nunemaker"]
+  gem.email         = ["nunemaker@gmail.com"]
+  gem.summary       = "Sequel adapter for Flipper"
+  gem.description   = "Sequel adapter for Flipper"
+  gem.license       = "MIT"
+  gem.homepage      = "https://github.com/jnunemaker/flipper"
+
+  extra_files = [
+    "lib/flipper/version.rb",
+  ]
+  gem.files         = `git ls-files`.split("\n").select(&flipper_sequel_files) + extra_files
+  gem.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n").select(&flipper_sequel_files)
+  gem.name          = "flipper-sequel"
+  gem.require_paths = ["lib"]
+  gem.version       = Flipper::VERSION
+
+  gem.add_dependency 'flipper', "~> #{Flipper::VERSION}"
+  gem.add_dependency 'sequel', ">= 4.0.0", "< 5"
+end
--- a/lib/flipper-sequel.rb
+++ b/lib/flipper-sequel.rb
@@ -1,0 +1,1 @@
+require 'flipper/adapters/sequel'
--- a/lib/flipper/adapters/sequel.rb
+++ b/lib/flipper/adapters/sequel.rb
@@ -1,0 +1,167 @@
+require 'set'
+require 'flipper'
+require 'sequel'
+
+module Flipper
+  module Adapters
+    class Sequel
+      include ::Flipper::Adapter
+
+      # Private: Do not use outside of this adapter.
+      class Feature < ::Sequel::Model(:flipper_features)
+        plugin :timestamps, update_on_create: true
+      end
+
+      # Private: Do not use outside of this adapter.
+      class Gate < ::Sequel::Model(:flipper_gates)
+        plugin :timestamps, update_on_create: true
+      end
+
+      # Public: The name of the adapter.
+      attr_reader :name
+
+      # Public: Initialize a new Sequel adapter instance.
+      #
+      # name - The Symbol name for this adapter. Optional (default :active_record)
+      # feature_class - The AR class responsible for the features table.
+      # gate_class - The AR class responsible for the gates table.
+      #
+      # Allowing the overriding of name is so you can differentiate multiple
+      # instances of this adapter from each other, if, for some reason, that is
+      # a thing you do.
+      #
+      # Allowing the overriding of the default feature/gate classes means you
+      # can roll your own tables and what not, if you so desire.
+      def initialize(options = {})
+        @name = options.fetch(:name, :sequel)
+        @feature_class = options.fetch(:feature_class) { Feature }
+        @gate_class = options.fetch(:gate_class) { Gate }
+      end
+
+      # Public: The set of known features.
+      def features
+        @feature_class.all.map(&:key).to_set
+      end
+
+      # Public: Adds a feature to the set of known features.
+      def add(feature)
+        # race condition, but add is only used by enable/disable which happen
+        # super rarely, so it shouldn't matter in practice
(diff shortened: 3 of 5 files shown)


lib/flipper-sequel.rb
```diff
@@ -0,0 +1,3 @@
+ActiveSupport.on_load(:sequel) do
```
jnunemaker: This could lead to some errors for people who aren't using active support. active support is [not a requirement for flipper](https://github.com/jnunemaker/flipper/blob/19f8c1aba5d56689d21dbe073f76318b81db2884/flipper.gemspec) or for flipper-sequel, so we can't use it here. The only reason the active record adapter does is because activesupport is a dependency of active record, which is listed in flipper active record.

I would just remove everything but the vanilla require here.

spec/flipper/adapters/sequel_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+require 'generators/flipper/templates/sequel_migration'
+require 'flipper/spec/shared_adapter_specs'
+
+RSpec.describe Flipper::Adapters::Sequel do
+  subject do
+    described_class.new(feature_class: feature_class,
+                        gate_class: gate_class)
+  end
+
+	let(:feature_class) { Flipper::Adapters::Sequel::Feature }
```
jnunemaker: Formatting looks a little odd here. Would you mind moving this line and the next one back a few characters to line it up with the subject and before blocks?

spec/flipper/adapters/sequel_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+	let(:gate_class) { Flipper::Adapters::Sequel::Gate }
+
+  before(:each) do
+    CreateFlipperTablesSequel.new(Sequel::Model.db).up
+    feature_class.dataset = feature_class.dataset
```
jnunemaker: I am not that familiar with sequel. What is the purpose of this dataset assignment?

flipper-sequel.gemspec
```diff
@@ -0,0 +1,26 @@
...
+Gem::Specification.new do |gem|
+  gem.authors       = ["John Nunemaker"]
+  gem.email         = ["nunemaker@gmail.com"]
+  gem.summary       = "Sequel adapter for Flipper"
+  gem.description   = "Sequel adapter for Flipper"
+  gem.license       = "MIT"
+  gem.homepage      = "https://github.com/jnunemaker/flipper"
+
+  extra_files = [
+    "lib/generators/flipper/templates/migration.rb",
```
jnunemaker: I believe this should be `lib/generators/flipper/templates/sequel_migration.rb`.

jnunemaker: Actually, I don't think this is necessary, as the migration file has sequel in the name, which means the flipper_sequel_files block should pick it up fine. I only had to do this for active record because the migration file did not have active_record in the name. I think you can just remove this line.

README.md
```diff
@@ -89,6 +89,10 @@ Of course there are more [examples for you to peruse](examples/). You could also
 2. `script/release`
 3. Profit.
 
+### Development in docker
```
jnunemaker: I know basically nothing about docker. 😬  Would you be willing to be a collaborator on the repo, answer any docker questions and review any docker PRs that come up?

lib/flipper/adapters/sequel.rb
```diff
@@ -0,0 +1,167 @@
...
+            args = {
+              feature_key: feature.key,
+              key: gate.key.to_s
+            }
+            @gate_class.where(args).delete
+
+            @gate_class.create(gate_attrs(feature, gate, thing))
+          end
+        when :set
+          @gate_class.create(gate_attrs(feature, gate, thing))
```
jnunemaker: What is the behavior for sql if this fails? Does it raise an exception or just return false?

lib/generators/flipper/templates/sequel_migration.rb
```diff
@@ -0,0 +1,24 @@
+class CreateFlipperTablesSequel < Sequel::Migration
```
jnunemaker: How do people know to use this file? Should we hook it up as a rails generator? Or is that bad form since perhaps sequel people may also lean toward sinatra? Does sequel provide any means of using this? Perhaps we just document that people need to require the file? I'd love your thoughts on this.

spec/flipper/adapters/sequel_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'helper'
+require 'sequel'
+
+Sequel::Model.db =  Sequel.sqlite
```
jnunemaker: Does this create a file or is it in memory like I use with the active record adapter?

lib/flipper-sequel.rb
lazebny: Ok, will do. Glad to hear positive vibration from your side.

spec/flipper/adapters/sequel_spec.rb
lazebny: Yeah it does. I'll change to memory.

README.md
lazebny: Sure, I am not e big expert in docker just use it in our testing/development suits. But I think it will be fun to learn new resoanable things in future PRs.

lib/flipper/adapters/sequel.rb
lazebny: I'll test it tomorrow. We don't use a validations in sequel models, any other stuff raise an exception.

lazebny:
I'll split up these features. 
What do you think about rubocop integration with Travis Ci?

spec/flipper/adapters/sequel_spec.rb
lazebny: Sure, sorry for this style. Rubocop made me lazy, because we use autocorrect before commit.

lib/flipper/adapters/sequel.rb
jnunemaker: Good point. There aren't any validations so it probably doesn't matter.

lib/generators/flipper/templates/sequel_migration.rb
jnunemaker: We can do that in subsequent PRs if you would prefer. I don't see it as a blocker to merging, just curious in general.

jnunemaker:
> I'll split up these features. 

:cool: 

> What do you think about rubocop integration with Travis Ci?

Sounds like a good idea but not sure what all it would mean. Perhaps a separate PR that we can look at on its own?

lib/flipper/adapters/sequel.rb
lazebny: > Regardless of whether you are using validations for data integrity or just for error messages, the usage is the same. Whenever you attempt to save a model instance, before sending the INSERT or UPDATE query to the database, Sequel::Model will attempt to validate the instance by calling validate. If validate does not add any errors to the object, the object is considered valid, and valid? will return true. If validate adds any errors to the object, valid? will return false, and the save will either raise a Sequel::ValidationFailed exception (the default), or return nil (if raise_on_save_failure is false). 
> 
source: https://github.com/jeremyevans/sequel/blob/master/doc/validations.rdoc

lazebny:
@jnunemaker PR is updated please check.

Issue #195, opened 2016-11-16
API V1 errors documentation / add to response


AlexWheeler:
1. Add documentation for API errors
2. add links to error docs in json error responses ( `Flipper::Api::ErrorResponse::Error#more_info` )

PR #200, opened 2016-11-18
Api v1 error docs and response messaging

AlexWheeler:
Adds error documentation to the api v1 README completing #195 

For now the more_info link will link to https://github.com/jnunemaker/flipper/tree/master/docs/api#error-code-reference.  The anchor tags don't work in preview because user-content is prepended to the ids, but will work when merged.

I updated the ErrorResponse initialize function to just take `(code, message, http_status)` and no link to documentation since for now we only have a few error messages and they all live in one place so that instance variable is just set in the initializer.  Maybe in the future we'll change this if we add more info about errors and want to link to more specific sections of the docs

updates specs to account for adding the more_info in the response

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -775,17 +775,16 @@ Successful disabling of a percentage of time will set the percentage to 0 and re
 ```
 
 ## Errors
-In the event of an error the Flipper API will return an error object.  The error object will contain a Flipper-specific error code, an error message, and a link to the documentation for the given error code.
+In the event of an error the Flipper API will return an error object.  The error object will contain a Flipper-specific error code, an error message, and a link to documentation providing more information about the error.
 
 *example error object*
 ```json
 {
     "code": 1,
     "message": "Feature not found",
-    "more_info": "https://github.com/jnunemaker/flipper/...",
+    "more_info": "https://github.com/jnunemaker/flipper/tree/master/docs/api#error-code-reference",
 }
 ```
-
 ### Error Code Reference
 
 #### 1: Feature Not Found
@@ -805,4 +804,5 @@ The `percentage` parameter is invalid or missing.  `percentage` must be an integ
 The `flipper_id` parameter is invalid or missing.  `flipper_id` cannot be empty.
 
 ####  5: Name Invalid
+
 The `name` parameter is missing.  Make sure your request's body contains a `name` parameter.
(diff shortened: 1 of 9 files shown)


docs/api/README.md
```diff
@@ -773,3 +773,36 @@ Successful disabling of a percentage of time will set the percentage to 0 and re
...
+    "message": "Feature not found",
+    "more_info": "https://github.com/jnunemaker/flipper/tree/master/docs/api#error-code-reference",
+}
+```
+
+### Error Code Reference
+
+#### 1: Feature Not Found
+
+The requested feature does not exist.  Make sure the feature name is spelled correctly and exists in your application's database.
```
jnunemaker: Love it.

docs/api/README.md
```diff
@@ -773,3 +773,36 @@ Successful disabling of a percentage of time will set the percentage to 0 and re
...
+#### 3: Percentage Invalid
+
+The `percentage` parameter is invalid or missing.  `percentage` must be an integer between 0-100 inclusive and cannot be blank.
+
+#### 4: Flipper ID Invalid
+
+The `flipper_id` parameter is invalid or missing.  `flipper_id` cannot be empty.
+
+####  5: Name Invalid
+The `name` parameter is missing.  Make sure your request's body contains a `name` parameter.
```
jnunemaker: I'd maybe drop a newline between the heading and paragraph to keep them consistent with the points above, but feel free to merge without that. No biggie.

lib/flipper/api/error_response.rb
```diff
@@ -20,13 +20,18 @@ def as_json
         end
       end
 
+      def self.documentation_path
```
jnunemaker: Can you remove this? Seems unused.

lib/flipper/api/error_response.rb
```diff
@@ -4,10 +4,10 @@ module ErrorResponse
       class Error
         attr_reader :http_status
 
-        def initialize(code, message, info, http_status)
+        def initialize(code, message, http_status)
```
jnunemaker: Works for me.

lib/flipper/api/error_response.rb
AlexWheeler: totally right! was playing with some other implementations thanks

PR #201, opened 2016-11-18
Show confirmation dialog when deleting a feature

mikesea:
Hi! Thanks so much for working on this. We just set it up internally and are loving it.

Although the Delete button is red and scary, I still found myself a bit surprised when clicking "Delete" didn't require an extra confirmation step. This adds a simple JavaScript dialog to be confirmed before the delete form is submitted.


jnunemaker:
Seems fair. I am not a huge fan of dialogs, but safety first! 😀

Issue #204, opened 2016-11-23
Whitespace in actors_value but not in actor's flipper_id


KevinBongart:
Hi @jnunemaker, thanks for this really useful gem (also hi @AlexWheeler! 👋)

I noticed that by copy-pasting emails as actors_value in the Flipper UI like `test@test.com \n`, then it won't let a User with `test@test.com` as `flipper_id` in.

The UI also makes it difficult to notice that kind of extra whitespace, so it took a little effort debugging on my side to find the root cause.

Should values stored in Redis always be stripped from leading-trailing whitespace? Should the user's custom `Thing#flipper_id` be in charge of this logic? Should the UI run `String#strip` before giving the value to Flipper?

I'd love to help out with this problem but I'm not sure where the responsibility lies between the user, the gem and its various components.

Let me know how I can help!

jnunemaker:
Great question! I would have the UI/API do the strip. Putting it in flipper
could cause unexpected things for users of flipper in the future.

KevinBongart:
Thanks, I also agree it shouldn't be in Flipper. The UI seems the right place, as the API should just save what it's given. I'll take a look today

PR #205, opened 2016-11-23
Strip whitespace out of feature/actor/group values posted by UI

KevinBongart:
Fix https://github.com/jnunemaker/flipper/issues/204

By investigating the initial issue, I was able to identify several areas impacted by it.

1. Using the UI to create two features with similar names but different whitespace:

```ruby
> $flipper.features
=> #<Set: {#<Flipper::Feature:70136261075520 name="   whitespace", state=:conditional, enabled_gate_names=[:actor], adapter=:memoizable>,
 #<Flipper::Feature:70136261217520 name="   whitespace   ", state=:conditional, enabled_gate_names=[:actor], adapter=:memoizable>}>
```

2. Still via the UI, adding an actor to the second one seemed to have actually added it to the first one.

3. Both links in the UI leads to the same feature:

screen shot 2016-11-23 at 11 01 10 am

4. Copy-pasting actor values from one Flipper UI to another adds some whitespace which ends up in Flipper:

```ruby
> $flipper["   whitespace"].actors_value
=> #<Set: {"    test@blueapron.com\t"}>
```

This PR fixes all of these issues above by applying [`String#strip`](http://ruby-doc.org/core-2.3.3/String.html#method-i-strip) to the params send from the front-end to the back-end of Flipper UI.

cc @jnunemaker & @AlexWheeler for review/feedback, and @susanblueapron for finding the bug in the first place

### DIFF
--- a/lib/flipper/ui/actions/actors_gate.rb
+++ b/lib/flipper/ui/actions/actors_gate.rb
@@ -24,7 +24,7 @@ module Flipper
         def post
           feature_name = Rack::Utils.unescape(request.path.split('/')[-2])
           feature = flipper[feature_name.to_sym]
-          value = params["value"].strip
+          value = params["value"].to_s.strip
 
           if Util.blank?(value)
             error = Rack::Utils.escape("#{value.inspect} is not a valid actor value.")
--- a/lib/flipper/ui/actions/actors_gate.rb
+++ b/lib/flipper/ui/actions/actors_gate.rb
@@ -24,7 +24,7 @@ module Flipper
         def post
           feature_name = Rack::Utils.unescape(request.path.split('/')[-2])
           feature = flipper[feature_name.to_sym]
-          value = params["value"]
+          value = params["value"].strip
 
           if Util.blank?(value)
             error = Rack::Utils.escape("#{value.inspect} is not a valid actor value.")
--- a/lib/flipper/ui/actions/features.rb
+++ b/lib/flipper/ui/actions/features.rb
@@ -34,7 +34,7 @@ module Flipper
             halt view_response(:feature_creation_disabled)
           end
 
-          value = params["value"].strip
+          value = params["value"].to_s.strip
 
           if Util.blank?(value)
             error = Rack::Utils.escape("#{value.inspect} is not a valid feature name.")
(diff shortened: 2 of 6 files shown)


lib/flipper/ui/actions/actors_gate.rb
```diff
@@ -24,7 +24,7 @@ def get
         def post
           feature_name = Rack::Utils.unescape(request.path.split('/')[-2])
           feature = flipper[feature_name.to_sym]
-          value = params["value"]
+          value = params["value"].strip
 
           if Util.blank?(value)
```
midu: looks like the code expects a possible `nil` value here, which means that your change on line `27` will raise an `undefined method for nil nilclass`

maybe `strip` inside the if block?

lib/flipper/ui/actions/features.rb
```diff
@@ -34,7 +34,7 @@ def post
             halt view_response(:feature_creation_disabled)
           end
 
-          value = params["value"]
+          value = params["value"].strip
```
midu: same as above

spec/flipper/ui/actions/actors_gate_spec.rb
```diff
@@ -46,13 +48,23 @@
...
     end
 
     context "disabling an actor" do
+      let(:value) { "User:6" }
+
       before do
         flipper[:search].enable_actor Flipper::UI::Actor.new("User:6")
         post "features/search/actors",
-          {"value" => "User:6", "operation" => "disable", "authenticity_token" => token},
+          {"value" => value, "operation" => "disable", "authenticity_token" => token},
```
midu: maybe add a test where you don't pass a `value`? seems like that's missing here

KevinBongart:
@midu Added more tests related to empty values (not exactly the same as a param hash missing the `"value"` key but close enough). Not sure what's best between:

1. `value = params["value"].to_s.strip`
2. `value = params["value"]&.strip` (only compatible with Ruby 2.3+ I believe?
3. `value = params["value"] && params["value"].strip

jnunemaker:
My preference is to_s.strip. It works with older versions of ruby and flipper doesn't care about nil uniquely since nil/blank are treated the same.

PR #210, opened 2016-12-26
Use primary keys with Sequel

cyberdelia:
Use primary keys with Sequel, which are usually more practically that simple indexes.

It appears the current implementation does not require a feature to exists when creating a gate, so a foreign key on ``flipper_gates.feature_key`` cannot be used.

### DIFF
--- a/lib/flipper/adapters/sequel.rb
+++ b/lib/flipper/adapters/sequel.rb
@@ -9,11 +9,15 @@ module Flipper
 
       # Private: Do not use outside of this adapter.
       class Feature < ::Sequel::Model(:flipper_features)
+        unrestrict_primary_key
+
         plugin :timestamps, update_on_create: true
       end
 
       # Private: Do not use outside of this adapter.
       class Gate < ::Sequel::Model(:flipper_gates)
+        unrestrict_primary_key
+
         plugin :timestamps, update_on_create: true
       end
 
--- a/lib/generators/flipper/templates/sequel_migration.rb
+++ b/lib/generators/flipper/templates/sequel_migration.rb
@@ -1,11 +1,10 @@
 class CreateFlipperTablesSequel < Sequel::Migration
   def up
     create_table :flipper_features do |t|
-      String :key, null: false
+      String :key, primary_key: true, null: false
       DateTime :created_at, null: false
       DateTime :updated_at, null: false
     end
-    add_index :flipper_features, :key, unique: true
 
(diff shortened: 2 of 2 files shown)


AlexWheeler:
@cyberdelia thanks for the PR.  I agree that in most cases it makes sense to give relations a primary key.  curious are you running into a problem or just think using primary keys is a better practice?

cyberdelia:
Using primary keys is usually a better practice. I'm looking into using flipper sequel adapter, and making sure that proper indexes and constraints are presents.

AlexWheeler:
cool would like to get @lazebny take on this as I'm not as familiar with the Sequel gem.

some thoughts from a db perspective:

currently `flipper_features` has a unique index + not null constraint on `key` so in practice adding the primary key would have the same effect as the current implementation (except records getting reordered on disk for databases like sql server that support clustered indexes)

flipper_gates has a unique index on `feature_key, key, value` so for most databases the primary key would have the same effect in regards to the unique index.

flipper_gates has a not null constraint on `feature_key, key`, so adding the composite primary key will add a not null constraint on `value` too which doesn't currently exist, requiring a data migration for current users wanting to have this schema in the case they have a null value (which they hopefully wouldn't)

other note the active_record adapter uses a surrogate primary key (integer) for what its worth

jnunemaker:
> other note the active_record adapter uses a surrogate primary key (integer) for what its worth

I think I would lean toward this. These tables shouldn't get very large, so having primary keys that aren't used and secondary indexes that handle most of the use cases seems fine to me. For some reason I am thinking that having an integer primary key is a bit easier for replication in certain db systems (maybe mysql? maybe just making that up?). 

I'm really not opinionated on this, but happy to switch if there is a strong reason. I think the more interesting thing this brings up is how do people on the older migration get in sync with the latest. I thought about that when I first started, but mostly hoped since these tables were simple it wouldn't be an issue. I guess people can just migrate manually by running the alters they need or creating new tables, backfilling the data and swapping them.

lazebny:
As for me this is a good idea, cant see any issues now. Only need create a migration notice for people which use this adapter.

cyberdelia:
Note that this change is backward compatible, it's essentially creating the same index as before but in a "stricter" way.

jnunemaker:
Ok, merging this. I don't see any problems. Only thing would be people who have already ran migration but they are fine as is. They can also do the alters on their own if they want.

PR #211, opened 2017-01-06
Add redis cache adapter

### LINKED ISSUE #142: Redis cache adapter
Redis adapter to cache adapter calls just like the Dalli Memcached adapter

camallen:
closes #142 - first attempt at adding a redis cache adapter. I've added some testing setup that was broken and documented running them via docker-compose.

I mimicked the dalli cache adapter almost entirely, i think the only choices i had to make were is this included in the flipper-redis gem and how to use the setex cmd (ttl) with redis. I chose yes and to marshal objects when storing / retrieving, limited testing on this show the features serializing in and out of redis ok.  Happy to change the implementation if you want.

Side note: when testing in the pry debugger, the inspect method on a feature would call the `adapter.get`[here](https://github.com/jnunemaker/flipper/blob/cb7ba8f9bc0c69671978d3f28d212b893b01323b/lib/flipper/feature.rb#L219) and warm the cache which took me a while to figure out. Maybe i should just make a note of this in the readme / wiki...?

### DIFF
--- a/spec/helper.rb
+++ b/spec/helper.rb
@@ -6,7 +6,6 @@ FlipperRoot = Pathname(__FILE__).dirname.join('..').expand_path
 
 require 'rubygems'
 require 'bundler'
-require 'pry'
 
 Bundler.setup(:default)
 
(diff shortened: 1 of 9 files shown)


spec/helper.rb
```diff
@@ -6,6 +6,7 @@
 
 require 'rubygems'
 require 'bundler'
+require 'pry'
```
camallen: happy to remove this if it's not used widely

PR #219, opened 2017-01-25
Add JsonParams to Api to add JSON request body to rack params

AlexWheeler:
This is an interesting issue uncovered from some work I've been doing on the api client.  RSpec sends post requests as x-www-form-urlencoded and rack adds any parameters to the Rack::Request#params hash.  The api actions look at params for the request data.  The problem is that for application/json requests rack does not add the parsed body to the params hash.  This PR adds simple custom middleware that for request with content-type application/json will parse the body and add the key/values to the params hash so that the api can handle both x-www-form-urlencoded and application/json requests.

I added a spec for the middleware and also to one of the api action specs to ensure the api middleware stack is working as expected.  The middleware spec keeps me confident the middleware works as expected, but wondering if there's a better way to test that the Flipper Api rack app uses the middleware other than testing every action that looks at the request body?  Maybe just testing JsonParams is included in the stack + the middleware test would be solid?

### DIFF
--- a/lib/flipper/api/json_params.rb
+++ b/lib/flipper/api/json_params.rb
@@ -1,0 +1,45 @@
+require 'rack/utils'
+
+module Flipper
+  module Api
+    class JsonParams
+      include Rack::Utils
+
+      def initialize(app)
+        @app = app
+      end
+
+      CONTENT_TYPE = 'CONTENT_TYPE'.freeze
+      QUERY_STRING = 'QUERY_STRING'.freeze
+      REQUEST_BODY = 'rack.input'.freeze
+
+      # Public: Merge request body params with query string params
+      # This way can access all params with Rack::Request#params
+      # Rack does not add application/json params to Rack::Request#params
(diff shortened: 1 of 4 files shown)


AlexWheeler:
looks like Rails 3 rack version doesn't have Request#update_param yet.  Will look into this after work today.  Might just need to set it on the env itself without the request helpers

lib/flipper/api/json_params.rb
```diff
@@ -0,0 +1,28 @@
...
+      def initialize(app)
+        @app = app
+      end
+
+      def call(env)
+        request = Rack::Request.new(env)
+        if request.content_type == 'application/json'
+          body = request.body.read
+          request.body.rewind
+          update_params!(request, body)
```
jnunemaker: I tend not to use bang methods unless there is a non bang version.

AlexWheeler: That is a great point. Will change this and keep that in mind for the future thanks!

PR #220, opened 2017-01-25
Add mount priority section to Flipper Api docs

AlexWheeler:
* mount priority matters.  mount /flipper-api before /flipper

* since mount uses match with anchor: false any route after flipper*
will match that rack application and so flipper-api would never get called


jnunemaker:
Good stuff!

PR #221, opened 2017-01-27
Http Adapter (WIP)

AlexWheeler:
Flipper Api Adapter.  Have been testing it out against a Rails app and working great - can do everything a user could do through the UI.  Still need to add some tests for error responses.  Will squash commits when ready to merge

*thoughts*

It might be best to namespace this under Http (although V1.new looks a little strange) to avoid conflicts in the adapter namespace.
`Flipper::Adapters::Http::V1`?

some adapter methods return true/false:
add
remove
enable
disable
clear

Right now I'm just checking if the request was successful.  These methods need to preserve the interface, but I'm thinking it would be nice to pass some sort of logger perhaps in the configuration or when initialize so that when configured a failed request could log the response somewhere (and default to stdout?)

```
# optional configuration if you want to send custom headers or basic auth on every request
Flipper::Adapters::Http.configure do |c|
  c.headers = { "X-Custom-Header" => "value" }
  c.basic_auth = { "my_username" => "my_password"
end

adapter = Flipper::Adapters::Http.new("http://rubyapp/flipper-api)
flipper = Flipper.new(adapter)
```


lib/flipper/adapters/http.rb
```diff
@@ -0,0 +1,197 @@
...
+
+      # Public: DELETE http request
+      def delete(path, data = {})
+        uri = URI.parse(path)
+        http = Net::HTTP.new(uri.host, uri.port)
+        request = Net::HTTP::Delete.new(uri.request_uri, @headers)
+        request.body = data.to_h.to_json
+        request.basic_auth(@basic_auth_username, @basic_auth_password) if basic_auth?
+        http.request(request)
+      end
```
AlexWheeler: These three methods have very similar structure, but figured it better to go the boring easy to understand path over abstracting this into something like a request method that takes the http verb, etc.

lib/flipper/adapters/http.rb
```diff
@@ -0,0 +1,197 @@
...
+      # feature - Feature instance
+      def add(feature)
+        response = @request.post(@path + '/api/v1/features', name: feature.key)
+        response.is_a?(Net::HTTPOK)
+      end
+
+      def get_multi(features)
+        # could be cool to add this feature as an api endpoint requesting multiple features
+        # or alternatively use a persistent connection and send multiple requests
+      end
```
AlexWheeler: Still need to implement this

lib/flipper/adapters/http.rb
jnunemaker: I love boring. :)

spec/flipper/adapters/http_spec.rb
```diff
@@ -0,0 +1,399 @@
+require 'helper'
+require 'flipper/adapters/http'
+require 'webmock/rspec'
+require 'flipper/spec/shared_adapter_specs'
+
+RSpec.describe Flipper::Adapters::Http do
```
jnunemaker: My only concern with this class is that it could easily get out of date with the shared adapter specs. I wonder how hard it would be to mount the API in a rack app, point this test adapter at that and then just use the shared adapter specs with additional specs for the http specific stuff like configuration. 

A quick google brought up [this](http://stackoverflow.com/questions/10166611/launching-a-web-server-inside-ruby-tests) which mentions/links to a few ideas. Something [like this](https://github.com/teamcapybara/capybara/blob/dbeec59726a983b801e56838670d8dd86b6e4f8f/spec/server_spec.rb#L7-L12) would be super rad! What do you think?

jnunemaker:
@AlexWheeler I just want to make it really clear, since I don't know when I last said it, it is **SO AWESOME** all the work you have done on flipper and I really appreciate it!

spec/flipper/adapters/http_spec.rb
AlexWheeler: yes that is for sure a great idea and have been thinking of some ways to do it.  I'll check out those links and find a way to make this happen because I think it is important to make sure its tested against the shared specs.

lib/flipper/adapters/http.rb
```diff
@@ -0,0 +1,197 @@
...
+
+      attr_reader :name
+
+      class << self
+        attr_accessor :configuration
+      end
+
+      # Public: initialize with api url
+      # http://www.myapp.com/api-mount-point
+      def initialize(path_to_mount)
```
AlexWheeler: yes you're right that's better 👍

AlexWheeler:
Thanks for the epic feedback as always - I'll work to get some changes up this weekend / next week. 

It's easy to find time to work on Flipper knowing there's always new code / design decisions I come across that blow my mind.  Have learned a lot from your work so thank you!

jnunemaker:
@AlexWheeler I took a look at this last night. Looks like it was really close. I'm realizing that the key difference is that flipper adapters expect the adapted thing to work regardless of if a feature exists or not. I picture it like strict mode vs not-strict mode. Adapters expect not-strict, but the API is currently strict and returns 404 for features that do not exist, rather than creating them and doing whatever the API request wants to do. I hacked on it for a bit and was able to get things passing in https://github.com/jnunemaker/flipper/pull/226. Check it out and let me know what you think.

AlexWheeler:
@jnunemaker sweet, have been offline for past week and so epic to see all this progress.  The only ideas I wanted to add about the http adapter was:

1. Have the adapter handle the disabling of a gate value.  I like the idea of only having one way to set a gate value through the enable endpoint, and then the disable endpoints are just for completely disabling a feature. 

So instead of the api handling for example: 
```
if percentage >= 0
  feature.enable_percentage_of_actors(percentage)
else
  feature.disable_percentage_of_actors
end
```
The adapter would just request the correct route (which would be enable when trying to disable a value > 0)

2.  `GET /features/:feature_that_doesnt_exist` should still 404 and have the adapter handle the returning the empty feature

This way we don't couple the api with the implementation of the adapter.  A rest client might want to do something different when it finds out a feature does not exist yet.  This logic would be the same for get_multi, where the adapter would just figure out which features didn't exist and output the right hash so they could be initialized by flipper.

Would love to hear your thoughts

jnunemaker:
Definitely open to a PRs for both of those. I was balancing a few different items and can't remember why I didn't go either of those routes. Look forward to seeing what you come up with when you have time.

PR #223, opened 2017-02-08
Api Flipper from Env

jnunemaker:
This makes it so that the API can use flipper from the request env instead of having to pass an instance in. One benefit is that the middleware is more simple. It now just reads `env["flipper"]`. Another benefit is that the flipper instance can be dynamically determined in upstream middleware. This allows the API to be mounted with a prefix or something and a little bit of middleware that sets the flipper instance based on that prefix.

This is all backwards compatible, it just adds a bit more flexiblity for something I have up my sleeve unrelated to this. 😁 @AlexWheeler any thoughts on this either way?

### DIFF
--- a/lib/flipper/api.rb
+++ b/lib/flipper/api.rb
@@ -2,18 +2,20 @@ require 'rack'
 require 'flipper'
 require 'flipper/api/middleware'
 require 'flipper/api/json_params'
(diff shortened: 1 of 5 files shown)


AlexWheeler:
ah sweet! I definitely think this makes sense and stoked to see what you have coming

PR #226, opened 2017-02-10
Get HTTP Adapter Passing

jnunemaker:
@AlexWheeler I got interested in the http adapter and went to town last night. This gets all the specs passing and shuffles some stuff around. I branched from your branch so if you are cool with this, we can just merge it into yours and go from there. Any thoughts on this?


lib/flipper/adapters/http.rb
```diff
@@ -115,8 +112,20 @@ def add(feature)
       end
 
       def get_multi(features)
```
AlexWheeler: awesome

jnunemaker: This should Just Work™ when https://github.com/jnunemaker/flipper/pull/224 is merged.

PR #229, opened 2017-02-11
UI Flipper from Env

jnunemaker:
This makes the UI load from the env. It also makes the memoizer middleware use the flipper from env for consistency sake. The downside is that it requires a parameter change to the memoizer middleware, but it is a minor change we can note it in the changelog for those upgrading.


jnunemaker:
Oh and lastly it adds the memoizer to the api.

Issue #231, opened 2017-02-12
Add redis cache adapter to Optimization.md


jnunemaker:
Dalli is [listed](https://github.com/jnunemaker/flipper/blob/e4fb002ad7c246bbe0116035c7247170c4ab4e33/docs/Optimization.md#cache-adapters). Just need to add the redis cache adapter with an example.

jnunemaker:
Closed by #234

PR #234, opened 2017-02-13
Add RedisCache to Optimization.md

AlexWheeler:
completes #231 

*Applications using [Redis](https://redis.io/) via the [redis-rb]* - I think its a good idea to be explicit here in case a project is using a redis client with a different public API, even though clients with the same public api will work this is the one we've tested it against and the official ruby client.

* removes trailing whitespace


jnunemaker:
Thanks! Looks great.

PR #235, opened 2017-02-13
Http tweaks

jnunemaker:
* Drops config. Felt cumbersome for just a few http options. We may need it back, but I'd rather go slim for now until we definitely do.
* Renames Request => Client. Felt more accurate since it deals with many requests.
* `to_json` to `JSON.generate`. I didn't realize that `to_json` exists in ruby and thought that was only coming from rails which might not always be there so I switched. I later realized that `to_json` does seem to be there in ruby for some things, though I'm not sure it is for hash (that maybe comes from rails?). Just seemed best to be explicit and use `JSON.generate`.
* made merging mount point with flipper api paths a bit more robust (uses URI for mount point and path and tries to merge them.


jnunemaker:
I'd still like to go farther on this class and make it a full wrapper around the api (e.g.: `client.features`, `client.feature("search")`, etc.), but calling it quits for the weekend.

PR #236, opened 2017-02-14
Api no longer returns features that do not exist yet - let HTTP adapter handle this

AlexWheeler:
* The HTTP adapter, as opposed to the api will return any features requested that do not exist.  This will not effect how flipper currently works with the http adapter.  It will only change how the api responds.

Right now for `http_adapter.get(flipper[:feature_that_is_not_in_database])` the json response will return json that makes it seem like the resource exists when it really doesn't.  The same applies for get_multi.

This PR changes this so that `/features/:feature_that_does_not_exist` will 404, and the http adapter will handle the returning of the feature so that it still works the same way from the perspective of a flipper.  It will just initialize a new flipper feature.  In other words `flipper[:some_non_existant_feature]` will still return a feature instance.

It also filters the api response for get_multi to only return features that exist.  Similarly to the get feature endpoint, the adapter handles returning the default_config for the features that don't exist.

* This will nicely decouple the REST api from the http adapter
* A REST client might want to know if a feature does not exist

### DIFF
--- a/lib/flipper/adapters/http.rb
+++ b/lib/flipper/adapters/http.rb
@@ -24,10 +24,14 @@ module Flipper
 
       def get(feature)
         response = @client.get("/features/#{feature.key}")
-        raise Error, response unless response.is_a?(Net::HTTPOK)
-
-        parsed_response = JSON.parse(response.body)
-        result_for_feature(feature, parsed_response.fetch('gates'))
+        if response.is_a?(Net::HTTPOK)
+          parsed_response = JSON.parse(response.body)
+          result_for_feature(feature, parsed_response.fetch('gates'))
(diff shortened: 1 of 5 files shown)


AlexWheeler:
@jnunemaker that's a really great point, the strict parameter option sounds possibly just as a query string parameter

jnunemaker:
@AlexWheeler I'd be even up for defaulting it to true if it doesn't exist so strict mode is the default. Then we could just allow the adapter to do strict=false as query param. If you are down with that I can try a PR and see how it feels.

AlexWheeler:
@jnunemaker yes that sounds like a great idea to me

PR #239, opened 2017-03-02
Update Gates.md

Jwan622:
I was working on a project that requires feature release in small batches at Blue Apron. I feel like this would greatly clarify the usage of groups. I fee like explanations about what the various api methods are doing on a Ruby-level clarifies this greatly.


jnunemaker:
@Jwan622 you rock! Really appreciate that you took the time to distill what you learned into the docs for others.

PR #244, opened 2017-03-12
Allow Unregistered Groups

jnunemaker:
This makes it possible to enable/disable groups that are not registered. The use case is API's that are mounted on their own apart from where the groups are registered in the code.

### DIFF
--- a/lib/flipper/api/v1/actions/groups_gate.rb
+++ b/lib/flipper/api/v1/actions/groups_gate.rb
@@ -28,7 +28,7 @@ module Flipper
 
           def ensure_valid_params
             return if allow_unregistered_groups?
-            return unless Flipper.group_exists?(group_name)
+            return if Flipper.group_exists?(group_name)
 
             json_error_response(:group_not_registered)
           end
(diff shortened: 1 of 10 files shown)


lib/flipper/api/v1/actions/groups_gate.rb
```diff
@@ -27,7 +27,23 @@ def delete
...
+            end
+
+            return if allow_unregistered_groups?
+            return if Flipper.group_exists?(group_name)
+
+            json_error_response(:group_not_registered)
+          end
+
+          def allow_unregistered_groups?
+            allow_unregistered_groups = params['allow_unregistered_groups']
```
jnunemaker: @AlexWheeler if you have any time to look this over, I'd be curious what your thoughts are. It will be a few weeks at least before I do another release, so no rush. I went with a specific param instead of `strict`. `strict` would affect everything whereas specific params are more isolated. Specific felt better as a way to make it possible to opt in to specific functionality instead of everything (if there are other things down the road).

AlexWheeler: cool, I really like this approach and saw that the adapters implements this now too from #245 ! I agree that specific parameters is a much better idea.  I think we still  need to do something similar for the preloading in `features#get` right (so that the adapter can pass a parameter to get all requested features even if they don't exist (avoiding unnecessary queries)?

PR #248, opened 2017-03-25
Rename shadowing variable

nilbus:
`Percentage#open?` has 2 things named key, a method and a local variable that shadows it. This works fine but is confusing to read.

This patch renames the local variable, since `#key` is part of the `Gate` interface.


jnunemaker:
Ah nice. Thanks!

PR #260, opened 2017-05-24
Allow custom layouts

cachar:
Allows the user to use a custom layout- indirectly addresses https://github.com/jnunemaker/flipper/issues/101, in that a custom layout could include a banner with information about the environment.

I'm planning on adding some tests, but here's a PR to activate your spidey sense. :)


jnunemaker:
Interesting idea. Is the only use case for you to add the banner or do you have other things in mind? I ask because if you only want the banner, I think I'd prefer only exposing the ability to add a banner, but if you have other needs, this might be a more flexible route.

cachar:
I could see wanting to make the flipper page look more like the other pages in an app so it looks more integrated- might involve adding a footer or something other than a banner.

PR #264, opened 2017-06-07
Alphabetize adapters

AlexWheeler:
super minor but thought it would be nice to have these alphabetized (case insensitive because it makes it easier to spot named adapters i.e. Redis v. memory) to make it easier to spot supported adapters


jnunemaker:
I'm all for minor improvements. Thanks for caring!

Issue #263, opened 2017-06-05
ActiveSupport::Cache::Store support


rzane:
ActiveSupport::Cache::Store is an abstraction on top of a bunch of different caching backends: file, memory, null, redis, dalli, etc.

This would be nice for rails users:

```ruby
Flipper::Adapters::Cache.new(some_other_adapter, Rails.cache, 4800)
```

Another similar project is [monetta](https://github.com/minad/moneta). By leveraging one of these existing cache abstractions, this project could support more backends with (I assume) less maintenance effort.

rzane:
Looks like there's a gem for that! https://github.com/hundredwatt/flipper-cache_store.

jnunemaker:
@rzane looks like that gem is pretty out of date. The adapter interface has changed and it will need an update.

Overall, I like the idea of an ActiveSupport cache adapter and a monetta adapter. I'm very familiar with monetta as it inspired https://github.com/jnunemaker/adapter a long time ago. 😄

jnunemaker:
https://github.com/jnunemaker/flipper/pull/265 fixed this.

PR #265, opened 2017-06-08
Add cache_store adapter for caching via ActiveSupport::CacheStore

AlexWheeler:
#263 

* Helpful adapter for Rails users that want to use their configured cache store cache.

http://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html

*  can pass options as described in docs above e.g. :expires_in to Adapters::CacheStore constructor to use this for all flipper keys.  Otherwise it will default to whatever the user has set on the cache via its constructor.

From docs ^ *Some implementations may not support all methods beyond the basic cache methods of fetch, write, read, exist?, and delete.*  Sinc we're using fetch_all we might want to check if the `@cache` responds to fetch_all and if not then call our own fetch_all that loops with fetch.  Thoughts?

### DIFF
--- a/lib/flipper/adapters/cache_store.rb
+++ b/lib/flipper/adapters/cache_store.rb
@@ -1,0 +1,103 @@
+module Flipper
+  module Adapters
+    # Public: Adapter that wraps another adapter with the ability to cache
+    # adapter calls in ActiveSupport::CacheStore caches.
+    #
+    class CacheStore
+      include ::Flipper::Adapter
+
+      Version = 'v1'.freeze
+      Namespace = "flipper/#{Version}".freeze
+      FeaturesKey = "#{Namespace}/features".freeze
+
+      # Private
+      def self.key_for(key)
+        "#{Namespace}/feature/#{key}"
+      end
+
+      # Internal
+      attr_reader :cache
+
+      # Public: The name of the adapter.
+      attr_reader :name
+
+      # Public
+      def initialize(adapter, cache, options = nil)
+        @adapter = adapter
+        @name = :cache_store
+        @cache = cache
+        @options = options
+      end
+
+      # Public
+      def features
+        @cache.fetch(FeaturesKey, @options) do
+          @adapter.features
+        end
+      end
+
+      # Public
+      def add(feature)
+        result = @adapter.add(feature)
+        @cache.delete(FeaturesKey, @options)
+        result
+      end
+
+      ## Public
+      def remove(feature)
+        result = @adapter.remove(feature)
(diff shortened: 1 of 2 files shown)


jnunemaker:
I think because not all methods support the same options, I would be inclined to be explicit about the options we support and where we support them. TTL/expires seems like the most sensible one to support out of the gate. Any other obvious ones to support?

> From docs ^ Some implementations may not support all methods beyond the basic cache methods of fetch, write, read, exist?, and delete. Sinc we're using fetch_all we might want to check if the @cache responds to fetch_all and if not then call our own fetch_all that loops with fetch. Thoughts?

I don't understand this part. I don't see `fetch_all` in the PR.

AlexWheeler:
*I would be inclined to be explicit about the options we support and where we support them*
That totally makes sense to me.  I'll check to see if any others make sense too.

*I don't see fetch_all in the PR.*
You are correct, that's a typo on my part, I meant `read_multi` which we're using in `get_multi`.

AlexWheeler:
It looks like no need for the custom read_multi nonsense I mentioned above.  Cache implements this and caches (such as MemoryStore) can implement it themselves if they want to optimize it.  So we're all good there, if the cache doesn't implement it it will just fall back to this default implementation via the looping read().

https://github.com/rails/rails/blob/master/activesupport/lib/active_support/cache.rb#L353

lib/flipper/adapters/cache_store.rb
```diff
@@ -0,0 +1,104 @@
...
+      # Public: The name of the adapter.
+      attr_reader :name
+
+      # Public
+      def initialize(adapter, cache, expires_in: nil)
+        @adapter = adapter
+        @name = :cache_store
+        @cache = cache
+        @write_options = {}
+        @write_options.merge!(expires_in: expires_in) if expires_in
```
AlexWheeler: ActiveSupport::Cache will use ttl set on the `@cache` getting passed into this initializer, but allows you to set a ttl for a given key when using fetch/write.  Doing it this way we won't clobber the already set ttl, but still allow users to use a different ttl for flipper queries.

Its also possible that users won't even want to use a different ttl here and will always want to use whatever they're cache store is using when they initialize it?

AlexWheeler:
The other options described [here](http://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html) are:
:force 
:compress
:race_condition_ttl

and I don't think it makes sense to let users set these at the moment just for flipper stuff.

:namespace can be set as well, but we don't want to change the namespace they've already set on their cache store!  Instead flipper keys will have their own namespace tacked on  the same way the other cache adapters work.

PR #268, opened 2017-06-14
Add CacheStore docs with examples of cache + adapter

AlexWheeler:
Addresses #266 

Adds CacheStore Docs

PR #269, opened 2017-06-14
Add Flipper Cache Store .gemspec

### LINKED ISSUE #267: Create gemspec for cache-store
Each of the other adapters have been pulled into their own gem with their own .gemspecs + deps.  We'll want to create a .gemspec for cache store with an activesupport dependency right?

AlexWheeler:
fixes #267 

* Has ActiveSupport runtime dependency

Docs PR up #268

### DIFF
--- a/flipper-cache-store.gemspec
+++ b/flipper-cache-store.gemspec
@@ -1,0 +1,24 @@
+# -*- encoding: utf-8 -*-
+require File.expand_path('../lib/flipper/version', __FILE__)
+
+flipper_cache_store_files = lambda do |file|
+  file =~ /cache_store/
+end
+
+Gem::Specification.new do |gem|
+  gem.authors       = ['John Nunemaker']
+  gem.email         = ['nunemaker@gmail.com']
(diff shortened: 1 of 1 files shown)


flipper-cache-store.gemspec
```diff
@@ -0,0 +1,24 @@
...
+require File.expand_path('../lib/flipper/version', __FILE__)
+
+flipper_cache_store_files = lambda do |file|
+  file =~ /cache_store/
+end
+
+Gem::Specification.new do |gem|
+  gem.authors       = ['John Nunemaker']
+  gem.email         = ['nunemaker@gmail.com']
+  gem.summary       = 'ActiveSupport: Cache::Store adapter for Flipper'
```
jnunemaker: Any particular reason this is different from description other than the warning when building the gem from the gemspec? If not, I'd probably let them be identical.

AlexWheeler: nope, making them identical! Must have not had my coffee this am

PR #270, opened 2017-06-20
Remove GruopNotRegistered error and doc

### LINKED ISSUE #262: GroupNotRegistered never raised
Have been reading through the library and don't see `GroupNotRegistered` raised anywhere.

According to [this comment](https://github.com/jnunemaker/flipper/blob/c2b7a7e126accb92db0bce7bc8206536701bc69d/lib/flipper/dsl.rb#L220) I think it's supposed to be raised in `Flipper.group` if the group is not present. [Here](https://github.com/jnunemaker/flipper/blob/75ec873585706d403393bd50a139cc960a03b0fd/lib/flipper.rb#L69) maybe? From the looks of it though the real expected behavior is that a `Types::Group` object with a name and no block to execute should be returned.

AlexWheeler:
fixes #262

* This error is no longer raised since it was removed in commit 0d314c54affc2077806a0cc62cfd1d043381afe1.

### DIFF
--- a/lib/flipper/dsl.rb
+++ b/lib/flipper/dsl.rb
@@ -217,7 +217,6 @@ module Flipper
     # name - The String or Symbol name of the feature.
     #
     # Returns an instance of Flipper::Group.
-    # Raises Flipper::GroupNotRegistered if group has not been registered.
     def group(name)
       Flipper.group(name)
     end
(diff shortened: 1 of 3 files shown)


jnunemaker:
Thanks!

PR #274, opened 2017-06-27
Support percentages with up to 3 decimals

### LINKED ISSUE #273: Support more precision with percentage of actors/time
rollout [supports percentages with decimal points](https://github.com/fetlife/rollout/pull/100). It would be nice for flipper as well. For apps with a whole lotta actors, 1% can be too aggressive.

jnunemaker:
Would love some eyes on this to make sure I'm doing it right or if there are any downsides with the approach.

closes #273

### DIFF
--- a/examples/percentage_of_actors.rb
+++ b/examples/percentage_of_actors.rb
@@ -19,7 +19,7 @@ class User
   alias_method :flipper_id, :id
 end
 
-total = 10_000
+total = 100_000
 
 # create array of fake users
 users = (1..total).map { |n| User.new(n) }
@@ -31,13 +31,12 @@ perform_test = lambda { |number|
     flipper[:stats].enabled?(user) ? true : nil
   }.compact
 
-  actual = (enabled.size / total.to_f * 100).round(2)
+  actual = (enabled.size / total.to_f * 100).round(3)
 
   puts "percentage: #{actual.to_s.rjust(6, ' ')} vs #{number.to_s.rjust(3, ' ')}"
 }
 
 puts "percentage: Actual vs Hoped For"
-
-[1, 5, 10, 20, 30, 40, 50, 60, 70, 80, 90, 95, 99, 100].each do |number|
+[0.001, 0.01, 0.1, 1, 5, 10, 20, 30, 40, 50, 60, 70, 80, 90, 95, 99, 100].each do |number|
   perform_test.call number
 end
--- a/lib/flipper/gates/percentage_of_actors.rb
+++ b/lib/flipper/gates/percentage_of_actors.rb
@@ -30,7 +30,9 @@ module Flipper
         if Types::Actor.wrappable?(context.thing)
           actor = Types::Actor.wrap(context.thing)
           id = "#{context.feature_name}#{actor.value}"
-          Zlib.crc32(id) % 100_000 < percentage * 1_000
+          # this is to support up to 3 decimal places in percentages
+          scaling_factor = 1_000
+          Zlib.crc32(id) % (100 * scaling_factor) < percentage * scaling_factor
         else
           false
         end
--- a/lib/flipper/gates/percentage_of_actors.rb
+++ b/lib/flipper/gates/percentage_of_actors.rb
@@ -30,7 +30,7 @@ module Flipper
         if Types::Actor.wrappable?(context.thing)
           actor = Types::Actor.wrap(context.thing)
           id = "#{context.feature_name}#{actor.value}"
-          Zlib.crc32(id) % 100 < percentage
+          Zlib.crc32(id) % 100_000 < percentage * 1_000
         else
           false
         end
--- a/spec/flipper/gates/percentage_of_actors_spec.rb
+++ b/spec/flipper/gates/percentage_of_actors_spec.rb
@@ -7,10 +7,10 @@ RSpec.describe Flipper::Gates::PercentageOfActors do
     described_class.new
   end
 
-  def context(integer, feature = feature_name, thing = nil)
+  def context(percentage_of_actors_value, feature = feature_name, thing = nil)
     Flipper::FeatureCheckContext.new(
       feature_name: feature,
-      values: Flipper::GateValues.new(percentage_of_actors: integer),
+      values: Flipper::GateValues.new(percentage_of_actors: percentage_of_actors_value),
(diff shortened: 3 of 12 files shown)


spec/flipper/gates/percentage_of_actors_spec.rb
```diff
@@ -51,5 +49,29 @@ def context(integer, feature = feature_name, thing = nil)
...
+
+      let(:actors) do
+        (1..number_of_actors).map { |n| Flipper::Actor.new(n) }
+      end
+
+      subject { described_class.new }
+
+      it 'enables feature for accurate number of actors' do
+        margin_of_error = 0.02 * number_of_actors
+        expected_open_count = 100_000 * decimal
```
jqr: Should this be `number_of_actors * decimal`? If not where is `100_000` coming from?

spec/flipper/gates/percentage_of_time_spec.rb
```diff
@@ -6,4 +6,38 @@
   subject do
     described_class.new
   end
+
+  def context(integer, feature = feature_name, thing = nil)
+    Flipper::FeatureCheckContext.new(
+      feature_name: feature,
+      values: Flipper::GateValues.new(percentage_of_actors: integer),
```
jqr: Should this be `percentage_of_time`?

spec/flipper/gates/percentage_of_actors_spec.rb
jnunemaker: Ha, yeah I think so. Good catch!

spec/flipper/gates/percentage_of_time_spec.rb
jnunemaker: Derp.

lib/flipper/gates/percentage_of_actors.rb
```diff
@@ -30,7 +30,7 @@ def open?(context)
         if Types::Actor.wrappable?(context.thing)
           actor = Types::Actor.wrap(context.thing)
           id = "#{context.feature_name}#{actor.value}"
-          Zlib.crc32(id) % 100 < percentage
```
jqr: This is on the wrong thing. Disregard! :)

lib/flipper/gates/percentage_of_actors.rb
```diff
@@ -30,7 +30,7 @@ def open?(context)
...
+          Zlib.crc32(id) % 100_000 < percentage * 1_000
```
jqr: I think a comment or breaking out the scaling factor (1000) might be useful here. (this is on the right thing!)

spec/flipper/gates/percentage_of_actors_spec.rb
jnunemaker: Fixed in dd85fe9 for both specs

spec/flipper/gates/percentage_of_time_spec.rb
jnunemaker: Fixed in dd85fe9

lib/flipper/gates/percentage_of_actors.rb
jnunemaker: Tweaked in bba3969f8aff09717ec007a2adcbcb530c237f62. Do you think that helps? Enough? Any other suggestions?

jqr: Perfect!

lib/flipper/gates/percentage_of_actors.rb
```diff
@@ -30,7 +30,9 @@ def open?(context)
         if Types::Actor.wrappable?(context.thing)
           actor = Types::Actor.wrap(context.thing)
           id = "#{context.feature_name}#{actor.value}"
-          Zlib.crc32(id) % 100 < percentage
+          # this is to support up to 3 decimal places in percentages
+          scaling_factor = 1_000
+          Zlib.crc32(id) % (100 * scaling_factor) < percentage * scaling_factor
```
AlexWheeler: 👍

AlexWheeler: Out of pure curiosity did ever look at any other hash functions?

examples/percentage_of_actors.rb
```diff
@@ -19,7 +19,7 @@ def initialize(id)
   alias_method :flipper_id, :id
 end
 
-total = 10_000
+total = 100_000
```
AlexWheeler: awesome, will probably want to update [percentage_of_time example](https://github.com/jnunemaker/flipper/blob/master/examples/percentage_of_time.rb) too

spec/integration_spec.rb
```diff
@@ -411,10 +434,27 @@
       end
     end
 
-    context 'for not enabled percentage of time' do
+    context 'for enabled float percentage of time' do
       before do
         # ensure percentage of time returns percentage that makes five percent
```
AlexWheeler: super minor, but should this now say *that makes 4.1 percent* instead of five since its being enabled for 4.1% now

spec/flipper/gates/percentage_of_actors_spec.rb
```diff
@@ -51,5 +49,29 @@ def context(integer, feature = feature_name, thing = nil)
...
+      let(:number_of_actors) { 10_000 }
+
+      let(:actors) do
+        (1..number_of_actors).map { |n| Flipper::Actor.new(n) }
+      end
+
+      subject { described_class.new }
+
+      it 'enables feature for accurate number of actors' do
+        margin_of_error = 0.02 * number_of_actors
```
AlexWheeler: These tests are great.  Also curious, any specific reason you chose 2% margin of error? For fun running it on my machine 100 times with 0.01 and getting 0 errors

jnunemaker: Mostly just a "this felt like an ok error margin for the number of iterations". Since I've increased the number of iterations, I would probably be ok with increasing the error margin as it is likely lower.

spec/integration_spec.rb
jnunemaker: fixed in 2b973945eb795fc6faba64b5be246f814c203b8a. Thanks!

examples/percentage_of_actors.rb
jnunemaker: fixed in ece70a37689f38d7c95b734ba5d0916d2b646f3b. Great idea.

PR #281, opened 2017-08-26
Improve randomness uniformity

### LINKED ISSUE #272: Improve Randomness Uniformity
Over in rollout, [they tweaked](https://github.com/fetlife/rollout/pull/120) the percentage of actors feature from modulus to ([source](https://github.com/fetlife/rollout/issues/58#issuecomment-307970042)):

> Doing a < or > vs an appropriate boundary in the 32 bit integer space

Might be nice to port this to flipper as well.

rodrigopinto:
This PR is for improving the randomness calculation for the percentage of actors. It's a port of the implementation on https://github.com/fetlife/rollout/pull/120. 

As the new implementation reduces the fails in a really small decimal part, I haven't added a spec for this scenario. I also believe that it could become an expensive spec.

On the rollout implementation, [there is a benchmark exposing](https://github.com/fetlife/rollout/pull/120#issuecomment-308239660) the approximate accuracy of this implementation. 

Closes #272

### DIFF
--- a/lib/flipper/gates/percentage_of_actors.rb
+++ b/lib/flipper/gates/percentage_of_actors.rb
@@ -25,17 +25,15 @@ module Flipper
       #
       # Returns true if gate open for thing, false if not.
       def open?(context)
+        return false unless Types::Actor.wrappable?(context.thing)
+
         percentage = context.values[key]
 
-        if Types::Actor.wrappable?(context.thing)
-          actor = Types::Actor.wrap(context.thing)
-          id = "#{context.feature_name}#{actor.value}"
-          # this is to support up to 3 decimal places in percentages
-          scaling_factor = 1_000
-          Zlib.crc32(id) % (100 * scaling_factor) < percentage * scaling_factor
-        else
-          false
-        end
+        actor = Types::Actor.wrap(context.thing)
+        id = "#{context.feature_name}#{actor.value}"
+        # this is to support up to 3 decimal places in percentages
+        scaling_factor = 1_000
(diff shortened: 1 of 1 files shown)


jnunemaker:
Hey thanks! Sorry I haven't responded to this. I was gone on vacation. Thank you so much for tackling it! I really apprecate the help. I'll try to look it over this weekend or first thing next week. 

One question: will I need to note this as a breaking change in the changelog? I'm assuming that individual actors who were enabled before, may or may not be following this change. I understand that the same percentage will be enabled, but would upgrading flipper with any % of actors enablements were not 100% cause the set of enabled actors to differ? Just want to make sure I note it correctly for people. 

I'll probably test it with 100 or 1k actors or something and get the enabled ones before this change and after and then do a set diff. If you have time to tackle that and drop the work here in a comment, it would help getting this merged faster, but if not I'm happy to look at it as soon as I have time.

rodrigopinto:
Hey @jnunemaker, so, it was my turn to be on vacation this weekend, sorry for the delayed answer. 

As you mentioned, the percentage will probably be similar, but with a really small decimal difference. In few cases, it will be higher & other cases lower than the old implementation %, so it makes sense to mention on the changelog this possibility.

I can do the test batch, just give me some time, so I  work on it and drop the result here. Not sure about this weekend, but I do my best to get it done asap. :-)

jnunemaker:
@rodrigopinto sounds great. If I get a chance I'll whip something together first, but if not I'd love the help. Thanks!

rodrigopinto:
It has a really long time since I opened this PR and it got stale, so I will close it but keep the fork implementation in case I come back to work with this lib again.

PR #284, opened 2017-09-07
Test against ruby 2.4 travis.yml

AlexWheeler:
* Keep tests up to date with Ruby versions.

*  Ran into issues when trying to run master build (.e.g https://travis-ci.org/jnunemaker/flipper/builds/272896851).  Seeing error with: `mongodb: unrecognized service`.  Removing this from the .travis.yml seemed to fix the problem, but I don't have context on why that status check is there, so maybe there's a better solution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ rvm:
   - 2.1
   - 2.2
   - 2.3
+  - 2.4
 bundler_args: --without guard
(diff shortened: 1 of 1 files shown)


AlexWheeler:
looks like might not be able to support 2.4 because of a dependency. closing to look into it when get a chance

PR #295, opened 2017-10-21
cache-store => cache_store

jnunemaker:
This matches active_record and ruby idioms in general. Mixing dash and underscore is a bit more ugly but matches [rubygems guidelines](http://guides.rubygems.org/name-your-gem/):

```
net-http-digest_auth	require 'net/http/digest_auth'	Net::HTTP::DigestAuth
```


jnunemaker:
Just realized why this was this way. Someone else owns the flipper-cache_store gem...

PR #298, opened 2017-11-04
All The Get Alls

jnunemaker:
In testing `preload_all` with Flipper::Cloud, I found that the memoizing and caching adapters weren't working correctly with `get_all` (which `preload_all` uses). Several things were subtly not right causing extra net/http calls. I added several specs that verify the number of network calls and tested externally with a few apps I control. This also adds get_all to all the adapters flipper supports out of the box. This ensures that when possible `preload_all` and `get_all` do the fewest network calls. 

There is some funny business with caching and get_all. The trick with get_all and caching is you don't know the features so you can't do a get_multi for all features. This left two obvious options:

1. wrap the entire get_all call with its own cache fetch block.
2. add an extra network call to avoid calling get_all more than necessary when using a caching adapter.

Option 1 meant get_all would be 1 network call which was dope. The downside was it would not share the primed caches from `get` and `get_multi`. Additionally, it would mean that all features would need to fit in a single cache value (usually 1MB with memcached). Probably fine, but I didn't want to make assumptions.

Option 2 was to do an unless exists `get_all` key. If the key was present, then we are moderately sure that both the set of feature keys and each feature is cached, so the cache adapters do at least 2 network calls (get the set of feature keys and then get_multi all the individual feature cache values). The upside is this shares the per feature cache values with get and get_multi. Another upside is that because the cache is per feature, having a lot of features won't overflow a single value like option 1. The downside of this is a minimum of 3 network calls. I guess I'm assuming that cache calls are super fast, so 3 small ones that share between many adapter methods is better than potentially 1 big one that doesn't share.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -40,3 +40,6 @@ Style/TrailingCommaInLiteral:
 
 RSpec/InstanceVariable:
   Enabled: false
+
+Style/AccessorMethodName:
+  Enabled: false
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -86,6 +86,17 @@ module Flipper
         result
       end
 
+      def get_all
+        db_gates = @gate_class.joins("INNER JOIN #{@feature_class.table_name} on #{@feature_class.table_name}.key = #{@gate_class.table_name}.feature_key").all.to_a
+        grouped_db_gates = db_gates.group_by(&:feature_key)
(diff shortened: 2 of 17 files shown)


jnunemaker:
Looks like rails 3.2 failures... I'll look at them, get this merged and release 0.11

jnunemaker:
It seems unless_exist wasn't added until rails 4, which is why rails 3 tests are failing.

xref https://github.com/rails/rails/commit/b5005b6259144eeea5b903d659f54bd8a0343d25

Issue #63, opened 2015-04-27
Allow customizing UI text


jnunemaker:
Each gates text/help/instructions should be customizable so companies/developers can make it feel more like there own. I can also see providing an actor to instance conversion block for changing things like `User:6` to `@jnunemaker` or whatever makes sense.

ptrippett:
We use [rails global ids](https://github.com/rails/globalid) as the flipper_id value which could make this pretty trivial to implement

AlexWheeler:
@jnunemaker what are your thoughts on doing this via the configure block?  similar to how the adapter can be set via [Flipper.configure](https://github.com/jnunemaker/flipper/commit/95bba012463c2a781d10d7a5641b7d6fde2bc7f8#diff-b5a75b6f83af650cc606d47af1581d85R16).

We could have a Flipper::UI.configure.  The downside is *anyone* couldn't customize since it would need to be set in the initializer, but it could work well.

jnunemaker:
@AlexWheeler I like that for anything that needs to be customized. 👍

jnunemaker:
This is done. UI text can now be configured.

jnunemaker:
For more on configuring: https://github.com/jnunemaker/flipper/tree/master/docs/ui#configuration

PR #306, opened 2017-11-17
Implement UI::Configuration so users can customize Flipper UI text

AlexWheeler:
spike on #63 to allow users to customize the Flipper UI text, curious what everyone thinks.

```
Flipper::UI.configure do |c|
  c.percentage_of_actors.title = "% of Actors"
  c.percentage_of_actors.description = "percentage of actors rule :)"

  c.actors.description = "Enable any actor by entering that actor's flipper_id"

  c.delete.title = "DO NOT PRESS!"
  c.delete.description = "Who knows what will happen?!"
end
```

results in:
screen shot 2017-11-17 at 8 19 50 am

I also played with the idea of making the configuration have a bit cleaner dsl something like. the latter is cleaner, but the above is simpler, would be cool to hear thoughts on this:

```
Flipper::UI.configure do |c|

  c.actors do
    title ""
    description ""
  end
  
end
```

### DIFF
--- a/lib/flipper/ui.rb
+++ b/lib/flipper/ui.rb
@@ -12,6 +12,7 @@ require 'flipper'
 require 'flipper/middleware/setup_env'
 require 'flipper/middleware/memoizer'
 require 'flipper/ui/middleware'
+require 'flipper/ui/configuration'
 
 module Flipper
   module UI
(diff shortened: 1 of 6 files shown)


jnunemaker:
I hit submit too soon. Lol. 

> I also played with the idea of making the configuration have a bit cleaner dsl something like. the latter is cleaner, but the above is simpler, would be cool to hear thoughts on this:

I like the way you did it. I'm a fan of pure/boring ruby over DSL's.

AlexWheeler:
@jnunemaker great idea, I'll add move feature_creation_enabled / application_breadcrumb_href to UI::Configuration in next PR.

what do you think about the failing line length code climate issues for the strings.  splitting them into multi lines to make code climate happy makes it messier imo

jnunemaker:
I’m fine either way on line length. I usually don’t worry about stuff like
that.

PR #315, opened 2017-12-04
Remove ui/public/ from code climate results

AlexWheeler:
* Code Climate is analyzing files under lib/flipper/ui/public/ which is
mostly vendored js/css giving us an F!

screen shot 2017-12-04 at 7 25 14 am


jnunemaker:
Thanks for getting this started. Excited to see where we can improve based on code climate info. I haven’t used it before.

PR #317, opened 2017-12-22
Allow to use redis-adapter with redis 4.0.x

jakubkosinski:
As [redis-rb 4.0.1](https://github.com/redis/redis-rb/releases/tag/v4.0.1) is now released, why don't allow to use it with flipper-redis adapter?


jnunemaker:
I don't see anything in the changelog that is worrysome and flipper's usage of this gem is pretty boring so I'm game. Thanks!

PR #318, opened 2017-12-22
Adds assertion for importing boolean enabled features

AlexWheeler:
* the feature was enabled using source flipper, but wasn't being asserted that the feature
was properly imported to the destination flipper


jnunemaker:
🤘

Issue #308, opened 2017-11-17
Add Rollout adapter


jnunemaker:
[Rollout](https://github.com/fetlife/rollout/) is another popular feature flipping gem for ruby. It is built on redis (or whatever supports an identical interface to that). Though flipper has a few more features than rollout, it should be relatively easy to make an adapter that reads the rollout data and maps it to flipper gates. Having this adapter + our adapter importing means it would be a few lines of code to import rollout data into flipper and peel out.

PR #319, opened 2017-12-22
Rollout Adapter for importing rollout data into Flipper

AlexWheeler:
Addresses #308 

Adds an adapter to import Rollout data into Flipper

Rollout supports the equivalent to Flipper's
* groups
* actors
* percentage_of_actors
* boolean (sort of...via 100% percentage enabled)

Rollout does not support
* percentage_of_time

Importing only requires `features`, `get` to be implemented.

Questions/Notes:

* for the methods the adapter doesn't implement I'm thinking of raising an error similar to the read only adapter to make it clear that this adapter is only for importing since rollout doesn't have feature parity

* I'm testing ["can have one feature imported"`](https://github.com/jnunemaker/flipper/compare/rollout-adapter?expand=1#diff-ce5ff96d129103e0d6d194b6be39790dR59) type stuff where its not really testing the rollout adapter its really testing adapter.import, but I think it does give some assurance that things are being imported correctly.  Alternatively we could just assume import works given that `get`, `features` work which are also unit tested, but I do like the more thorough test.

* this will only work with rollout ~> 2.0 because the Rollout public api changed from 1 -> 2 and this is reflected in the [gemspec dependency](https://github.com/jnunemaker/flipper/compare/rollout-adapter?expand=1#diff-b6812c21ec89079fd92064a7e29a5673R24).

* Rollout 2.3[ introduced casting percentages](https://github.com/fetlife/rollout/commit/ac5f55bfd5aea59bf7c4c9cea189cacd49b7d290#diff-201938e8c8fb9ba71dea56ce449c387f) `to_f` before saving.  For flipper users using Flipper < 0.11 (before Flipper supported decimal percentages) the values between Rollout (decimal) and Flipper(integer) would be different, although I don't think this is a big deal?

* users importing Rollout data would need to reregister their groups since those are an evaluated block in memory

### DIFF
--- a/flipper-rollout.gemspec
+++ b/flipper-rollout.gemspec
@@ -1,0 +1,25 @@
+# -*- encoding: utf-8 -*-
+require File.expand_path('../lib/flipper/version', __FILE__)
+
+flipper_rollout_files = lambda do |file|
+  file =~ /rollout/
+end
+
+Gem::Specification.new do |gem|
+  gem.authors       = ['John Nunemaker']
+  gem.email         = ['nunemaker@gmail.com']
+  gem.summary       = 'Rollout adapter for Flipper'
+  gem.description   = 'Rollout adapter for Flipper'
+  gem.license       = 'MIT'
+  gem.homepage      = 'https://github.com/jnunemaker/flipper'
+
+  gem.files         = `git ls-files`.split("\n").select(&flipper_rollout_files) + ['lib/flipper/version.rb']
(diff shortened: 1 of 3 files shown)


jnunemaker:
I can drop a new release after this merges or wait and we can put some docs in. I am leaning toward 0.11.1 instead of 0.12 since this would just be an addition. @AlexWheeler What do you think?

AlexWheeler:
@jnunemaker I'll open another issue to document this and hopefully get to that this week

I think its all good to release new version then get docs up. I'd say 0.12 would be the correct semver since as you mention its an addition and not a bugfix, but 0.11.1 is cool too.

PR #322, opened 2018-01-04
Allow feature removal from the UI to be disabled

mateusg:
This change allows the `Danger Zone` + `Delete Feature` button to be removed from the Flipper UI.

I thought it was a good idea do have it, because sometimes adding or removing features is a job for developers only, who might want to do it only via console, rake tasks, migrations or seeds.

What do you guys think?

### DIFF
--- a/lib/flipper/ui.rb
+++ b/lib/flipper/ui.rb
@@ -24,14 +24,19 @@ module Flipper
 
       # Public: Is feature creation allowed from the UI? Defaults to true. If
       # set to false, users of the UI cannot create features. All feature
-      # creation will need to be done through the conigured flipper instance.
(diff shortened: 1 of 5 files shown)


lib/flipper/ui.rb
```diff
@@ -24,14 +24,19 @@ class << self
 
       # Public: Is feature creation allowed from the UI? Defaults to true. If
       # set to false, users of the UI cannot create features. All feature
-      # creation will need to be done through the conigured flipper instance.
+      # creation will need to be done through the configured flipper instance.
```
jnunemaker: Nice! Thanks.

mateusg:
:+1:

Issue #324, opened 2018-01-05
When preload_all is enabled, disabled features don't show up in the flipper-ui


geetotes:
I'm using Rails 5 and my `config/initializers/flipper.rb` looks like this:

```
require 'flipper'
require 'flipper/adapters/active_record'
require 'flipper/adapters/active_support_cache_store'

Flipper.configure do |config|
  config.default do
    adapter = Flipper::Adapters::ActiveRecord.new
    cached_adapter = Flipper::Adapters::ActiveSupportCacheStore.new(
      adapter,
      Rails.cache,
      expires_in: 10.seconds
    )
    Flipper.new(cached_adapter)
  end
end

require 'flipper/middleware/memoizer'
Rails.configuration.middleware.use Flipper::Middleware::Memoizer, preload_all: true
```

When using the `preload_all: true` option, Flipper features are loaded with this query:
```  Flipper::Adapters::ActiveRecord::Gate Load (2.5ms)  SELECT `flipper_gates`.* FROM `flipper_gates` INNER JOIN flipper_features
          ON flipper_features.key = flipper_gates.feature_key```

Unfortunately, disabled features don't show up in this list, so when I visit the flipper-ui, my disabled features don't appear in the list of features available at the `/features` route. 

When I remove `preload_all` from my Memoizer options, I see my disabled features in flipper-ui

**Comment by jnunemaker** [comment]
@geetotes hi! Nothing comes to mind. What version of flipper, flipper-ui and flipper-active_record are you on? Could you print out your middleware stack (`bin/rails middleware` I think)? Also, what version of rails? 

The only other thing I can think is that the UI also includes the memoizer. Shouldn't hurt anything as it does no preloading. Some other debugging data that could help would be selecting all the rows from the flipper_features and flipper_gates tables so I could try to re-create your problem.

**Comment by geetotes** [comment]
I'm running Rails 5.1.4. From my gemfile.lock:
```
    flipper (0.11.0)
    flipper-active_record (0.11.0)
      flipper (~> 0.11.0)
    flipper-active_support_cache_store (0.11.0)
      flipper (~> 0.11.0)
    flipper-ui (0.11.0)
      flipper (~> 0.11.0)
activerecord (5.1.4)
```
Here's my middleware stack:
```
use Webpacker::DevServerProxy
use Rack::Sendfile
use ActionDispatch::Static
use ActionDispatch::Executor
use ActiveSupport::Cache::Strategy::LocalCache::Middleware
[...]

PR #327, opened 2018-01-06
Active Record get_all Fix

jnunemaker:
Fixes the sequel and active record adapters to include even disabled features in get_all. The issue was using an inner join between feature and gate tables, which means features in the feature table but not in the gate table (because they were fully disabled) were not represented in get_all. 

fixes https://github.com/jnunemaker/flipper/issues/324

### DIFF
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -89,11 +89,11 @@ module Flipper
       end
 
       def get_all
-        joins = <<-EOJ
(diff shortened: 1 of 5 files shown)


jnunemaker:
I think I should probably make this an adapter shared spec/test so we can catch this for all of them.

jnunemaker:
The sequel adapter also has this problem. I'll move this to a shared test/spec for all adapters and then merge it.

Issue #307, opened 2017-11-17
Moneta Adapter


jnunemaker:
https://github.com/minad/moneta is a really great project. I think it would be great to add a moneta adapter. That would allow people to store feature data in a plethora of data stores with very little maintenance on our end. It might even be worth moving the official adapters that we maintain to just be moneta adapters to reduce maintenance (redis, etc.). Moneta also has an active support cache store adapter so it is possible to use it as a cache adapter currently. Might be good to document that.

PR #330, opened 2018-01-11
Moneta adapter

AlexWheeler:
#307 

I almost thought we could pass in a Moneta instance to the memory adapter as its source and it would work because the Moneta api is very similar to Hash, but it turns out Moneta returns copies of objects, which is why [set_add](https://github.com/jnunemaker/flipper/compare/moneta-adapter?expand=1#diff-975176d71feb497dd53ac5b6311196eaR145) and set_delete need to be implemented a little different than the memory adapter.  This is pretty much the memory adapter with a few tweaks.

* The earliest version that passes *flipper/adapters/moneta_spec.rb* is Moneta [0.7.19](https://github.com/jnunemaker/flipper/compare/moneta-adapter?expand=1#diff-f66ab524b960774a9c36616b5c0a9187R23). Not sure if we should require that dependency or 1.0? Would be cool to hear ya thoughts!

* moneta is very cool and this opens the door to so many data stores

### DIFF
--- a/lib/flipper/adapters/moneta.rb
+++ b/lib/flipper/adapters/moneta.rb
@@ -1,9 +1,11 @@
+require 'moneta'
+
 module Flipper
   module Adapters
     class Moneta
       include ::Flipper::Adapter
 
-      FeaturesKey = :features
+      FEATURES_KEY = :features
 
       # Public: The name of the adapter.
       attr_reader :name
@@ -20,16 +22,14 @@ module Flipper
 
       # Public: Adds a feature to the set of known features.
       def add(feature)
-        @moneta[FeaturesKey.to_s] ||= Set.new
-        @moneta[FeaturesKey.to_s] = @moneta[FeaturesKey.to_s].add(feature.key)
+        set_add(FEATURES_KEY, feature.key)
         true
       end
 
       # Public: Removes a feature from the set of known features and clears
       # all the values for the feature.
       def remove(feature)
-        @moneta[FeaturesKey.to_s] ||= Set.new
-        @moneta[FeaturesKey.to_s] = @moneta[FeaturesKey.to_s].delete(feature.key)
+        set_delete(FEATURES_KEY, feature.key)
         clear(feature)
         true
       end
@@ -55,7 +55,6 @@ module Flipper
         features = read_feature_keys.map do |key|
           Flipper::Feature.new(key, self)
         end
-
         read_many_features(features)
       end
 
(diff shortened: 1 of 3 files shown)


jnunemaker:
> The earliest version that passes flipper/adapters/moneta_spec.rb is Moneta 0.7.19. Not sure if we should require that dependency or 1.0? Would be cool to hear ya thoughts!

I would be fine with start at 1.0 and relaxing the constraint if people request it.

> moneta is very cool and this opens the door to so many data stores

Right? I have a similar jnunemaker/adapter project, but my thought was that moneta is actually maintained and supports far more data stores. It also supports expiration and things like that.

lib/flipper/adapters/moneta.rb
```diff
@@ -0,0 +1,160 @@
...
+      # Public: Adds a feature to the set of known features.
+      def add(feature)
+        set_add(FEATURES_KEY, feature.key)
+        true
+      end
+
+      # Public: Removes a feature from the set of known features and clears
+      # all the values for the feature.
+      def remove(feature)
+        set_delete(FEATURES_KEY, feature.key)
```
jnunemaker: It would be neat if moneta supported transactions for data stores that did so we could do these two operations in a transaction, but I doubt it will be a problem in practice.

lib/flipper/adapters/moneta.rb
```diff
@@ -0,0 +1,160 @@
...
+        clear(feature)
+        true
+      end
+
+      # Public: Clears all the gate values for a feature.
+      def clear(feature)
+        feature.gates.each do |gate|
+          delete key(feature, gate)
```
jnunemaker: Each of these operations will require a network call for any networked data store, correct? 

I was picturing everything we need for a feature stored in a single key/value (e.g. `"feature/#{feature.key}"` with value serialized by marshal or json or whatever moneta supports). It would increase the number of possible network calls on enable/disable/etc (read + modify value + write vs single write as you currently have it), but would reduce the number of keys to fetch when preloading (1 per feature instead of N per feature where N is the number of gates). 

I've tried to optimize reads (get, get_multi, get_all) in flipper since those are the hotter path (than enabled/disable/clear). How do you feel about switching this to a single key per feature? Any downsides I haven't considered? Only thing I can think of is perhaps concurrent writes, but we could always use Moneta's locking for that if necessary down the road.

lib/flipper/adapters/moneta.rb
```diff
@@ -0,0 +1,160 @@
...
+
+      # Private
+      def read_feature(feature)
+        result = {}
+
+        feature.gates.each do |gate|
+          result[gate.key] =
+            case gate.data_type
+            when :boolean, :integer
+              read key(feature, gate)
```
jnunemaker: Ah, actually same thought goes here as above. If we move to a single key per feature we can fetch a feature in one network call instead of N (one per gate).

lib/flipper/adapters/moneta.rb
AlexWheeler: That's a really great idea!  Definitely makes sense to be optimizing for reads.  Been busy past few weeks with some stuff, but going to spike on that and see how it goes. Off top of my head can't think of any other downsides, but will put more thought into it as I'm jamming

AlexWheeler:
Thanks for the great comments, closing in favor of #361 as I finally got around to jamming on this again on my flight to nyc last night

PR #332, opened 2018-01-20
Rollout tweaks

jnunemaker:
* Adds require for flipper/errors since `AdapterMethodNotSupportedError` inherits from Flipper::Error defined in that file.
* Ensures rollout works fine for features that do not exist. In some basic scripts I was playing with this was causing problems.
* Makes rollout percentage 100 equal to flipper fully enabled. Rollout doesn't clear other gate values (users/groups) when fully enabled and uses percentage == 100 as fully enabled instead of a different key (like boolean: false in flipper). Flipper.enabled?(:foo) returns false for 100% of actors enabled because no actor was passed, but rollout returns true (as it has no other way to denote 100% of actors enabled vs enabled regardless of actor). Making rollout 100% == flipper fully enabled means that rollout.active?(:chat) and flipper.enabled?(:chat) are equivalent, which feels right.
* Adds a basic and import example.
* Updates docs to match import example.

### DIFF
--- a/docs/rollout/README.md
+++ b/docs/rollout/README.md
@@ -23,18 +23,25 @@ Or install it yourself with:
 
 ## Usage
 
-Assuming you've configured a [default Flipper instance](https://github.com/jnunemaker/flipper#user-content-examples)
-
 ```ruby
 require 'redis'
+require 'rollout'
+require 'flipper'
+require 'flipper/adapters/redis'
 require 'flipper/adapters/rollout'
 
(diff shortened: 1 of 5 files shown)


jnunemaker:
Actually, I'm feeling antsy because I wrote a post, so I'm going to merge and release this, but I'm still 💯 to any feedback on it. 😄

AlexWheeler:
Nice! will for sure still check it out later today when I get some time 👍

PR #290, opened 2017-09-26
Display production mode warning

igorbozato:
Closes #101

PR #337, opened 2018-02-09
Add environment mode banner in UI

thetimbanks:
Closes https://github.com/jnunemaker/flipper/issues/101 and is a replacement for https://github.com/jnunemaker/flipper/pull/290

## Purpose
Show an information banner across the top of the page in order to differentiate environments or to just display some message.

## Notable Changes
- new configuration options for Flipper::UI::Configuration: `banner_text`, `banner_class`
- when `banner_text` is set a banner will show at the top of the screen showing the text entered
- the banner color (`banner_class`) can be changed based on the basic bootstrap color scheme (info, danger, success, etc)
- add configuration information to the UI README and screenshot

## Screenshot
image


thetimbanks:
The other option is to just allow the user to set a string value to `environment` and display that in the alert instead of `${environment} Environment`.

thetimbanks:
Another thought I had this morning was to keep the `environment` and `environment_banner_color` has `Options`. That way for the `environment` there could be a `title` and a `description`. Not sure what benefit it gives to the `environment_banner_color`, but it would at least stay consistent with the other configuration options.

lib/flipper/ui/views/layout.erb
```diff
@@ -16,7 +16,13 @@
     <![endif]-->
   </head>
   <body>
-    <div class="container header">
+    <div class="container">
+      <%- unless Flipper::UI.configuration.environment.nil? -%>
+        <div class="alert alert-<%= Flipper::UI.configuration.environment_banner_color %> text-center font-weight-bold">
+          <%= Flipper::UI.configuration.environment.capitalize %> Environment
```
jnunemaker: I like your idea of just letting environment be a string. Perhaps you could also change the options to `Flipper::UI.configuration.banner_text` and `Flipper::UI.configuration.banner_class`? That feels a little more generic so people can use it for whatever text and alert class they want. What do you think?

lib/flipper/ui/configuration.rb
```diff
@@ -9,12 +9,17 @@ class Configuration
...
+                    :environment_banner_color
+
       def initialize
         @actors = Option.new("Actors", "Enable actors using the form above.")
         @groups = Option.new("Groups", "Enable groups using the form above.")
         @percentage_of_actors = Option.new("Percentage of Actors", "Percentage of actors functions independently of percentage of time. If you enable 50% of Actors and 25% of Time then the feature will always be enabled for 50% of users and occasionally enabled 25% of the time for everyone.") # rubocop:disable Metrics/LineLength
         @percentage_of_time = Option.new("Percentage of Time", "Percentage of actors functions independently of percentage of time. If you enable 50% of Actors and 25% of Time then the feature will always be enabled for 50% of users and occasionally enabled 25% of the time for everyone.") # rubocop:disable Metrics/LineLength
         @delete = Option.new("Danger Zone", "Deleting a feature removes it from the list of features and disables it for everyone.") # rubocop:disable Metrics/LineLength
+        @environment = nil
+        @environment_banner_color = :danger
```
jnunemaker: Not a blocker for this PR by any means, but what do you think about defining the writer method here and enforcing valid values? Since bootstrap has a defined list of alert- classes, I'm wondering if we should make an array of the valid ones and verify that the one a user is trying to use is valid to avoid confusion if they pick a non-existent one and see no color.

spec/flipper/ui_spec.rb
```diff
@@ -153,5 +153,29 @@
         expect(config).to be_instance_of(Flipper::UI::Configuration)
       end
     end
+
+    describe 'environment' do
+      it 'does not include the banner if environment is not set' do
```
jnunemaker: Thanks for the great specs!

jnunemaker:
Also! If you don't mind, what are you using flipper for? Are you just contributing/making awesome or are you also using flipper in production on some app somewhere? If you aren't comfortable dropping this in a public issue, you can email me directly at john@featureflipper.com.

lib/flipper/ui/views/layout.erb
thetimbanks: I like your idea much better. Having it more generic will be more valuable.

lib/flipper/ui/configuration.rb
thetimbanks: Good idea. I'll see what I can come up with.

thetimbanks:
Thanks for the comments! I have some time today to get your comments resolved. I'll shoot you an email regarding the last comment.

thetimbanks:
@jnunemaker All updated based on your feedback.

PR #341, opened 2018-02-22
Sync Adapter

jnunemaker:
Adds a sync adapter. When you have a remote that is relatively latent (say http and not in same datacenter, region or whatever) and want a local "cache" that is more persistent than a cache (across process restarts and serves stale instead of raising if remote is unavailable), this adapter can come in handy.

* Reads go to local adapter (so reads are fast).
* Writes go to remote adapter and then to local adapter if remote successful. 
* Every N milliseconds, the local adapter will be diff'd against the remote and appropriate enable/disable calls will attempt to converge them.
* All syncing is done inline (nothing fancy in a background thread at this point). 
* Any errors during sync are swallowed (but instrumented for visibility) to avoid causing any disturbances to the request/job currently being performed.


jnunemaker:
I'm going to keep moving forward, but would love it if anyone would give this a look over.

Issue #326, opened 2018-01-06
Move Flipper::UI config to Flipper::UI::Configuration


jnunemaker:
Specifically, these 3 options:

https://github.com/jnunemaker/flipper/blob/49167082d655d07141622bced1f8ac98bf9f5904/lib/flipper/ui.rb#L20-L32

Add backwards incompatible note to changelog as well so people know how to upgrade. Another option would be to leave the reader and writer methods on Flipper::UI and make them raise with an error that describes the code changes necessary. We could leave those for a few versions.

AlexWheeler:
*Another option would be to leave the reader and writer methods on Flipper::UI and make them raise with an error that describes the code changes necessary*

I really like this idea, we can link to the docs too so users know exactly what to do

PR #345, opened 2018-03-31
Move Flipper::UI configuration options to Flipper::UI::Configuration

thetimbanks:
Resolves: https://github.com/jnunemaker/flipper/issues/326

Alex had the idea of linking to the docs to let the user know how to update the configuration, but I wasn't really sure what that would look like. I just mentioned in the exception that the configuration options moved to `Flipper::UI::Configure` which might be enough?

## Purpose
Now that Flipper::UI::Configuration exists, some configuration options on Flipper::UI should be moved over.

## Notable Changes
- Moved attribute_breadcrumb_href, feature_creation_enabled, and feature_removal_enabled from Flipper::UI to Flipper::UI::Configuration
- Deprecate the values on Flipper::UI by raising an exception

### DIFF
--- a/lib/flipper/errors.rb
+++ b/lib/flipper/errors.rb
@@ -29,4 +29,12 @@ module Flipper
       super(message || default)
     end
   end
+
+  # Raised when accessing a configuration property that has been deprecated
+  class ConfigurationDeprecated < Flipper::Error
+    def initialize(message = nil)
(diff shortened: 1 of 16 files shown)


thetimbanks:
I figure we can yank it in another release or two. I'll make a note to come back and do that. Thanks for the merge!

PR #353, opened 2018-04-30
Allow feature names to end with "features" in UI

### LINKED ISSUE #352: Flipper UI can't handle feature names containing _features
If the name contains _features, clicking on the name on the UI index page has no effect. 

screen shot 2018-04-30 at 12 13 57

In the screenshot, clicking on "another_features" or "production_features" does not take you to the edit page. (Admittedly, it's a bad name, anyway. But still.)

jnunemaker:
Because of the order of the action classes, the features index route was picked. I put it last so now the feature actions will get picked prior to the list.

Fixes #352

### DIFF
--- a/lib/flipper/ui/middleware.rb
+++ b/lib/flipper/ui/middleware.rb
@@ -16,7 +16,6 @@ module Flipper
         @action_collection = ActionCollection.new
 
         # UI
-        @action_collection.add UI::Actions::Features
         @action_collection.add UI::Actions::AddFeature
         @action_collection.add UI::Actions::Feature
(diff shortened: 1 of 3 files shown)


jnunemaker:
> My only question is are we likely to see this issue with any of the other UI::Actions?

All the other actions include the gate which makes them unique (e.g. /features/.../{boolean,actors,etc.}) so they should be ok. The only other odd one is that the form for a new feature is at /features/new, which means that if someone uses a feature named "new" the form will show up instead of the feature page. Not sure how to work around that outside of changing urls, but open to that if necessary.

PR #358, opened 2018-05-18
Reveal real boolean value for feature in UI

oboxodo:
Before this change, enabling a feature for 100% of actors or time would hide the "Enable" action for the Boolean Gate. After this change, the boolean gate actions can be used independently of the current value of the other gates.

A feature being enabled for 100% actors is not necessarily the same as enabling it for "everyone", depending on how you configure your actors. In a typical scenario, users are actors. Enabling a feature for 100% actors would enable it for 100% *logged in* users. But `Flipper.enabled?(:my_feature)` would return false for logged out users.

### DIFF
--- a/lib/flipper/ui/views/feature.erb
+++ b/lib/flipper/ui/views/feature.erb
@@ -14,7 +14,7 @@
         </h4>
         <div class="card-body">
           <p>
-            <% if @feature.on? %>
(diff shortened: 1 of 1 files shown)


jnunemaker:
That is a great point! Thanks for contributing.

PR #359, opened 2018-05-25
Introduce `disabled?` for easier disabled? condition checking

mad-raz:
Introduce `disabled?` for easier condition checking.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,8 @@
-## master (to be 0.15.0)
+## master
+
+* Introduce `Flipper.disabled?(:feature)` to use instead of `!Flipper.enabled?(:feature)` (https://github.com/jnunemaker/flipper/pull/359)
+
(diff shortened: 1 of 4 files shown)


jnunemaker:
Hi! Thanks for taking the time to contribute. I appreciate it. This has actually come up before: https://github.com/jnunemaker/flipper/pull/94#issuecomment-162917454. I’m inclined to stick to that response unless you or anyone else has strong feelings or thoughts on how to avoid that confusion.

mad-raz:
I sort of understand the philosophy behind keeping the API minimal, but there are some cases that reading `validate :restricted, if: Proc.new { |_| !Flipper.enabled?(:public_access) }`

in that case the condition will read better with `disabled?` than using `unless` (because unless will feel like double negation)...  which is the actual use case in our code (and the reason behind the PR)

PR #361, opened 2018-06-02
Moneta adapter each feature as key

AlexWheeler:
Adds Moneta adapter #307 

This is an improved PR over #330, where we're now optimizing for reads. Moneta has a Logging middleware that I used with a [simple script to log database accesses](https://gist.github.com/AlexWheeler/0a6562f3805db6c00e7931993a664017) and below are the results:

#330  (previous PR)
```
############### 3x enable(:feature) #####################

Moneta load("features") -> nil
Moneta store("features", #<Set: {"feature"}>) -> #<Set: {"feature"}>
Moneta load("feature/feature/actors") -> nil
Moneta store("feature/feature/actors", #<Set: {"1"}>) -> #<Set: {"1"}>
Moneta load("features") -> #<Set: {"feature"}>
Moneta store("features", #<Set: {"feature"}>) -> #<Set: {"feature"}>
Moneta load("feature/feature/actors") -> #<Set: {"1"}>
Moneta store("feature/feature/actors", #<Set: {"1", "2"}>) -> #<Set: {"1", "2"}>
Moneta load("features") -> #<Set: {"feature"}>
Moneta store("features", #<Set: {"feature"}>) -> #<Set: {"feature"}>
Moneta load("feature/feature/actors") -> #<Set: {"1", "2"}>
Moneta store("feature/feature/actors", #<Set: {"1", "2", "3"}>) -> #<Set: {"1", "2", "3"}>

############### feature.inspect #####################

Moneta load("feature/feature/boolean") -> nil
Moneta load("feature/feature/groups") -> nil
Moneta load("feature/feature/actors") -> #<Set: {"1", "2", "3"}>
Moneta load("feature/feature/percentage_of_...) -> nil
Moneta load("feature/feature/percentage_of_...) -> nil
Moneta load("feature/feature/boolean") -> nil
Moneta load("feature/feature/groups") -> nil
Moneta load("feature/feature/actors") -> #<Set: {"1", "2", "3"}>
Moneta load("feature/feature/percentage_of_...) -> nil
Moneta load("feature/feature/percentage_of_...) -> nil

############### feature.clear #######################

Moneta delete("feature/feature/boolean") -> nil
Moneta delete("feature/feature/groups") -> nil
Moneta delete("feature/feature/actors") -> #<Set: {"1", "2", "3"}>
Moneta delete("feature/feature/percentage_of_...) -> nil
Moneta delete("feature/feature/percentage_of_...) -> nil
```

This PR
```
############### 3x enable(:feature) #####################

Moneta load(:flipper_features) -> nil
Moneta store(:flipper_features, #<Set: {"feature"}>) -> #<Set: {"feature"}>
Moneta load("flipper_features/feature") -> nil
Moneta store("flipper_features/feature", {:boolean=>nil, :groups=>#<Set:...) -> {:boolean=>nil, :groups=>#<Set:...
Moneta load(:flipper_features) -> #<Set: {"feature"}>
Moneta store(:flipper_features, #<Set: {"feature"}>) -> #<Set: {"feature"}>
Moneta load("flipper_features/feature") -> {:boolean=>nil, :groups=>#<Set:...
Moneta store("flipper_features/feature", {:boolean=>nil, :groups=>#<Set:...) -> {:boolean=>nil, :groups=>#<Set:...
Moneta load(:flipper_features) -> #<Set: {"feature"}>
Moneta store(:flipper_features, #<Set: {"feature"}>) -> #<Set: {"feature"}>
Moneta load("flipper_features/feature") -> {:boolean=>nil, :groups=>#<Set:...
Moneta store("flipper_features/feature", {:boolean=>nil, :groups=>#<Set:...) -> {:boolean=>nil, :groups=>#<Set:...

############### feature.inspect #####################

Moneta load("flipper_features/feature") -> {:boolean=>nil, :groups=>#<Set:...
Moneta load("flipper_features/feature") -> {:boolean=>nil, :groups=>#<Set:...

############### feature.clear #######################

Moneta store("flipper_features/feature", {:boolean=>nil, :groups=>#<Set:...) -> {:boolean=>nil, :groups=>#<Set:...
```
*As you can see reads are much more performant!*

* keys are namespaced under `flipper_features`
`flipper[:name].enable` would store a `flipper_features/name` key in moneta to avoid a collision

would be cool to get some feedback on a few thoughts:

1. To store a value I'm using the `[]=(key, value)` method. The docs state that keys stored this way will never expire, however I set an expiration on the store and set a key this way with the Memory adapter and it did expire. I need to do a bit more reading...

2. there's a similar method to `[]=` called `create(key, value)` which creates an entry atomically.  Not all stores support it, but we could check if a store does support it and favor that over `[]=`.
```
if moneta.supports?(:create)
  moneta.create(key, value)
else
  moneta[key] = value
end
```
Was hoping it would just fallback to using `[]=`, but it looks like it [will raise an error if not supported](https://github.com/minad/moneta/blob/master/lib/moneta/mixins.rb#L240)

3. There's another method for adding keys `#store(key, value, options = {})`, which accepts options, one being `expires:` If users are using moneta as a cache they might want to set expiration on writes:
```
moneta.store(key, value, expires: 200)
```
So we could optionally accept options in the Moneta initializer and pass these onto the write methods, however it also seems that a moneta user can set an expiration on the store when they initialize their moneta instance so I believe it makes sense to just let them do that and not worry about passing options

### DIFF
--- a/flipper-moneta.gemspec
+++ b/flipper-moneta.gemspec
@@ -20,5 +20,5 @@ Gem::Specification.new do |gem|
   gem.version       = Flipper::VERSION
 
   gem.add_dependency 'flipper', "~> #{Flipper::VERSION}"
-  gem.add_dependency 'moneta', '>= 0.7.19'
+  gem.add_dependency 'moneta', '>= 0.7.0', '<= 1.0.0'
 end
--- a/flipper-moneta.gemspec
+++ b/flipper-moneta.gemspec
@@ -1,0 +1,24 @@
+# -*- encoding: utf-8 -*-
+require File.expand_path('../lib/flipper/version', __FILE__)
+
+flipper_moneta_files = lambda do |file|
+  file =~ /moneta/
+end
+
+Gem::Specification.new do |gem|
+  gem.authors       = ['John Nunemaker']
+  gem.email         = ['nunemaker@gmail.com']
+  gem.summary       = 'Moneta adapter for Flipper'
+  gem.description   = 'Moneta adapter for Flipper'
+  gem.license       = 'MIT'
+  gem.homepage      = 'https://github.com/jnunemaker/flipper'
+
+  gem.files         = `git ls-files`.split("\n").select(&flipper_moneta_files) + ['lib/flipper/version.rb']
+  gem.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n").select(&flipper_moneta_files)
+  gem.name          = 'flipper-moneta'
+  gem.require_paths = ['lib']
+  gem.version       = Flipper::VERSION
+
+  gem.add_dependency 'flipper', "~> #{Flipper::VERSION}"
+  gem.add_dependency 'moneta', '>= 0.7.19'
+end
--- a/lib/flipper/adapters/moneta.rb
+++ b/lib/flipper/adapters/moneta.rb
@@ -23,16 +23,14 @@ module Flipper
 
       # Public: Adds a feature to the set of known features.
       def add(feature)
-        moneta[FEATURES_KEY] = (moneta[FEATURES_KEY] || Set.new) << feature.key.to_s
+        moneta[FEATURES_KEY] = features << feature.key.to_s
         true
       end
 
       # Public: Removes a feature from the set of known features and clears
       # all the values for the feature.
       def remove(feature)
-        features = moneta[FEATURES_KEY] || Set.new
-        features.delete(feature.key.to_s)
-        moneta[FEATURES_KEY] = features
+        moneta[FEATURES_KEY] = features.delete(feature.key.to_s)
         moneta[key(feature.key)] = default_config
         true
       end
@@ -47,14 +45,14 @@ module Flipper
       #
       # Returns a Hash of Flipper::Gate#key => value.
       def get(feature)
-        moneta[key(feature.key)] || default_config
+        default_config.merge(moneta[key(feature.key)].to_h)
       end
 
       # Public
       def get_multi(features)
         result = {}
         features.each do |feature|
-          result[feature.key] = default_config.merge(moneta[key(feature.key)].to_h)
+          result[feature.key] = get(feature)
         end
         result
       end
@@ -63,7 +61,7 @@ module Flipper
       def get_all
         result = {}
         moneta[FEATURES_KEY].each do |feature_key|
-          result[feature_key] = default_config.merge(moneta[key(feature_key)].to_h)
+          result[feature_key] = get(Feature.new(feature_key, self))
         end
         result
       end
@@ -78,12 +76,11 @@ module Flipper
       def enable(feature, gate, thing)
         case gate.data_type
         when :boolean, :integer
-          result = moneta[key(feature.key)] || {}
+          result = get(feature)
           result[gate.key] = thing.value.to_s
           moneta[key(feature.key)] = result
         when :set
-          result = moneta[key(feature.key)] || {}
-          result[gate.key] ||= Set.new
+          result = get(feature)
           result[gate.key] << thing.value.to_s
           moneta[key(feature.key)] = result
         end
@@ -102,12 +99,12 @@ module Flipper
         when :boolean
           moneta[key(feature.key)] = default_config
         when :integer
-          result = moneta[key(feature.key)] || {}
+          result = get(feature)
           result[gate.key] = thing.value.to_s
           moneta[key(feature.key)] = result
         when :set
-          result = moneta[key(feature.key)]
-          result[gate.key] = result[gate.key].delete(thing.value.to_s) if result[gate.key]
+          result = get(feature)
+          result[gate.key] = result[gate.key].delete(thing.value.to_s)
           moneta[key(feature.key)] = result
         end
         true
--- a/lib/flipper/adapters/moneta.rb
+++ b/lib/flipper/adapters/moneta.rb
@@ -1,9 +1,12 @@
+require 'moneta'
+require 'pry'
+
 module Flipper
   module Adapters
     class Moneta
       include ::Flipper::Adapter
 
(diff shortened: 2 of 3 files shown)


lib/flipper/adapters/moneta.rb
```diff
@@ -0,0 +1,122 @@
...
+      def enable(feature, gate, thing)
+        case gate.data_type
+        when :boolean, :integer
+          result = get(feature)
+          result[gate.key] = thing.value.to_s
+          moneta[key(feature.key)] = result
+        when :set
+          result = get(feature)
+          result[gate.key] << thing.value.to_s
+          moneta[key(feature.key)] = result
```
AlexWheeler: possible race condition here, but since the Moneta key points to a plain old ruby hash we need to update the hash in memory and then reset the key

lib/flipper/adapters/moneta.rb
```diff
@@ -0,0 +1,122 @@
...
+        moneta[FEATURES_KEY] = features << feature.key.to_s
+        true
+      end
+
+      # Public: Removes a feature from the set of known features and clears
+      # all the values for the feature.
+      def remove(feature)
+        moneta[FEATURES_KEY] = features.delete(feature.key.to_s)
+        moneta[key(feature.key)] = default_config
+        true
```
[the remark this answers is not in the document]
jnunemaker: I was thinking they had support for locking, which could get the job done, but that seems to just be for thread safety, not across processes. No biggie. We can note that it isn't transactional and likely move on with life. 😄

lib/flipper/adapters/moneta.rb
jnunemaker: Yep, no problem. The KV branch I have going takes the same approach. Writes are rare so this seems totally fine to me. If people can't live with this, they can create a locking adapter to wrap moneta with some form of locking to avoid the race.

lib/flipper/adapters/moneta.rb
```diff
@@ -0,0 +1,122 @@
...
+      #
+      # feature - The Flipper::Feature for the gate.
+      # gate - The Flipper::Gate to disable.
+      # thing - The Flipper::Type being disabled for the gate.
+      #
+      # Returns true.
+      def disable(feature, gate, thing)
+        case gate.data_type
+        when :boolean
+          moneta[key(feature.key)] = default_config
```
jnunemaker: [Suggestion] I often use clear(feature) here to avoid duplicating it.

jnunemaker:
Feel free to merge whenever you go through the feedback and I'll cut a release after.

lib/flipper/adapters/moneta.rb
```diff
@@ -0,0 +1,122 @@
...
+
+      # Public: Gets the values for all gates for a given feature.
+      #
+      # Returns a Hash of Flipper::Gate#key => value.
+      def get(feature)
+        default_config.merge(moneta[key(feature.key)].to_h)
+      end
+
+      # Public
+      def get_multi(features)
```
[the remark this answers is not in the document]
AlexWheeler: *I so wish that moneta had bulk reads. Maybe we can try adding that someday upstream. *
yes that would be awesome and a fun feature to implement. Would also be sweet if  Moneta had instrumentation support like Flipper.  They have it for ActiveSupport::Cache, but not all adapters

You're totally right this can just use those methods

flipper-moneta.gemspec
```diff
@@ -0,0 +1,24 @@
...
+  gem.homepage      = 'https://github.com/jnunemaker/flipper'
+
+  gem.files         = `git ls-files`.split("\n").select(&flipper_moneta_files) + ['lib/flipper/version.rb']
+  gem.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n").select(&flipper_moneta_files)
+  gem.name          = 'flipper-moneta'
+  gem.require_paths = ['lib']
+  gem.version       = Flipper::VERSION
+
+  gem.add_dependency 'flipper', "~> #{Flipper::VERSION}"
+  gem.add_dependency 'moneta', '>= 0.7.19'
```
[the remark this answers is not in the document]
AlexWheeler: totally makes sense. This is passing against 0.7.0 - 1.0 so going to update the constraint to that (can't remember why I had issues w/ < 0.7.19, but is fine on the old branch too 🤷‍♂️ )

AlexWheeler:
@jnunemaker sweet, do you have any insight into the failing spec for `Flipper::Adapters::Sync::IntervalSynchronizer`. I see it failing on #360 for the `gemspec` metadata changes, but not on master so not sure why it would fail on those changes and why it might be failing on this branch too

flipper-moneta.gemspec
```diff
@@ -0,0 +1,24 @@
...
+  gem.homepage      = 'https://github.com/jnunemaker/flipper'
+
+  gem.files         = `git ls-files`.split("\n").select(&flipper_moneta_files) + ['lib/flipper/version.rb']
+  gem.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n").select(&flipper_moneta_files)
+  gem.name          = 'flipper-moneta'
+  gem.require_paths = ['lib']
+  gem.version       = Flipper::VERSION
+
+  gem.add_dependency 'flipper', "~> #{Flipper::VERSION}"
+  gem.add_dependency 'moneta', '>= 0.7.0', '<= 1.0.0'
```
jnunemaker: Looks good to me!

jnunemaker:
> @jnunemaker sweet, do you have any insight into the failing spec for Flipper::Adapters::Sync::IntervalSynchronizer. I see it failing on #360 for the gemspec metadata changes, but not on master so not sure why it would fail on those changes and why it might be failing on this branch too

I'm not sure why yet, but I think it is intermittently failing. I need to take a look but haven't had time.

PR #382, opened 2018-11-02
Make it easier to decorate flipper

Trevoke:
Hello!

This is obviously not complete: the change as proposed would have to be applied to the other files.

This came about because we were trying to decorate Flipper and log enabling and disabling of features -- so we wrote a simple delegator that wrapped Flipper, but that did not log changes made through the web interface. As you see below, we then went as far as creating a decorator for the internal "feature" object, but a user of Flipper shouldn't be decorating anything beyond the Flipper API.

```ruby
  module Flipper
    class Decorator < SimpleDelegator
      def initialize(adapter)
        @adapter = ::Flipper.new(adapter)
        super(@adapter)
      end

      # overridden to log when used in console
      def enable(name, *args)
        Rails.logger.info "ENABLE #{name} with #{args} from API"
        super(name, *args)
      end

      # overridden to log when used in console
      def disable(name, *args)
        Rails.logger.info "DISABLE #{name} with #{args} from API"
        super(name, *args)
      end

      def [](feature_name)
        feature = super(feature_name)
        Flipper::FeatureDecorator.new(feature)
      end
    end
  end
```

So, this is a suggestion for a change that would allow people to decorate Flipper more easily.

### DIFF
--- a/lib/flipper/ui/actions/boolean_gate.rb
+++ b/lib/flipper/ui/actions/boolean_gate.rb
@@ -10,13 +10,13 @@ module Flipper
         route %r{\A/features/(?<feature_name>.*)/boolean/?\Z}
 
         def post
-          feature = flipper[feature_name]
-          @feature = Decorators::Feature.new(feature)
+          feature = feature_name.to_sym
+          @feature = Decorators::Feature.new(flipper[feature])
 
           if params['action'] == 'Enable'
-            feature.enable
+            flipper.enable(feature)
           else
-            feature.disable
+            flipper.disable(feature)
           end
 
           redirect_to "/features/#{@feature.key}"


AlexWheeler:
@Trevoke have you looked at Flipper's [instrumentation capabilities](https://github.com/jnunemaker/flipper/blob/master/docs/Instrumentation.md)?  You can subscribe to any Flipper events without having to decorate anything.  

For example if you're using the ActiveRecord adapter and want to log any enables/disables of features you could add the following to your Flipper initializer (or wherever you initialize Flipper in your app if not using Rails)

```
class FlipperSubscriber

 # this will be called anytime any of your adapter's public instance methods are called
  def call(*args)
    event = ActiveSupport::Notifications::Event.new(*args)
    payload = event.payload

    case payload[:operation]
    when :enable
      Rails.logger.info "Enabled #{payload[:feature_name]}"
    when :disable
      Rails.logger.info "Disabled #{payload[:feature_name]}"
    end
    # you could check for more operations here, such as add, remove, etc.  Check the instrumented adapter for all the available operations
  end
  ActiveSupport::Notifications.subscribe(/flipper/, new)
end

active_record_adapter = Flipper::Adapters::ActiveRecord.new
instrumented = Flipper::Adapters::Instrumented.new(active_record_adapter, :instrumenter => ActiveSupport::Notifications)

Flipper.configure do |config|
  config.default do
    Flipper.new(instrumented, :instrumenter => ActiveSupport::Notifications)
  end
end
```

Trevoke:
Thanks for the prompt reply! Certainly, reading the documentation more carefully may have avoided a headache.

If I understand correctly based on your example, what happens is that we initialize an adapter of our choosing, then use one of the provided decorators, and then we pass the decorated adapter to Flipper itself -- although for both the decorator and flipper, we have to specify the instrumenter that we want to enable?

I can see now that it's not a violation of the open/closed principle, but it kind of feels like one at first glance. Other bits of documentation indicate that we're always just wrapping the adapter and that makes me wonder why the user ever initializes Flipper if it's not actually something that the user should feel free to modify?

AlexWheeler:
Looking at this again we don't need to use the instrumented adapter and can just specify the instrumenter when initializing Flipper

```
Flipper.configure do |config|
  config.default do
    Flipper.new(active_record_adapter, :instrumenter => ActiveSupport::Notifications)
  end
end
```

*that makes me wonder why the user ever initializes Flipper if it's not actually something that the user should feel free to modify?*
I'm not totally sure what you mean here, can you expand on this, thanks!

Trevoke:
@AlexWheeler Ah, that new bit of code does make a lot more sense.

Sure, sorry for not being very clear.

My perspective here is that whatever code is not provided by Flipper itself is code that the user should be able to change. Here, it seems like `Flipper` the object is the message receiver and that it forwards them to the instance of Flipper that was instantiated during the initialization. I interpret this to mean that if I want to wrap Flipper's behavior, I just have to modify the Flipper instance that gets passed in. Hopefully, this logic explains why I thought this pull request was a good idea.

Since it seems like that is not how Flipper is intended to be used (instead, it seems like what the user may want to modify is the adapter), it seems counterproductive to force every codebase in existence to have `Flipper.new` when that code could just be built-in.

The DSL could be extended like this, similar to the DSL that puma uses:

```ruby
Flipper.configure do
  adapter active_record_adapter
  instrumenter ActiveSupport::Notifications
end
```

What do you think?

AlexWheeler:
ah yeah so it sounds like you'd just like another layer of abstraction so when using flipper you don't need to know about the Flipper instance. Flipper used to not use a configure block and you'd just interact directly with a flipper instance.  Abstractions can be great, but imo I think this one would be unnecessary.  I'm going to close this issue since it sounds like we've resolved it, but if not just let me know and we'll open it back up!

Trevoke:
I'm actually not talking about another layer of abstraction at all! I'm
talking about removing a bit of code that should not be presented to the
user. I gave an example that included a DSL but it could just as easily be
a hash of options passed to the config object in the block. The actual
implementation is not as important as the idea of not letting the user have
the impression that Flipper is modifiable if it isn't.

--

Aldric.

AlexWheeler:
Nice yea I for sure think its an interesting idea. Check out [this PR](https://github.com/jnunemaker/flipper/pull/279) for more context on when the configure block was introduced.

*I'm talking about removing a bit of code that should not be presented to the user.*

`Flipper.configure` just gives you a way to configure the default Flipper instance per thread and delegates the public method calls to that instance. You can still initialize multiple Flipper instances if you want and use those instead, in fact, that's how Flipper was used for a long time

jnunemaker:
@AlexWheeler as far as I understand I agree with what you are saying. 

@Trevoke maybe some sort of example or documentation could have helped you? Any specific place you would have checked that we could maybe update?

Trevoke:
From my paradigm, the `configure` block is a neat idea that also lies to the user, but it seems like you think that it does not lie to the user. Is this a fair representation of where we are right now?

jnunemaker:
@Trevoke Thanks for the chat here. We really enjoy thinking through use cases and talking with real flipper users. I think it would be helpful to circle back to what prompted this PR for you. 

**tldr**: You wanted to log things. For this type of interaction, we recommend instrumentation. We even provide a [log subscriber for you](https://github.com/jnunemaker/flipper/blob/a04de088bed7f28209f5586a438cf78b975594aa/lib/flipper/instrumentation/log_subscriber.rb). Adapters are definitely the place where decorating takes place (verse `Flipper`). 

## Background

* The core of flipper is `Flipper::Feature`. 
* To avoid needing to initialize those all the time, we provide `Flipper::DSL` (akin to a client of sorts). 
* `Flipper.new` (for better or worse) is a short way of initializing an instance of `Flipper::DSL`. 
* `Flipper.configure` serves a different purpose. Nearly everyone who was using flipper needed somewhere to store the dsl instance they were using. Everyone was doing it differently. Most only have one instance. `configure` is a ruby culture/common way of doing this type of interaction so that is what I went with. 
* The `default` block makes it easy to have an instance generator that allows `Flipper` to do some of the work to help you stay thread safe by keeping an instance per thread.

## Conclusion 

If you just need to log things, that is what instrumentation is for. I haven't hit a point where I needed to decorate `Flipper` itself and I don't see how the DSL you provided makes it easier to decorate `Flipper`. My personal opinion (could be entirely faulty) is that I dislike DSLs, especially `instrumenter`  (where receiver is hidden) vs `config.instrumenter` (where receiver is not hidden).

## Where we are at 

I think where we are at is I'm sorry you lost some time and got confused. That sucks. If you just want to log, do what I mentioned above. If you come up with a new use where you'd like to decorate Flipper, let us know. We can provide the Flipper way for you and if there isn't one we talk semantics about whether or not you've been lied to. Hope that helps and have a great day!

Also, while I have you here, I would love to know what you are using flipper for as well and how it is working out for you (what went well, what was hard other than what you mentioned here). If you aren't comfortable leaving that information here in a public comment, feel free to email me directly john@featureflipper.com. I'm just always curious. :)

PR #384, opened 2018-11-20
Remove YouTube embed from default landing

cbartlett:
I find the default landing page frustratingly unprofessional. You can call me unfunny if you want, but I always need to ensure I add some features in every environment, lest I get strange looks and comments from my clients.

Am I the only one who thinks we should just keep this page simple and meme-free?

screen shot 2018-11-20 at 7 26 51 am

### DIFF
--- a/lib/flipper/ui/views/features.erb
+++ b/lib/flipper/ui/views/features.erb
@@ -3,11 +3,8 @@
     <span class="mega-octicon octicon-plus"></span>
     <span class="mega-octicon octicon-list-unordered"></span>
     <span class="mega-octicon octicon-zap"></span>
-    <h4>But I've got a blank space baby...</h4>
-    <p>And I'll flip your features.</p>
-    <div class="embed-responsive embed-responsive-16by9">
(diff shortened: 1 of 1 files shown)


jnunemaker:
I can see that. Could I talk you into adding a config option for this to Flipper::UI config?  Perhaps `config.taylor_swift = true/false` and default it to `true`?

jnunemaker:
Or config.business_time, or config.professional, etc. :)

jcraigk:
I support this PR.  This is getting to be a mature gem used by a lot of folks, and I feel it has probably outgrown the whimsical nature of the YouTube embed.  I appreciate the fun intent, but I think the OP makes a valid overriding point.

jnunemaker:
I have a history of maintaining whimsicality long into maturity with gems, so I'll be leaving this. Happy to have a config option. Closing this to keep things tidy, but if anyone wants to open a new one with config option or update this one, I'd be happy to merge.

jcraigk:
Ah, I didn't realize you maintain HTTParty as well.  That rather unusual (inappropriate?) use of the post-install message "you must party hard" (https://github.com/jnunemaker/httparty/blob/master/httparty.gemspec#L22) makes more sense  to me now.  Your gems, your call :)

PR #385, opened 2018-11-21
Allow to customize Rack::Protection middleware list

knapo:
This PR allows to customize `Rack::Protection` middleware list.

Currently we are forced to use `Rack::Protection` defaults, when we may want to use or skip some of them: https://github.com/sinatra/sinatra/blob/master/rack-protection/lib/rack/protection.rb#L23-L51.

### DIFF
--- a/lib/flipper/ui.rb
+++ b/lib/flipper/ui.rb
@@ -41,15 +41,15 @@ module Flipper
 
     def self.app(flipper = nil, options = {})
       env_key = options.fetch(:env_key, 'flipper')
+      rack_protection_options = options.fetch(:rack_protection, use: :authenticity_token)
       app = ->() { [200, { 'Content-Type' => 'text/html' }, ['']] }
       builder = Rack::Builder.new
       yield builder if block_given?
-      builder.use Rack::Protection
-      builder.use Rack::Protection::AuthenticityToken
+      builder.use Rack::Protection, rack_protection_options
       builder.use Rack::MethodOverride
       builder.use Flipper::Middleware::SetupEnv, flipper, env_key: env_key
       builder.use Flipper::Middleware::Memoizer, env_key: env_key
-      builder.use Middleware, env_key: env_key
+      builder.use Flipper::UI::Middleware, env_key: env_key
       builder.run app
       klass = self
       builder.define_singleton_method(:inspect) { klass.inspect } # pretty rake routes output


lib/flipper/ui.rb
```diff
@@ -41,15 +41,15 @@ def self.root
...
       builder = Rack::Builder.new
       yield builder if block_given?
-      builder.use Rack::Protection
-      builder.use Rack::Protection::AuthenticityToken
+      builder.use Rack::Protection, rack_protection_options
       builder.use Rack::MethodOverride
       builder.use Flipper::Middleware::SetupEnv, flipper, env_key: env_key
       builder.use Flipper::Middleware::Memoizer, env_key: env_key
-      builder.use Middleware, env_key: env_key
+      builder.use Flipper::UI::Middleware, env_key: env_key
```
jnunemaker: 🙇

lib/flipper/ui.rb
```diff
@@ -41,15 +41,15 @@ def self.root
 
     def self.app(flipper = nil, options = {})
       env_key = options.fetch(:env_key, 'flipper')
+      rack_protection_options = options.fetch(:rack_protection, { use: :authenticity_token })
```
jnunemaker: I noticed you are changing the default options here from nothing to only using authenticity token. Why is that?

knapo: No, I kept it as it was - some of rack protection middlewares, are disabled by default, unless explicitly enabled (e.g. `authenticity_token`), and some are enable by default, unless skipped: https://github.com/sinatra/sinatra/blob/master/rack-protection/lib/rack/protection.rb#L23-L51

So:
```ruby
builder.use Rack::Protection, { use: :authenticity_token }
```

has the same effect as:
```ruby
builder.use Rack::Protection
builder.use Rack::Protection::AuthenticityToken
```

knapo:
> Hi there! Thanks for the pull request. What do you specifically need to customize and what is your reason for doing so?

I want to be able to disable specific rack-protection middlewares which are enabled by default(because of stack architecture reasons they don't work or doesn't bring any value) and enabled other which are disabled by default . Currently flipper ui forces to use rack-protection defaults, when rack-protection allows to customize it by itself.

shraykay:
what's the status of this PR?

kaimou1357:
Yeah, we are running into issues with this middleware. Is this going to be merged soon?

knapo:
Finally 🎉

Issue #394, opened 2019-01-21
Flipper gates cannot be audited 


abelvarga:
Hi!

We are using Audited gem to audit all changes deemed important. We would like to audit flipper gate changes. Create works OK, but the active record adapter uses `delete_all` which bypasses all callbacks. This makes actor removals not auditable.

Could we just use `destroy_all` instead or is there a known drawback?

Happy to open PR if there's no objection!

Cheers,
Abel

jnunemaker:
The primary reason was to reduce network calls (1 in delete_all vs many in destroy_all), but the uses of delete_all are all for writes so that probably isn't a huge concern. I do my best to keep reads (the hottest path in any app) to be the minimum network calls and sometimes it carries over into writes. I'd be open to a PR that changes this for sure.

abelvarga:
@jnunemaker, just opened the PR. Please advise on if I need to add specs or not. I'm guessing the latter.

Also, some CI tests fail on bundler issues. Should I be concerned? Or is that to be expected?

jnunemaker:
@abelvarga I wouldn't be concerned. Likely a travis environment thing.

abelvarga:
@jnunemaker, thanks for the merge! When can we expect a release?

abelvarga:
@jnunemaker, is this a breaking change?

jnunemaker:
Releasing 0.16.1 now!

PR #395, opened 2019-01-22
Replace delete_all with destroy_all in ActiveRecord adapter

abelvarga:
#394


abelvarga:
@jnunemaker, dunno how I missed that one! Thanks, will update now.

PR #398, opened 2019-01-25
remove Taylor Swift joke

nachokb:
Empty features page features an embedded Taylor Swift reference as a joke.

Don't mean to bother, but it grew old fast.


nachokb:
hadn't seen #384

PR #399, opened 2019-01-29
Remove version constraint for Sequel gem

danilobarion1986:
I want to use this gem, but we are heavy-users of Sequel, and limiting the version to < 5 will block us...
I removed it, and apparently nothing (rspec and rubocop) broke.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,6 +10,7 @@ AllCops:
     - 'examples/**/*'
     - 'tmp/**/*'
     - 'bin/**/*'
+    - 'vendor/bundle/**/*'
   TargetRubyVersion: 2.0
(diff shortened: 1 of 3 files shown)


danilobarion1986:
@jnunemaker , I tried to edit .travis.yml file, passing an specific version of bundler, but I won't was able to fix the build error yet :cry: ... It doens't seem to be an error due to my changes, so if you could help me, I'll appreciate it.

danilobarion1986:
> @jnunemaker , I tried to edit .travis.yml file, passing an specific version of bundler, but I won't was able to fix the build error yet cry ... It doens't seem to be an error due to my changes, so if you could help me, I'll appreciate it.

Well, I think that it's everything ok now! :smiley:

Issue #392, opened 2019-01-16
Bump flipper-redis redis constraint


AlexWheeler:
[Redis 4.1.0](https://rubygems.org/gems/redis/versions/4.1.0) was released December 13 and flipper-redis requires redis < 4.0, which lead to bundler resolving to a super old version of flipper-redis without a redis constraint in #391 .

PR #401, opened 2019-02-17
Update Redis dependency constraint to < 5

zokioki:
Redis gem v4.1 was recently released, so any app running it will
cause bundler to resolve to using an older version of `flipper-redis`.
We should consider relaxing the constraint to anything less than v5.

Addresses https://github.com/jnunemaker/flipper/issues/392


jnunemaker:
Hey thanks! I'll get this merged and release it soonish.

Issue #405, opened 2019-04-12
Symbol procs as groups don't work as expected.


supersam654:
Consider these two groups:

```
Flipper.register(:working) do |actor|
  actor.admin?
end

Flipper.register(:broken, &:admin?)

class User
  def admin?
    true
  end
end

```

The first group works as expected. However, the second group raises an exception about passing the wrong number of arguments to `admin?` even though it appears to be syntactic sugar for the first one.

I did some digging and the problem appears to be at https://github.com/jnunemaker/flipper/blob/master/lib/flipper/types/group.rb#L17.

Notably, the arity of the symbol proc is `-1` while the arity of the regular block is `1`.

```
def print_arity(&block)
  puts block.arity
end

print_arity(&:to_s) # Prints -1
print_arity { |a| a.to_s } # Prints 1
```

With all of that said, I'm not very familiar with [arity](https://apidock.com/ruby/Proc/arity) checking in Ruby. However, it seems that checking the absolute value of the arity would work in this situation. If someone else thinks this is a good idea, I'd be more than happy to make the (tiny) PR for this.

jnunemaker:
Definitely open to a PR with a spec/fix.

thetimbanks:
Closing since this was resolved by https://github.com/jnunemaker/flipper/pull/406

PR #406, opened 2019-04-20
Allow shorthand block notation on group types

thetimbanks:
@jnunemaker @supersam654 I went ahead and took a shot at fixing the the issue in #405. See if this makes sense.

## Description of changes
When a shorthand block was used when initializing a Group, an error would be thrown about not passing the correct number of arguments. That was due to the arity of the shortand being -1 instead of 1. By taking the absolute value of the arity, an arity of -1 will still correctly set the single_argument instance variable.

### DIFF
--- a/lib/flipper/types/group.rb
+++ b/lib/flipper/types/group.rb
@@ -14,7 +14,7 @@ module Flipper
 
         if block_given?
           @block = block
-          @single_argument = @block.arity == 1
+          @single_argument = @block.arity.abs == 1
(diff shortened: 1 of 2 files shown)


jnunemaker:
I feel like there is some edge case here but I can’t come up with it quickly so I’m inclined to say let’s roll with it and find it later if it does exist.

jnunemaker:
@thetimbanks thanks for the tests and the PR!

PR #407, opened 2019-04-24
Replace Erubis with Erubi (2)

pcreux:
Following up on https://github.com/jnunemaker/flipper/pull/397.

- [x] Merge master
- [x] Tell rubocop using `eval` is ok here
- [x] Configure Erubi to escape `<%=`

This should make `flipper-ui` Rails 5.2 compatible.

### DIFF
--- a/lib/flipper/ui/action.rb
+++ b/lib/flipper/ui/action.rb
@@ -1,6 +1,6 @@
 require 'forwardable'
 require 'flipper/ui/error'
-require 'flipper/ui/eruby'
+require 'erubi'
 require 'json'
 
 module Flipper
@@ -208,9 +208,9 @@ module Flipper
 
         raise "Template does not exist: #{path}" unless path.exist?
 
-        contents = path.read
-        compiled = Eruby.new(contents)
-        compiled.result proc {}.binding
+        # rubocop:disable Lint/Eval
+        eval(Erubi::Engine.new(path.read, escape: true).src)
+        # rubocop:enable Lint/Eval
       end
 
       # Internal: The path the app is mounted at.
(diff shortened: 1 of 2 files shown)


lib/flipper/ui/action.rb
```diff
@@ -208,9 +208,7 @@ def view(name)
 
         raise "Template does not exist: #{path}" unless path.exist?
 
-        contents = path.read
-        compiled = Eruby.new(contents)
-        compiled.result proc {}.binding
+        eval(Erubi::Engine.new(path.read).src)
```
pcreux: Rubocop says: 
```
Offenses:
lib/flipper/ui/action.rb:211:9: W: Lint/Eval: The use of eval is a serious security risk.
        eval(Erubi::Engine.new(path.read).src)
        ^^^^
219 files inspected, 1 offense detected
The command "bundle exec rubocop -D && bundle exec rake" exited with 1.
```

jnunemaker: Yes. Ok.

pcreux:
@jnunemaker It seems that we had it configured to [flip that](https://github.com/jnunemaker/flipper/pull/407/files#diff-6c802ecba72159995fc160ce4cfd5ac3L7) with erubis. It might be better to go with the default behavior?

jnunemaker:
I switched it intentionally. The reason is because most people use `<%=` and I want that to be escaped by default. I'd prefer it to be intentional to not escape something. Does that make sense? What do you think?

pcreux:
Sorry for the late reply @jnunemaker. This makes total sense. I configured erubi to escape `<%=`.

jnunemaker:
A couple things i want to do on this:

* Remove the flipper/ui/eruby file as now unused
* Tighten up the constraint to > 1 and < 2

Going to merge. Thanks for the work. I'll knock out the above items.

jnunemaker:
63e8dc087499a0ee6077388c33bfa3bd1f126cfb modifies gemspec.

pcreux:
Thank you @jnunemaker !

cforcey:
Many thanks -- excited to see this hit an official release as we are seeing the same security vulnerability on our security monitors!

PR #411, opened 2019-05-13
[UI Configuration] Allow for feature icon to be configured to another octicon

the-robear:
This PR adds a configuration option in `Flipper::UI::Configuration` called `feature_icon` to allow for use of any of the other [ocitcons](https://octicons.github.com/), and defaults it to the good ol' classic squirrel.

### DIFF
--- a/lib/flipper/ui/configuration.rb
+++ b/lib/flipper/ui/configuration.rb
@@ -10,7 +10,8 @@ module Flipper
                   :percentage_of_time
 
       attr_accessor :banner_text,
-                    :banner_class
+                    :banner_class,
+                    :feature_icon
 
       # Public: If you set this, the UI will always have a first breadcrumb that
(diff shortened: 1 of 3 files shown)


jnunemaker:
@the-robear thanks for taking the time. Could you explain why you are making this change? Do you use a different octicon? If so, what is it?

the-robear:
Yep! I will be using the lightning bolt, [called `zap` in the octicon set](https://octicons.github.com/icon/zap/). Also considering using the lightbulb.

jnunemaker:
Ok. Why change it? I know it seems small, but I will need to support it going forward and I'm struggling a bit with why it is even worth changing it or what sparked the desire. Are you migrating from something else where you used a zap or light bulb or just want to tinker and change it?

jnunemaker:
Closing to keep things clean here, but I'd still love to know the why and may still re-open/merge. Thanks!

PR #417, opened 2019-05-27
Include thing_value in payload of Instrumented#enable and #disable

rafaelmillan:
When subscribing to notifications it can be useful to know the value of the gate being enabled/disabled. For example, that way you could set up notifications and output the name of the group or the percentage being changed.

This adds a new `thing_value` attribute to the payload of `Instrumented#enable` and `Instrumented#disable`.

Thanks for the great gem!

### DIFF
--- a/lib/flipper/adapters/instrumented.rb
+++ b/lib/flipper/adapters/instrumented.rb
@@ -124,6 +124,7 @@ module Flipper
           adapter_name: @adapter.name,
           feature_name: feature.name,
           gate_name: gate.name,
+          thing_value: thing.value,
         }
 
         @instrumenter.instrument(InstrumentationName, payload) do |payload|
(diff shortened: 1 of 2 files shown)


jnunemaker:
For some reason I am on the fence about this, but I can't come up with a strong reason why not to and the PR is simple and tested so I'm going to merge. Thanks! 

I think part of it might be that one could just do Flipper[feature_name].gate_values[gate_name] to get the value and it should be memoized (to avoid the extra read lookup).

rafaelmillan:
Thank you for the approval 🙌 

> I think part of it might be that one could just do Flipper[feature_name].gate_values[gate_name] to get the value and it should be memoized (to avoid the extra read lookup).

I hadn't think of that. I could be missing something but I think using your suggestion it still wouldn't be possible to find out which value was disabled for gates that use sets (groups and actors).

jnunemaker:
> I could be missing something but I think using your suggestion it still wouldn't be possible to find out which value was disabled for gates that use sets (groups and actors).

Great point.

PR #418, opened 2019-06-17
Feature timestamps

mtodzo:
I have added a timestamp attribute to Feature.rb. The idea is to have timestamps that can be evaluated in the UI so that it will be obvious which features have gone stale (should be permanently added or removed from the codebase). A common issue with the feature toggle design pattern is that the codebase becomes riddled with flags and they grow difficult to track, so this aims to mitigate that pain.

Current Status:
- Timestamps are currently only compatible with the Redis adapter ( others or I could add PR's for other adapters )
- Timestamps are updated on each change to  a gate (not upon polling enabled?/disabled?)
- Timestamps are displayed in the UI in a third column if the adapter being used is a Redis adapter. I am hoping add toggles for showing timestamps and sorting by timestamp but I am having some trouble since I don't have much frontend experience.

### DIFF
--- a/lib/flipper/ui/views/features.erb
+++ b/lib/flipper/ui/views/features.erb
@@ -28,10 +28,13 @@
           </th>
           <th class="col">Feature</th>
           <th class="col">Enabled Gates</th>
+          <% if flipper.adapter.adapter.name.eql? :redis %>
+            <th class="col">Gate Timestamp</th>
+          <% end %>
         </tr>
       </thead>
       <tbody>
-        <% if true %>
+        <% if false %>
           <% @features.sort { |a,b| a.time_compare(b) }.each do |feature| %>
             <tr class="d-flex">
               <td class="col-1">
@@ -43,6 +46,9 @@
               </td>
               <td class="col">
                 <%= feature.pretty_enabled_gate_names %>
+              </td>
+              <td class="col">
+                <%= feature.pretty_timestamp %>
               </td>
             </tr>
             <% end %>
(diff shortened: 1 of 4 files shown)


lib/flipper/ui/views/features.erb
```diff
@@ -28,22 +28,50 @@
...
-            </td>
-            <td class="col">
-              <a href="<%= "#{script_name}/features/#{feature.key}" %>">
-                <%= feature.key %></a>
-            </td>
-            <td class="col">
-              <%= feature.pretty_enabled_gate_names %>
-            </td>
-          </tr>
+        <% if false %>
```
mtodzo: This is where I was hoping to have the toggle to turn Time Sorting on/off. I wasn't exactly sure how to go about doing so within an erb file. Would love some suggestions!

jnunemaker:
Hey! Thanks for starting/taking a stab at this. All data definitely needs to be managed at the adapter layer. The downside is that means things like returning sorted data would also need to happen there. Not every data storage system loves this type of sorting. I've always found the line difficult to walk around working with the lowest common denominator and supporting great ideas/features like yours here. 

Another problem is keeping things like this performant across all adapters which is a requirement for me. 

Last "problem" is I've been very seriously considering switching to lower level for the adapters for a variety of reasons and sorting would complicate that. 

All that said, there are creative solutions to work around this. What I and most others who have done this did is to build their own adapter and UI, which really doesn't take that long (list features, show feature, update gate gets the job done). For example if you use the redis adapter, copy/paste the code for it into your project and name it whatever you like (MyCustomAdapter, MyRedisAdapter, TimestampedRedisAdapter, etc.). Tweak it to include the extra data you need. Then, in the UI you can show this data by hitting the data layer directly and avoiding 3rd party flipper/adapter code. 

We did this at GitHub and I've done this on other projects, most notably for https://www.flippercloud.io which I've been working on (though not the past few months, hoping to start up again on it this fall). 

I'm going to close this but please feel free to keep the conversation/questions flowing. I'm here to help, I'm just not ready to put this in flipper core yet.

PR #423, opened 2019-07-02
added(flipper-ui): ability to search features by name

hhuynhlam:
### Changes
- add the ability to search features by name 
Screen Shot 2019-07-02 at 8 21 20 AM
Screen Shot 2019-07-02 at 8 21 31 AM

### DIFF
--- a/lib/flipper/ui/actions/features.rb
+++ b/lib/flipper/ui/actions/features.rb
@@ -10,9 +10,15 @@ module Flipper
 
         def get
           @page_title = 'Features'
+          @search = params['search']
+
           @features = flipper.features.map do |feature|
             Decorators::Feature.new(feature)
           end.sort
+
+          if @search
+            @features = @features.select { |feature| feature.name.match(@search) }
+          end
 
           @show_blank_slate = @features.empty?
 
--- a/lib/flipper/ui/views/features.erb
+++ b/lib/flipper/ui/views/features.erb
@@ -1,25 +1,38 @@
-<% if @show_blank_slate %>
-  <div class="jumbotron text-center">
-    <span class="mega-octicon octicon-plus"></span>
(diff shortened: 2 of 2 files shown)


jnunemaker:
@hhuynhlam hi there! Thanks for taking the time to make a pull request. The work looks good. That said, I'd like to take a step back. The primary goals for search/filter are to 1) move less data over the wire or 2) find the needle in the haystack faster. 

1. Because the adapters have no search functionality built in, this will pull all data over the wire and filter in ruby, which doesn't help reduce what is going over the wire so we don't get any benefit there. It will save the generation of html on the server for features that are filtered out and reduce the html over the wire from server to client but I'm not sure that alone convinces me.
2. All features show on the list page. That may result in a daunting list, but I'm not sure that adding search to the UI and doing an extra request is faster than using every browser's built in find on page functionality. Have you tried that? 

If my primary goal is wrong or I'm missing something let me know. If we had smarter adapters with built in pagination or search, I'd be way more inclined to merge this, but since we are dealing with a single page, I'm inclined to suggest people use find on page in their browser of choice.

Out of curiosity, how many features do you have or what specific inspired this PR? That might help me better understand your use case. Thanks!

hhuynhlam:
Hey @jnunemaker, 

I appreciate the feedback! The motivation for this PR might be very specific to our use case. Currently, we have several clients that interact with the same API. The API manages feature flags for all of these clients, which makes it difficult to view and manage feature flags by client. We've standardize our feature flagging naming convention to relevant context (client, platform, feature, scope, etc), so the ability to search effectively gives us the ability to create views based on the context.

jnunemaker:
@hhuynhlam gotcha. Neat use. Thanks for letting me know. Sounds like something flippercloud.io might be able to help with someday if I can get around to fully shipping it.

PR #424, opened 2019-07-18
Query single gate if `Actor` is provided

khalilovcmd:
Just want to say thanks for the great work done on this gem 🙌 

> The individual actor gate is typically not designed for hundreds or thousands of actors to be enabled. 

Quoting the [Caveats](https://github.com/jnunemaker/flipper/blob/master/docs/Caveats.md#caveats) section, I am wondering if this design decision can be tweaked.

I can think of plenty of use cases requiring Actors to be in the hundreds or thousands, i.e.: rolling out a feature incrementally to waves of customers, where each wave is about 100 customers. Feature Flags is an ideal candidate for this mode of release, it is safer, and easier to manage than most methods.

Assuming the ActiveRecord adapter is used, if the Actor based FF check was at a performance critical path in the application, we will need to always: 

```sql
SELECT `flipper_gates`.* FROM `flipper_gates` WHERE `flipper_gates`.`feature_key` = 'my-cool-feature'
```

Which is not ideal, and it can slow things down significantly, specifically if it was in a pipeline-based code path where hundreds or thousands of jobs being processed.

If an Actor object is passed, I propose either:

1. Look up for a "[get_by_value](https://github.com/jnunemaker/flipper/compare/master...khalilovcmd:query-single-gate-if-actor-is-provided?expand=1#diff-076f15bc0496c95e3b3657b153002fedR111)" method support in an adapter to do a query fetch by ID

In the ActiveRecord adapter, it would look like this:

```sql
SELECT `flipper_gates`.* FROM `flipper_gates` WHERE `flipper_gates`.`feature_key` = 'my-cool-feature' AND `flipper_gates`.`value` = '3'
```

2. Or introduce an entirely new method, let's say, `enabled_single?(....)` (bad naming, sorry!) that does the same operation as above.

On another note, the newly introduced method would probably have an equivalent in other adapters too, maybe it is worth the effort.

-----

⚠️  @jnunemaker  My implementation attempt here is definitely not ideal, but I would love to know what you think of the proposal in general 🙌

### DIFF
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -86,6 +86,11 @@ module Flipper
         result_for_feature(feature, db_gates)
       end
 
+      def get_by_value(feature, value)
+        db_gates = @gate_class.where(feature_key: feature.key, value: value)
+        result_for_feature(feature, db_gates)
+      end
+
       def get_multi(features)
         db_gates = @gate_class.where(feature_key: features.map(&:key))
         grouped_db_gates = db_gates.group_by(&:feature_key)
--- a/lib/flipper/feature.rb
+++ b/lib/flipper/feature.rb
@@ -102,13 +102,25 @@ module Flipper
     # Returns true if enabled, false if not.
     def enabled?(thing = nil)
       instrument(:enabled?) do |payload|
-        values = gate_values
-        thing = gate(:actor).wrap(thing) unless thing.nil?
-        payload[:thing] = thing
+        values = []
+
+        if thing.nil?
+          values = gate_values
+        else
+          if adapter.respond_to?(:get_by_value)
+            values = gate_values_by_id(thing.is_a?(Integer) ? thing : thing.id)
+          else
+            values = gate_values
+          end
+
+          thing = gate(:actor).wrap(thing)
+          payload[:thing] = thing
+        end
+
         context = FeatureCheckContext.new(
-          feature_name: @name,
-          values: values,
-          thing: thing
+            feature_name: @name,
+            values: values,
+            thing: thing
         )
 
         if open_gate = gates.detect { |gate| gate.open?(context) }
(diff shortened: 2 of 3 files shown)


jnunemaker:
I **really** appreciate you caring enough to take a stab at this. I'm going to be straightforward and say that it is going to be at least a week or two before I can give this the in depth review a change like this deserves, but the tab is still open on my computer and I will get to it.

khalilovcmd:
Thanks @jnunemaker ! And no worries, would be happy to read your thoughts.

jnunemaker:
> I'm going to be straightforward and say that it is going to be at least a week or two

🤦 So sorry about this. I think I hadn't fully processed it and wasn't ready to think about all the ways it would affect flipper and I just overlooked it because of that. I'm going to try to push through and just respond from my gut. You shouldn't have to wait a year to hear from me.

> Quoting the Caveats section, I am wondering if this design decision can be tweaked.

It can, but it will ripple through all of flipper. Many things are designed for actors to be a small limited set. Designing flipper this way keeps it super lightweight to query all config for a feature, cache it, preload it, etc. If we remove this concept, then those all become a lot more difficult. I'm not opposed, but it is much more than just adding an adapter value. Preloading would need to be able to preload parts of features but not all or we just say preloading doesn't work at all if you are doing single actor checks. 

If what you need were not possible with flipper currently, I would be more likely to entertain the idea or try to build it, but since you can add a group that checks an attribute/method on an actor and achieve the same thing, I'm inclined to avoid changing my initial choice to keep group and actor set sizes small.

> I can think of plenty of use cases requiring Actors to be in the hundreds or thousands, i.e.: rolling out a feature incrementally to waves of customers, where each wave is about 100 customers. Feature Flags is an ideal candidate for this mode of release, it is safer, and easier to manage than most methods.

When you need to release to larger groups of users, we recommend % based rollouts. 

If you can't do % based, groups can be used akin to individual actor rollouts. In postgres, I add a jsonb `settings` or `data` column for extra information. If I wanted to rollout to many actors but not in a percentage based way or allow them to opt in, I'd do that. 

For example, we (boxoutsports.com) recently used flipper to launch a redesign and allowed people to opt in the first week or two. We added a redesign_opt_in key in a jsonb column on users table to do this. When they opted in, it added the `redesign_opt_in` key set to `true` in the jsonb column. Then, we had a flipper group named `redesign_opt_in` which simple checked if that key was there and set to true. We enabled the group for the `redesign` feature and people could opt in without overloading our actor set size. 

The only downside of this is you need an extra column or key/value storage somewhere to flip the bit. If we allowed any number of actors you wouldn't need that extra bit, but you'd lose all the current performance and simplicity that is gained by keep the config small enough to store in a single record or key or whatever the data storage needs. 

Hope that helps. Happy to talk more so let me know if you have any feedback or thoughts. Again, I'm totally up for someone trying this and taking it all the way and I will help design it, but using groups solves the problem for me so I will probably stay focused on building other parts of flipper.

PR #426, opened 2019-07-22
Collapse SQL statement to 1 line

listrophy:
Makes logging (e.g., tagging & filtering) much easier

### DIFF
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -97,7 +97,7 @@ module Flipper
       end
 
       def get_all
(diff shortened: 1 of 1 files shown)


jnunemaker:
Any aversion to just single lining the statement in the code and adding some rubocop ignore stuff to the long line? That would avoid needing to do replacement in production for something that really never changes. I can see the single line thing being useful enough to just let the code not look as nice.

listrophy:
Doesn't really matter to me. Since it doesn't change after initialization, you could store it in a new ivar inside `initialize`. Otherwise, using rubocop:disable on a long line doesn't bother me either.

All I really care about is getting the logged statement to be one line. 😀

PR #439, opened 2019-09-18
Add documentation about how to disable fun mode

vbalazs:
Documentation for configuration option introduced in #413


jnunemaker:
Nice smile in your avatar.

PR #448, opened 2019-11-25
Describe actor in Flipper.register

stevendaniels:
The documentation doesn't make it clear that the `actor`'s class when checking a group isn't the same as the the class of the actor used in the arguments. 

This change is a small documentation to help clear up that misconception.


jnunemaker:
Thanks!

Issue #450, opened 2019-12-03
Deleting features in a Redis Cluster raises Redis::Cluster::CrossSlotPipeliningError


phstc:
Hi

I switched Flipper to a Redis cluster and everything seems to be working well, except for deleting features.

When I try to delete a feature I get `Redis::Cluster::CrossSlotPipeliningError` 👇 

```
Redis::Cluster::CrossSlotPipeliningError: Cluster client couldn't send pipelining to single node. The commands include cross slot keys. ["flipper_features", "pablo_test"]
```

Apparently, the problem is with cluster + multi, used in here https://github.com/jnunemaker/flipper/blob/a3a318dd65fcdfbc60a43f3677c6e0d85bfc40e5/lib/flipper/adapters/redis.rb#L36-L42.

jnunemaker:
That multi is probably not overly important. I added it only to reduce network calls for remove from two to one. The upside of removing it is cluster works. The downside is an extra network call per feature remove, which shouldn't be a huge deal outside of big imports/syncs. I would be open to a PR removing the multi.

phstc:
Thank you @jnunemaker. That's a great explanation and I agree that it seems benefits pay off the downsides. 

I created a PR for removing `multi` #451.

PR #451, opened 2019-12-04
Remove multi on the Redis adapter for removing a feature

phstc:
multi fails on a cluster set.

Fix #451

See https://github.com/jnunemaker/flipper/issues/450#issuecomment-561702603


phstc:
Hi @jnunemaker - thanks for merging this! 

Do you have an ETA for the next release including this change?

PR #453, opened 2019-12-18
📝 Update docs for mounting Flipper::UI

TristanToye:
You need to target `Flipper` here as in the example no instance has been assigned to the `flipper` variable.

Just a small change that I ran into when referencing the examples here and trying to load a file.


jnunemaker:
Thanks!

PR #459, opened 2020-04-01
Updated active_record/README.md with Rails initialization guidance

stuartmccoll:
This pull request adds some additional context to the `active_record/README.md` for those looking to follow the general conventions on initialising Flipper. This is something that caught my team out in our continuous integration pipeline, so hopefully this is helpful to others that might be following the same pattern.

### DIFF
--- a/docs/active_record/README.md
+++ b/docs/active_record/README.md
@@ -37,43 +37,7 @@ flipper = Flipper.new(adapter)
 # profit...
 ```
 
-## Rails Initialization
-
-The current convention in the Ruby-on-Rails community is to instantiate a Flipper object in a Rails initializer, like so:
-
-```ruby
-# config/initializers/flipper.rb
-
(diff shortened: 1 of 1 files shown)


jnunemaker:
Hi thanks for caring and taking the time to update. 

This only happens if you try to use Flipper prior to the migrations being run. I use flipper exactly as configured here without problems in CI. I’m not sure this is the correct thing to document. Perhaps it should document that using Flipper add, enable or disable would cause the problem you encountered?

stuartmccoll:
Hi, that makes sense and happy to change the pull request to reflect. I'll get a change pushed up either today or over the weekend 👍.

juanroldan1989:
@stuartmccoll I've setup a PR (https://github.com/jnunemaker/flipper/pull/462) that fixes failing tests, including those on this PR : )

PR #461, opened 2020-05-14
Add `descriptions` support for `features`

juanroldan1989:
1. Simple way to provide insights/details about a specific feature within `Flipper::UI`.

2. Avoids generating `feature` keys too long when developers/admins/users need more **context** about a feature's behaviour:
- What is it meant to do ?
- What is it not meant to do ?
- Which sections of the platform it applies to ?
- How it differentiates itself from similar features ?
- What user roles does it apply to ?
- Etc.

### DIFF
--- a/docs/ui/README.md
+++ b/docs/ui/README.md
@@ -148,6 +148,17 @@ Flipper::UI.configure do |config|
 end
 ```
 
+We can also associate a `description` for each `feature` by providing a descriptions source:
+
+```ruby
+Flipper::UI.configure do |config|
+  config.descriptions_source = ->(keys) do
+    # descriptions loaded from YAML file or database (postgres, mysql, etc)
+    # return has to be hash of {String key => String description}
+  end
+end
+```
+
 results in:
 
 ![configure](images/configured-ui.png)
--- a/lib/flipper/ui/action.rb
+++ b/lib/flipper/ui/action.rb
@@ -1,4 +1,5 @@
 require 'forwardable'
+require 'flipper/ui/configuration'
 require 'flipper/ui/error'
 require 'erubi'
 require 'json'
@@ -63,6 +64,14 @@ module Flipper
       def self.public_path
         @public_path ||= Flipper::UI.root.join('public')
       end
+
+      ##### RFC: What if we place the `descriptions` call here instead ?
+      #####      so we avoid calling it from both `feature` and `features` actions
+
+      # Private: The path to descriptions source
+      # def self.descriptions
+      #   Flipper::UI.configuration.descriptions_source
+      # end
 
       # Public: The instance of the Flipper::DSL the middleware was
       # initialized with.
--- a/lib/flipper/ui/actions/feature.rb
+++ b/lib/flipper/ui/actions/feature.rb
@@ -11,6 +11,8 @@ module Flipper
 
         def get
           @feature = Decorators::Feature.new(flipper[feature_name])
+          descriptions = Flipper::UI.configuration.descriptions_source
+          @feature.description = descriptions[@feature.key]
           @page_title = "#{@feature.key} // Features"
           @percentages = [0, 1, 5, 10, 15, 25, 50, 75, 100]
 
--- a/lib/flipper/ui/actions/features.rb
+++ b/lib/flipper/ui/actions/features.rb
@@ -10,8 +10,11 @@ module Flipper
 
         def get
           @page_title = 'Features'
+          descriptions = Flipper::UI.configuration.descriptions_source
           @features = flipper.features.map do |feature|
-            Decorators::Feature.new(feature)
+            decorator = Decorators::Feature.new(feature)
+            decorator.description = descriptions[feature.key]
+            decorator
           end.sort
 
           @show_blank_slate = @features.empty?
--- a/lib/flipper/ui/configuration.rb
+++ b/lib/flipper/ui/configuration.rb
@@ -35,6 +35,11 @@ module Flipper
       # application needs. Defaults to "a flipper id".
       attr_accessor :add_actor_placeholder
 
+      # Public: If you set this, Flipper::UI will fetch descriptions
+      # from your external source. Descriptions for `features` will be shown on `feature`
+      # and `features` pages. Defaults to empty hash.
+      attr_accessor :descriptions_source
+
       VALID_BANNER_CLASS_VALUES = %w(
         danger
         dark
@@ -58,6 +63,7 @@ module Flipper
         @feature_removal_enabled = true
         @fun = true
         @add_actor_placeholder = "a flipper id"
+        @descriptions_source = ->(args) {}
       end
 
       def banner_class=(value)
--- a/lib/flipper/ui/decorators/feature.rb
+++ b/lib/flipper/ui/decorators/feature.rb
@@ -11,6 +11,8 @@ module Flipper
         # Public: The feature being decorated.
         alias_method :feature, :__getobj__
 
+        attr_accessor :description
+
         # Public: Returns name titleized.
         def pretty_name
           @pretty_name ||= Util.titleize(name)
@@ -22,6 +24,7 @@ module Flipper
           {
             'id' => name.to_s,
             'name' => pretty_name,
+            'description' => description,
             'state' => state.to_s,
             'gates' => gates.map do |gate|
               Decorators::Gate.new(gate, gate_values[gate.key]).as_json
--- a/lib/flipper/ui/views/feature.erb
+++ b/lib/flipper/ui/views/feature.erb
@@ -12,6 +12,9 @@
           <span class="octicon octicon-squirrel <%= @feature.color_class %>"></span>
           <%= @feature.key %>
         </h4>
+        <% if @feature.description %>
+          <p><%= @feature.description %></p>
+        <% end %>
         <div class="card-body">
           <p>
             <% if @feature.boolean_value %>
--- a/lib/flipper/ui/views/features.erb
+++ b/lib/flipper/ui/views/features.erb
@@ -27,6 +27,7 @@
             <span class="octicon octicon-squirrel"></span>
           </th>
           <th class="col">Feature</th>
+          <th class="col-6">Description</th>
           <th class="col">Enabled Gates</th>
         </tr>
       </thead>
@@ -39,6 +40,9 @@
             <td class="col">
               <a href="<%= "#{script_name}/features/#{feature.key}" %>">
                 <%= feature.key %></a>
+            </td>
+            <td class="col-6">
+              <%= feature.description %>
             </td>
             <td class="col">
               <%= feature.pretty_enabled_gate_names %>
--- a/spec/flipper/ui/configuration_spec.rb
+++ b/spec/flipper/ui/configuration_spec.rb
@@ -113,4 +113,24 @@ RSpec.describe Flipper::UI::Configuration do
       expect(configuration.fun).to eq(false)
     end
   end
+
+  describe "#descriptions_source" do
+    it "has default value" do
+      lambda do |key_name, description|
+        expect(key_name).to eq nil
+        expect(description).to eq nil
+      end
+    end
+
+    context "descriptions source is provided" do
+      it "can be updated" do
+        file = YAML.load_file(FlipperRoot.join('spec/support/descriptions.yml'))
+        configuration.descriptions_source = -> { file }
+        lambda do |key_name, description|
+          expect(key_name).to eq("some_awesome_feature")
+          expect(description).to eq("Awesome feature description")
+        end
+      end
+    end
+  end
 end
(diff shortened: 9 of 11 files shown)


annaswims:
I like this idea, for the most part but Id suggest making the yaml file a bit more generic.  

At vag.gov we have a [features.yml](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/features.yml) file and we use it for initializing features in all environments, descriptions, specifying the actor (sometimes we want a toggle for logged in users, sometimes we want a toggle for all users with a cookie value). Our latest addition to the features.yml file is that we want to specify default values for the toggle by environment (i.e. most, but not all features are enabled by default in development when they're initialized).

juanroldan1989:
> I like this idea, for the most part but Id suggest making the yaml file a bit more generic.
> 
> At VA.gov we have a [features.yml](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/features.yml) file and we use it for initializing features in all environments, descriptions, specifying the actor (sometimes we want a toggle for logged in users, sometimes we want a toggle for all users with a cookie value). Our latest addition to the features.yml file is that we want to specify default values for the toggle by environment (i.e. most, but not all features are enabled by default in development when they're initialized).

Thank you @annaswims for your feedback! really appreciated : ) adjustments made 🚀

jnunemaker:
I’ll try to give this a look over tomorrow. Sorry for delay.

juanroldan1989:
> I’ll try to give this a look over tomorrow. Sorry for delay.

Thank you @jnunemaker ! really appreciated : )

Test suite is failing because of these un-related details:

```ruby
Inspecting 220 files
......................................................................................................................................................................W.......................................C.............
Offenses:
spec/flipper/feature_spec.rb:211:88: W: Lint/RedundantCopDisableDirective: Unnecessary disabling of RSpec/RepeatedDescription.
      it "always instruments #{thing.inspect} as #{wrapped_thing.class} for enable" do # rubocop:disable RSpec/RepeatedDescription
                                                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/flipper/feature_spec.rb:300:90: W: Lint/RedundantCopDisableDirective: Unnecessary disabling of RSpec/RepeatedDescription.
      it "always instruments #{thing.inspect} as #{wrapped_thing.class} for enabled?" do # rubocop:disable RSpec/RepeatedDescription
                                                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/integration_spec.rb:4:1: C: RSpec/FilePath: Spec path should end with flipper*_spec.rb.
RSpec.describe Flipper do
^^^^^^^^^^^^^^^^^^^^^^
220 files inspected, 3 offenses detected
```

Perhaps Rubocop settings for `Lint/RedundantCopDisableDirective` could be adjusted or we could just fix these issues. Either way, I can give you a hand with that 🚀 

Cheers 🍻

juanroldan1989:
Failing tests are un-related to this PR - There's a separate PR addressing that: https://github.com/jnunemaker/flipper/pull/462

juanroldan1989:
> Failing tests are un-related to this PR - There's a separate PR addressing that: #462

PR #462 rebased on this PR - test suite green again : )

jnunemaker:
@juanroldan1989 you are a fantastic contributor. You've setup a great PR and completed a lot of other work to get things green and help out. Thanks. 

My struggle with this PR is that it uses YAML. What if someone wanted to store the descriptions somewhere that someone could update them like mysql or postgres? They'd be out of luck. Part of what took me so long to respond was thinking through what I might do instead.

```ruby
  # keys would be an Array
  config.description_fetcher = ->(keys) do
    # do your yaml load or query mysql or whatever you want
    # return would be hash of {String key => String description}
  end
end
```

I don't love the `description_fetcher` name but I'm lacking a better name currently. `description_loader` and `description_finder` were other thoughts. I hate them too. :smile: 

Something like above would allow you to use yaml and allow anyone else to use whatever they want. I'd rather have native storage of other attributes like description, but I totally get this is useful and want to make it possible for you so I think something like above is what I'd lean toward so it is more flexible. 

What are your thoughts? Would you be interested in take a stab at it? If so, do you even have time? If not, let me know and I can try to whip something together. I just want to make sure whatever we do works for everyone who uses flipper rather than on specific use case.

annaswims:
I think the feature toggles straddle the awkward line between what's code and data. If I wanted to keep a list of all features that may be toggled, I'd lean towards doing that as code.  That way, I commit the feature toggle name and description in the same PR that I could write the code to check the feature value.  When a new feature is added, I'd prefer to have it added to the interface and defaulted to disabled rather than not visible in the interface at all. 

It doesn't seem like the right flow for me for a developer to merge a PR that references a feature toggle, wait for deploy, then add the feature to the interface for staging and prod (or add the feature as a migration).

juanroldan1989:
> @juanroldan1989 you are a fantastic contributor. You've setup a great PR and completed a lot of other work to get things green and help out. Thanks.

Thank you @jnunemaker ! I really appreciate that : ) the tool is such a gem that is really hard not to take care of it honestly 🚀 
> 
> My struggle with this PR is that it uses YAML. What if someone wanted to store the descriptions somewhere that someone could update them like mysql or postgres? They'd be out of luck. Part of what took me so long to respond was thinking through what I might do instead.

It makes total sense now I think of it.

Honestly, I went with the "descriptions loaded from a YAML file" idea simply because that's how several libraries that I've used over time do it, e.g: `devise`, `figaro`, `activeadmin`, `simple_form`, `wicked`, `wicked-pdf`, `money_rails`.

Allowing `descriptions` to be loaded with as **many options** as possible is `key`, no pun intended haha, to keep this feature available for everyone who wants to improve their experience using it. I like that! 💯 

> 
> ```ruby
> Flipper::UI.configure do |config|
>   # keys would be an Array
>   config.description_fetcher = ->(keys) do
>     # do your yaml load or query mysql or whatever you want
>     # return would be hash of {String key => String description}
>   end
> end
> ```
> 
> I don't love the `description_fetcher` name but I'm lacking a better name currently. `description_loader` and `description_finder` were other thoughts. I hate them too. 😄
> 

How about `config.descriptions_source = ->(keys) do` ? 🤔 

> Something like above would allow you to use yaml and allow anyone else to use whatever they want. I'd rather have native storage of other attributes like description, but I totally get this is useful and want to make it possible for you so I think something like above is what I'd lean toward so it is more flexible.
> 

Totally agree with this. This is the second company I worked with where the subject comes up at some point and questions emerge within team members:
- "Yeah I can read the key name but ... what was this feature for?"
- "This key is really too long but hey, works for me and the rest of the Admins"
- "Maybe we should add another `*_extra_info` section for keys to make it more understandable"

That's the reason why I wanted to give it a shot and see how far we could go to solve this issue, or at least provide folks with an option to make things clearer : ) hope that makes sense.

> What are your thoughts? Would you be interested in take a stab at it? If so, do you even have time? If not, let me know and I can try to whip something together. I just want to make sure whatever we do works for everyone who uses flipper rather than on specific use case.

I'm not sure where to start regarding setting up more **sources** for `Flipper::UI`. More questions come to me than answers:

- How many options to connect should we provide? E.g: Redis, Postgres, MySQL, ...
- Maybe we should define an `entry_point` for the database within `Flipper::UI` configuration module?
- Should we define an `entry_point_type` for connectors, with possible values like: `yaml`, `pg`, `mysql`, `redis`. This `entry_point_type` will be used to decided which source read descriptions from.

I would love to see your approach on this and learn of that process 🚀 

I really appreciate the feedback on this and all the possibilities involved! Thank you ! 🍻

jnunemaker:
> How about config.descriptions_source

👍 Works for me.

> I'm not sure where to start regarding setting up more sources

I wouldn't worry about building all the adapters for different data stores. I'd just focus on adding description_source as a block that takes keys and returns key => description. Then making the UI use that block to get the descriptions to show. Does that simplify things? Just ruby. No yaml or mysql or anything. You could easily add an example in the examples/ui folder that shows how to use it with yaml or something if you wanted.

jnunemaker:
For example, once the descriptions_source is in Flipper::UI with a default empty block, you could do this in the features action:

```ruby
module Flipper
  module UI
    module Actions
      class Features < UI::Action
        route %r{\A/features/?\Z}

        def get
          @page_title = 'Features'
          descriptions = Flipper::UI.configuration.description_source(flipper.features.map(&:key))
          @features = flipper.features.map do |feature|
            decorator = Decorators::Feature.new(feature)
            decorator.description = descriptions[feature.key]
            decorator
          end.sort
  # ... clipped for brevity ...
```

You could then do the same thing for showing a feature. You would need an `attr_accessor :description` on the feature decorator or something like that. 

Does that help?

juanroldan1989:
@jnunemaker I've added the latest changes 🚀 

While working on these, one question came up:

Should we or should we not include a **default** message whenever a `description` is not found within whatever **source** provided for a specific `key`?

Currently we show `Description not provided` message in both `feature` and `features` pages.

Leaving this **as is** now, whenever a developer updates the gem to its most up to date version, all of a sudden such message would pop up on features listing and feature pages.

The **positive** side of this would be: developers realising they can now add/edit descriptions for features, prompting them to go to the docs and see how it can be done. (besides of course acknowledging the fact that developers using this gem for the first time would **read** docs about how the gem works).

The **not-so-positive** side of this would be: for projects with the gem already implemented, in case developers/users/admins don't like the `Description not provided` message to be displayed, they would need to provide let's say a `-` content for keys they don't need a description for.

Perhaps, we could just leave this **default** description as an empty string **and** expect/hope/know-for-sure developers in need of descriptions for their features will read the docs and find out about the existing configuration to provide them ?

Anyway, food for thought : )

hstebbins-dailypay:
Love this idea, thanks for putting in the work to get this PR up. Is there anything I can help with to get it over the finish line? 

Not that my opinion matters much, but an empty string makes more sense to me as a default. It makes it feel more like an optional feature rather than something that's required/expected, if that makes sense.

juanroldan1989:
@hstebbins-dailypay thanks for your feedback ! greatly appreciated as well : D

I've made up my mind thinking again about that 🚀 

I've just updated the PR so whenever a `key` does not have a `description` associated no message is displayed.

This way developers will only need to care about keys/features they **do** want to have a `description` available for.

docs/ui/README.md
```diff
@@ -148,6 +148,17 @@ Flipper::UI.configure do |config|
 end
 ```
 
+We can also associate a `description` for each `feature` by providing a descriptions source:
+
+```ruby
+Flipper::UI.configure do |config|
+  config.descriptions_source = ->(keys) do
```
jnunemaker, line 155: This documentation doesn't match the code. It shows setting it to a block here, but is defaulted to a hash in ui config. If it were suppose to be a block, it should be defaulted to an empty block in ui config, right?

lib/flipper/ui/util.rb
```diff
@@ -11,6 +11,10 @@ def self.blank?(str)
       def self.titleize(str)
         str.to_s.split('_').map(&:capitalize).join(' ')
       end
+
+      def self.descriptions
```
jnunemaker, line 15: Having this here means once the process has booted and set the descriptions there is no changing them without a restart or redeploy, right? The first part of descriptions_source is allowing mysql or whatever to set the descriptions. The second part, that I think is missing from this PR is allowing people to add descriptions anytime and having them show up, without needing to restart processes or redeploy their app. Adding a row to mysql should make the description just show up, but I think this would mean that they would have to restart to force a reload. 

What are your thoughts on that? Also, what are your thoughts on moving to a block for loading and putting that in the actions instead of here?

juanroldan1989: > Having this here means once the process has booted and set the descriptions there is no changing them without a restart or redeploy, right? 

I really hadn't think of that up until this point. I wasn't aware of that to be honest. For sure, let's move the block call to the `actions` module for both `features` and `feature` resources 🚀

jnunemaker:
A few parts weren't actually wired up right in the specs and the UI but it was close enough so I pulled it down, got it working and merged it. Thanks for all your help! I'll release and blog about it soon hopefully.

juanroldan1989:
> A few parts weren't actually wired up right in the specs and the UI but it was close enough so I pulled it down, got it working and merged it. Thanks for all your help! I'll release and blog about it soon hopefully.

Thank you @jnunemaker ! I was hoping it was getting there haha 🎊 I really appreciate your guidance and insights on this feature : D

Kind Regards 🍻

juanroldan1989:
> A few parts weren't actually wired up right in the specs and the UI but it was close enough so I pulled it down, got it working and merged it. Thanks for all your help! I'll release and blog about it soon hopefully.

Thank you @jnunemaker ! I was hoping it was getting there haha 🎊 I really appreciate your guidance and insights on this feature : D

Kind Regards 🍻

PR #468, opened 2020-06-12
Remove unnecessary rubocop disabling

hstebbins-dailypay:
Hi folks, this PR removes two unnecessary rubocop disablings. These disablings cause the CI to fail as they violate linting rules. 

Here's the failing output, without this PR:
```
$ rubocop
Inspecting 220 files
......................................................................................................................................................................W.....................................................

Offenses:

spec/flipper/feature_spec.rb:211:88: W: Lint/RedundantCopDisableDirective: Unnecessary disabling of RSpec/RepeatedDescription.
      it "always instruments #{thing.inspect} as #{wrapped_thing.class} for enable" do # rubocop:disable RSpec/RepeatedDescription
                                                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/flipper/feature_spec.rb:300:90: W: Lint/RedundantCopDisableDirective: Unnecessary disabling of RSpec/RepeatedDescription.
      it "always instruments #{thing.inspect} as #{wrapped_thing.class} for enabled?" do # rubocop:disable RSpec/RepeatedDescription
                                                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

220 files inspected, 2 offenses detected
```

### DIFF
--- a/spec/flipper/feature_spec.rb
+++ b/spec/flipper/feature_spec.rb
@@ -208,7 +208,7 @@ RSpec.describe Flipper::Feature do
       percentage_of_time => percentage_of_time,
       percentage_of_actors => percentage_of_actors,
     }.each do |thing, wrapped_thing|
-      it "always instruments #{thing.inspect} as #{wrapped_thing.class} for enable" do # rubocop:disable RSpec/RepeatedDescription
+      it "always instruments #{thing.inspect} as #{wrapped_thing.class} for enable" do
         Flipper.register(:admins) {}
         subject.enable(thing)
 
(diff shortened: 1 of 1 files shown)


jnunemaker:
I'm going to close this because I am doing #469 instead, but I **really** appreciate you noticing this and taking the time to fix it. :heart: 

For background, these are what led to #469. On my laptop it was requiring me to add those and on travis it was not. I tried locking versions and few other things, but the slow feedback loop with travis ci runs and the general pain I've felt using rubocop the past few years means I'm just done with it. Maybe I'll use it opt in on a few specific styles in the future that I care about but most of it feels unimportant to me.

PR #478, opened 2020-08-13
Get specs passing in docker

jbarrieault:
Most specs relying on Memcached are failing when run in the docker environment. 

This was happening because the Dalli clients used weren't being passed a server URL, and were defaulting to `127.0.0.1:11211`. That works when running the specs in a typical local environment because memcached is usually exposed at that IP/port. But docker exposes it elsewhere!

### DIFF
--- a/spec/flipper/adapters/dalli_spec.rb
+++ b/spec/flipper/adapters/dalli_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe Flipper::Adapters::Dalli do
   let(:memory_adapter) do
     Flipper::Adapters::OperationLogger.new(Flipper::Adapters::Memory.new)
   end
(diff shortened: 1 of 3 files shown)


spec/flipper/adapters/dalli_spec.rb
```diff
@@ -7,7 +7,7 @@
   let(:memory_adapter) do
     Flipper::Adapters::OperationLogger.new(Flipper::Adapters::Memory.new)
   end
-  let(:cache)   { Dalli::Client.new(ENV['MEMCACHED_URL'] || '127.0.0.1:11211') }
```
jbarrieault, line 10: Removing the fallback since that would also cause failure in the docker environment if used (_and_ it is the fallback Dalli [already uses](https://github.com/petergoldstein/dalli/blob/ca8b286de230546c3bfe2789e9b6bc5a2c619570/lib/dalli/client.rb#L35
)).

Issue #163, opened 2016-09-29
V2 Adapters


jnunemaker:
While thinking about or working on [description for features](https://github.com/jnunemaker/flipper/issues/26), [custom gates for features](https://github.com/jnunemaker/flipper/issues/121) and [runtime controls](https://github.com/jnunemaker/flipper/issues/162), it became apparent that the current adapter style of a method per thing we want to store isn't going to scale to more functionality well. 

## The Beginning

Initially when I started with adapters for flipper (way back in the day), the interface was more key/value based (get, set, delete). All the information for a feature was stored in one key (other rollout tools do similar). 

### The Benefits of Key/Value

* **flexibility** - we can store a feature in one key and serialize whatever data makes sense for the feature, e. g. a description for the feature, a collection of disabled gates, or even use the key/value style for new features (think runtime controls). Several new features (just listed and also listed above) are kind of stuck because of the amount of work needed not having this flexibility is annoying.
* **simplicity** - Making an adapter becomes dramatically easier. All adapter creators would need is get, set and delete.

### The Problem (read then write race)

The main problem with key/value and storing all information in a feature (to make reads faster) is that without some form of locking, race conditions exist. Imagine the following scenario:

* developer 1 views "profile_next" feature
* developer 2 views "profile_next" feature
* developer 1 adds "fred" to enabled actors and developer 2 removes "barney" from enabled actors at the same time, flipper does a `get` for the feature, modifies the actor set and then does a `set` for the feature
* whichever `set` happens last nullifies the other developers change, this would be frustrating/confusing to not see a change you just made that appeared to be successful

### A Solution (to read then write race)

The solution is to either lock around the read/write operation or use datastore functionality that allows adding or removing a thing from a set (`sadd`/`srem` in redis, `insert`/`delete` in mysql with unique index, etc.). For this reason, I went with making it so adapters could take advantage of data store functionality to avoid this read then write race. That is why adapters have an enable and disable method, to allow for adapter creators to customize the operation of enabling or disabling something to be atomic for the backend they are adapting to.

### The Reality

The reality is that features are enabled and disabled very rarely compared to how often they are checked. Write throughput (enable/disable) is usually so low that the chance of this race happening is extremely low. 

[...]

Issue #121, opened 2016-05-10
Allow customizing which gates are used per feature


jnunemaker:
Currently all gates are used for all features all the time. I've found that sometimes I don't pass in an actor, so the % of actors, actor and group gates don't make sense for the feature. I'd like to be able to deactivate those gates and have that show up in the UI and DSL. In the UI, we could hide the forms for deactivated gates. In the DSL, we could raise if you attempt to enable/disable those gates. 

As far as naming, enabled/disabled is pretty dominant in flipper so I don't want to and really can't use them for this. Thinking about it for a bit all I could come up with is activate/deactivate/activated?. enabled/disabled/enabled? are for determining feature enabled-ness and activate/deactivate/activated? would be for determining if the gates should be used when determining feature enabled-ness.

As far as where this should be stored, the easiest thing is to just store it in memory in the code. The stinky thing is that if we do that any place that mounts the UI or API will also need that shared code. I may start with this to see what the feature feels like overall and then look at how to persist it at the adapter level in a way that it doesn't need to be defined in code and will work wherever.

Some code scribbles so I remember:

```ruby
# will need DSL methods as well, but here is what it would/could look like from feature
flipper[:search].activate(:boolean)
flipper[:search].deactivate(:boolean)
flipper[:search].gate_activated?(:boolean)

flipper[:search].gate(:boolean).activate
flipper[:search].gate(:boolean).deactivate
flipper[:search].gate(:boolean).activated?
```

jnunemaker:
I've got a branch going for this, but something doesn't feel right. 
https://github.com/jnunemaker/flipper/compare/custom-gates

I was trying to do in memory first and avoid adapter stuff, but there really is no way. You have to at least involve the adapter when deactivating, or you end up with weird state in the UI. For example, let's say you enable an actor, then deactivate the actors gate, the feature shows up as partially enabled, but that isn't true. Because the gate is deactivated, it is actually disabled. The way to fix that is to clear all gate values when deactivating a gate. The problem with that is that there is no adapter method to clear all values for a gate. Currently it only works to clear a single value for a gate. That means we need a new method for the adapters to clear all values. The alternative is to do a get and then a disable for each value (ie: get all actors and disable each). 

activate/deactivate also have to permeate the code. feature needs to know what gates are active before it does anything (enable, disable, etc.). The UI needs to know so it can hide things and deny certain actions. The API needs the same.

[...]

PR #481, opened 2020-08-27
Proposed feature: Check if enabled only for some actors

iamvery:
## Context

I realize that this may not be a feature you want to support, but it came up in a use case that we have for Flipper, so I wanted to at least open the conversation here in case there is an opportunity to pay it forward. I will also say that the PR as it is being opened is more a conversation starter than anything, so please know that I'm not deeply committed to the implementation. I'm happy to iterate on the solution.

## Problem

We recently started using Flipper as a way of empowering some testing-in-production features in our infrastructure. In particular, we wanted a way to flag certain domain object flowing through our system to behave in a certain way. For example:

> Any time user:1234 attempts to buy something, behave as if their card is declined.

To do this, we use Flipper, but as a safety mechanism we added an additional check to make certain that the features aren't enabled globally, because we wouldn't want to decline every purchase for all users by accident 😎 

## Solution

Our solution involved a small wrapper around Flipper, so we thought I might be something to add to the library. It essentially amounts to checking for a feature which is globally enabled and faulting on that condition.

Please let me know your thoughts! I would love to understand your perspective on this problem and whether it is something you could like to include in Flipper.

### DIFF
--- a/lib/flipper/dsl.rb
+++ b/lib/flipper/dsl.rb
@@ -34,6 +34,14 @@ module Flipper
       feature(name).enabled?(*args)
     end
 
+    def enabled_for_some?(name, *args)
+      if enabled?(name)
+        raise Flipper::Error, "The feature #{name.inspect} is enabled for all!"
+      end
+
+      enabled?(name, *args)
+    end
+
     # Public: Enable a feature.
     #
     # name - The String or Symbol name of the feature.
--- a/lib/flipper/dsl.rb
+++ b/lib/flipper/dsl.rb
@@ -34,7 +34,7 @@ module Flipper
       feature(name).enabled?(*args)
     end
 
-    def enabled_for_some?(name, actor)
+    def strict_enabled?(name, actor)
       if enabled?(name)
         yield if block_given?
         return false
--- a/lib/flipper/dsl.rb
+++ b/lib/flipper/dsl.rb
@@ -36,7 +36,8 @@ module Flipper
 
     def enabled_for_some?(name, actor)
       if enabled?(name)
-        raise Flipper::Error, "The feature #{name.inspect} is enabled for all!"
+        yield if block_given?
+        return false
       end
 
       enabled?(name, actor)
--- a/lib/flipper/dsl.rb
+++ b/lib/flipper/dsl.rb
@@ -34,12 +34,12 @@ module Flipper
       feature(name).enabled?(*args)
     end
 
-    def enabled_for_some?(name, *args)
+    def enabled_for_some?(name, actor)
       if enabled?(name)
         raise Flipper::Error, "The feature #{name.inspect} is enabled for all!"
       end
 
-      enabled?(name, *args)
+      enabled?(name, actor)
     end
 
     # Public: Enable a feature.
--- a/lib/flipper.rb
+++ b/lib/flipper.rb
@@ -56,7 +56,7 @@ module Flipper
   # Public: All the methods delegated to instance. These should match the
   # interface of Flipper::DSL.
   def_delegators :instance,
-                 :enabled?, :enabled_for_some?, :enable, :disable, :bool, :boolean,
+                 :enabled?, :strict_enabled?, :enable, :disable, :bool, :boolean,
                  :enable_actor, :disable_actor, :actor,
                  :enable_group, :disable_group,
                  :enable_percentage_of_actors, :disable_percentage_of_actors,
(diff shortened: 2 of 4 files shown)


lib/flipper/dsl.rb
```diff
@@ -34,6 +34,14 @@ def enabled?(name, *args)
       feature(name).enabled?(*args)
     end
 
+    def enabled_for_some?(name, actor)
+      if enabled?(name)
+        raise Flipper::Error, "The feature #{name.inspect} is enabled for all!"
```
iamvery, line 39: It probably doesn't make sense to raise an error. I did so on a whim here, because I think it would be desirable to provide some logging in the case that this condition happens. I think a better solution might be an optional block or argument that can be leveraged by the caller to respond as needed.

iamvery: I went ahead and converted this to a block in https://github.com/jnunemaker/flipper/pull/481/commits/b962c0b574d3f1f6641d659efeb0b0b05ccecf89.

lib/flipper/dsl.rb
```diff
@@ -34,6 +34,15 @@ def enabled?(name, *args)
       feature(name).enabled?(*args)
     end
 
+    def enabled_for_some?(name, actor)
```
iamvery, line 37: My colleague made a good point, that this is mostly about "enabled for actors" or actor-only feature flags that aren't enabled by percentage or globally. That's a much better way to put it. There may be a better way to model this deeper in the abstractions, but my hot take would be name this method `strict_enabled?`.

lib/flipper/dsl.rb
```diff
@@ -34,6 +34,15 @@ def enabled?(name, *args)
       feature(name).enabled?(*args)
     end
 
+    def strict_enabled?(name, actor)
+      if enabled?(name)
```
iamvery, line 38: This still leaves open the possibility of enabled by percentage which could also be problematic 🤔

jnunemaker:
Could you show a couple examples of how this is used in the code? I can see and understand the code/specs, but I kind of want to see it in real life so I better understand the use case and if there is a different way we could maybe solve it using what already exists. I think a few real life code examples would help me understand. Thanks!

iamvery:
Yes, of course! Here is an example.

Give some small testing-in-production abstraction like:
```ruby
module TestingInProduction
  def self.simulate_decline?(user)
    Flipper.strict_enabled?(:simulate_decline, user, on_reject: ->(feature){
      Rails.logger.warn <<~MSG
        The feature #{feature.inspect} is enabled too liberally to be used safely for testing-in-production.
      MSG
    })
  end
end
```
Note: In this example, I went with an `on_reject:` argument rather than a block as presented in the current PR code changes. Again just illustrating options. I don't have a strong opinion.

And later on in application code:
```ruby
class PurchaseController < ApplicationController
  def create
    # Note: We only want this to return `true` when the user is an enabled actor. Not when the feature is enabled
    # by percentage or globally as that would affect other users in an unpredictable or unintended way.
    if TestingInProduction.simulate_decline?(user)
      render json: { error: "declined" }, status: 400
    else
      # ... (do things to actually complete the purchase)
      render json: { success: true }
    end
  end
end
```

I'm going to noodle more on another use case. We've repeated a pattern similar to this a few times, but it's all in the name of testing-in-production, so I don't have another use case handy in mind right now.

iamvery:
In discussion with a colleague, another interesting perspective came up. Potentially this feature could be limited to flipper-ui in that the primary risk that we're addressing is the potential for mistakenly toggling a feature on in a more "global" sense, so perhaps there would be room to add some level of restriction to the UI that is configurable. For example, disable the "Fully Enable" button for particular features. 🤔

jnunemaker:
Interesting. Seems like what you really want is a way to remove certain gates from a feature and disallow using them. I actually definitely want to support that. I haven't built it because it will involve adapter changes -- either some sort of KV or config storage per feature or switching to v2 #163  adapters that are more KV based and dumber, but with the ability to store more.

jnunemaker:
@iamvery Did you end up working around this or is it still hanging open? Just curious. I'm probably going to close it due to age, but I'm very happy to keep working through this (and will do a better job of responding promptly, sorry about that).

jnunemaker:
Oh, and [here is the issue](https://github.com/jnunemaker/flipper/issues/121) about customizing which gates are used per feature.

iamvery:
I'm not actively working on this or thinking about it, so thank you for closing it out. For the time being on our side we're not actively making changes. If something changes, I'll come back to this. Thanks for responding! And no problem at all on the timing 🙏

PR #485, opened 2020-11-25
Use a middleware example for use with Warden

hlascelles:
The original example would give 404 when you are not authorised, which is not what you'd expect. 

Using middleware we can return a 403.

### DIFF
--- a/docs/ui/README.md
+++ b/docs/ui/README.md
@@ -86,20 +86,28 @@ end
 Another example is to use the `current_user` when using a gem-based authentication system (i.e., [warden](https://github.com/hassox/warden) or [devise](https://github.com/plataformatec/devise)):
 
 ```ruby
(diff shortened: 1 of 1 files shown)


jnunemaker:
Hi there! Really appreciate you taking the time to both read, think about and update the docs. 

I'd be good with showing this as an example in addition, but I like the original. 404 is intentional as it hides where flipper is mounted. 

The thought process is akin to how github does repos. If you go to a private repo that you don't have access to, you get a 404 status. 

If you return 403 instead of 404, people can use that to determine if (and where) flipper is mounted and try to attack. If you return 404 then they have less of an idea. 

Closing this to keep things clean, but if you want to add that example as another section on the doc page, I'd be happy to merge it for those that want to do it that way and are curious how.

PR #487, opened 2020-11-30
Bump rack-protection version to < 2.2

jeobrien:
In https://github.com/jnunemaker/flipper/pull/156 rack-protection was set to >= 1.5.3 and < 2.1.0, which was because it no longer worked with `session["_csrf_token"]`, but did work with `session[:csrf]` in both 1.5.3 and 2.0.0.beta2. Hoping to update it now that that beta changes are live, as this allows installation of flipper alongside the current version of rack-protection.


jnunemaker:
:cool: Thanks for caring! 

Looks good to me. I'll wait for CI to pass and then merge.

PR #490, opened 2020-12-28
Use Rails.configuration instead of config in documentation

clinejj:
This updates the documentation for Optimization to use `Rails.configuration.middleware` instead of `config.middleware`, which makes this a bit more copy/paste ready for new users. As the first comment suggests the code is in `config/initializers/flipper.rb`, the code as described doesn't work. Updating as `Rails.configuration.middleware` makes it ready to go.


jnunemaker (approved):

jnunemaker:
@clinejj thanks!

Issue #82, opened 2015-09-04
Allow making actors disabled when feature is globally turned on


thenonameguy:
In the current UI there is no way to disable for site administrators to disable a certain feature for a given user/account.

jnunemaker:
See caveat 2 here: https://github.com/jnunemaker/flipper/blob/1c1b18763190dcfda6df41133ceb6aa63c53c35e/docs/Caveats.md

Flipper is for explicitly letting people through gates, not keeping them out. That said, can you further describe your use case? Maybe there is a currently supported way of doing what you need to do or maybe it makes sense to add.

thenonameguy:
We are guarding one of our product features with a feature flipper. After enabling it for a few actors, we enabled it globally, but one of our customers did not want to expose it for their accounts users. Our site admins did not find a way to disable it only for them on the UI, other than turning the feature off and listing every actor manually.

jnunemaker:
Ah, yeah so not a common case probably. One other way to solve it could be a group. You could make a group and enable it like this:

```ruby
Flipper.register(:some_group) do |actor|
  !disabled_customer_ids.include?(actor.id)
end

Flipper[:some_feature].disable
Flipper[:some_feature].enable(:some_group)
Flipper[:some_feature].enabled?(disabled_customer) # false
Flipper[:some_feature].enabled?(other_customer) # true
```

rmosolgo:
I just came looking for something like this too:

In the course of rolling out a feature, there are a few specific cases where it doesn't seem to work yet 😖   So I found myself looking for a flow like this: 

- Enable a feature for everyone; THEN 
- Disable the feature for a few actors 

I found a workaround, using two different features:

- "Use Feature X": Enabled 
- "Blacklist Feature X": Enable for a few actors

then do the application logic like this: 

```ruby
def feature_x_enabled?
  if flag_enabled?("Blacklist Feature X")
    false 
  else 
    flag_enabled?("Use Feature X")
  end 
end 
```

It's a bit tough to have the gate logic split between two, but it provides the behavior I was looking for!

rmosolgo:
> I still wonder if the complexity is worth it

I would guess _not_ since there's about one request per year, judging by this thread 😆 

I'm happy to "work around" it, and like you said, this is a very fair design goal/constraint: 

> Flipper is for explicitly letting people through gates, not keeping them out. 

[...]

PR #494, opened 2021-01-26
Adds method disabled, inverse of enabled

kinduff:
Hello there!

First of all, a lot of thanks to @jnunemaker, the maintainers and the community for this wonderful gem. We use it a lot @dmstk ([domestika.org](https://domestika.org) and we can't be more happy.

One thing we've noticed with heavy use (around 20-25 feature flags and more to come) is that we usually end up doing this in order to have the inverse of a gate:

```ruby
return if counter == 1 and !Flipper.enabled?(:feature)
```

This is okay, but it's not as explicit as we would like.
v
The proposed change adds a new method in the `Gate` class in order to have the inverse of the method, like so:

```ruby
return if counter == 1 and Flipper.disabled?(:feature)
```

We can of course do this downstream, but having it as part of the gem would be golden.

I did not see the necessity of adding tests for this since the `enabled?` method is being tested in all the gates, but let me know if you want me do so.

Thank you!

### DIFF
--- a/lib/flipper/gate.rb
+++ b/lib/flipper/gate.rb
@@ -22,6 +22,10 @@ module Flipper
       raise 'Not implemented'
     end
 
+    def disabled?(value)
+      !enabled?(value)
+    end
+
     # Internal: Check if a gate is open for a thing. Implemented in subclass.
     #
     # Returns true if gate open for thing, false if not.


kinduff:
@jnunemaker any chance this can get a look? if it's something you're not looking to have, happy to close it up

bkeepers:
@kinduff could you explain more about the real world use case? I know you gave an abstract example, but my brain is having a hard time imagining a use case where `unless` wouldn’t fit the bill. 

It doesn’t seem like it would be hard to add ‘disabled?’, I’m just curious to see a concrete example.

jnunemaker:
@kinduff I could have sworn I responded to this. But I must have in my head and not on github because here we are. 😆 Sorry about that! 

Something similar to this (maybe the same thing) has come up before ([previous response](https://github.com/jnunemaker/flipper/issues/82#issuecomment-759626499)). Can you take a look at that and let me know if you have any thoughts? Tried to link directly to my response but you might need to scan up or down a bit for more context if what I'm saying doesn't make sense. 

tldr: The primary issue is flipper is about enabling things and letting them through gates, not disabling things. Making flipper work as is for disabling is easy for storage (the adapters) but makes the super simple api of flipper (a selling point IMO) a lot more complicated to understand. 

I really want to make this work somehow. But it is not an easy, simple change. If anyone comes up with a simple solution or even just a bunch of responses that make sense to the matrix I listed above, I'd be happy to implement it or merge a PR. I'll probably sit down with @bkeepers soon and talk through this one. We'll probably be able to come up with something that I couldn't on my own.

jnunemaker:
Closing to keep things tidy but I really do want to keep talking about this if you have more thoughts.

jnunemaker:
I actually wonder if instead of calling it `disabled?` (which feels right but could fudge things down the road), we call it `not_enabled?`. Flipper.not_enabled?(...) isn't as pretty but is more descriptive of what is actually happening. We are checking to see if something is disabled. We are only checking to see if it is NOT enabled. Subtle but different.

kinduff:
Hey @jnunemaker! Totally missed the notifications around this.

I actually like the idea a lot more, and makes sense given the context you posted, as well as to keep things simple. The word disabled can lead to incorrect assumptions of which way the stuff needs to be toggled.

Regarding the use case @bkeepers asked about, I cannot think of a clearer example as the one I posted. I think it really depends on the implementation. We usually need to add features that are both toggleable and backward compatible, and a method like this instead of using a bang operator can clean up our code a lot.

Happy to contribute and add a PR to add the method, or even continue with the discussion.

jnunemaker:
@kinduff just wrote up more here https://github.com/jnunemaker/flipper/issues/514. I really like being able to disable specific actors as well. That said, it might not quite aline with what you are attempting. 

I think what @bkeepers was trying to ask was why not just do this:

```ruby
return if counter == 1
return unless Flipper.enabled?(feature)
```

I often do a set of guards at the beginning of a method like that rather than checking them all on one line. Views are probably the only place where more than one conditional on a line happens for me.

PR #496, opened 2021-02-18
Replace feature_name.strip for gsub

gabriel-murakami:
I had a problem with `InvalidURIError` for accidentally creating a feature with whitespace between the words of the feature name using the Flipper UI. For example "feature _name". The error is due to the problem of accessing a URL with whitespaces like: `.../features/feature _name`.

The `strip` method didn’t fix this because it only removes whitespaces from the "edges":
```rb
"   feature _name      ".strip
=> "feature _name"
```

So I switched to a `gsub` to remove all whitespaces in the feature name:
```rb
"   feature _name    ".gsub(/\s/, '')
=> "feature_name"
```

### DIFF
--- a/lib/flipper/ui/actions/features.rb
+++ b/lib/flipper/ui/actions/features.rb
@@ -46,7 +46,7 @@ module Flipper
             halt view_response(:feature_creation_disabled)
           end
 
-          value = params['value'].to_s.strip
+          value = params['value'].to_s.gsub(/\s/, '')
 
           if Util.blank?(value)
             error = Rack::Utils.escape("#{value.inspect} is not a valid feature name.")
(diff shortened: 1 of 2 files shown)


jnunemaker:
Interesting. I think we need to fix the invalid uri error instead of the space. Technically, spaces are allowed, though I would definitely frown on them and cry. 😀 

I thought I had test cases for spaces. Was the problem on a GET request or were you enabling/disabling?

haxxxton:
would it be worth considering a slugize solution instead of just replacing spaces? Something like:
https://github.com/kwdowicz/slugize

gabriel-murakami:
> Interesting. I think we need to fix the invalid uri error instead of the space. Technically, spaces are allowed, though I would definitely frown on them and cry. 😀
> 
> I thought I had test cases for spaces. Was the problem on a GET request or were you enabling/disabling?

More specifically accessing the route through the browser. When I access with a nonexistent feature name it goes to his creation, but when I access a route with a feature name with space, even if it exists I have the URI problem

PR #500, opened 2021-03-10
Onboarding improvements

bkeepers:
This is a WIP for some onboarding improvements that I talked about with @jnunemaker.  The goal is to remove as much friction in the onboarding process as possible by clarifying some of the documentation and setting sensible defaults.

Some of these I'll probably work on in separate PRs, but here's what's on my mind so far:

- [x] Include high-level overview of features
- [x] Start with simple examples
- [x] Default to `Memory` adapter when using `require "flipper"` #501 
- [x] Adapters should register themselves where possible so `require "flipper-adapter-name"` is all that is needed. #502 
- [x] flipper-active-record: Define default implementation of `#flipper_id` on `AR::Base` #505 
- [x] Create `flipper-rails` (or maybe just add a `Railtie` to existing gems) to register middleware/etc #506 
- [ ] flipper-cloud: warn instead of raising errors if ENV vars are not defined

PR #502, opened 2021-03-10
Configure adapters on require

bkeepers:
For most use cases, a user should just be able to require an adapter and have it configure itself with sensible defaults. They can still override the default configuration, but this will make it just a little easer to get started.

- [x] active_record
- [x] mongo
- [x] redis
- [x] sequel

These are going to be harder because they need to wrap another adapter. Ideally the would just get the current default, and replace it with a wrapped adapter:

- [x] ~~dalli~~
- [x] ~~active_support_cache_store~~
- [x] ~~cloud~~

Edit: Skipping cache adapters for now.

And overall:

- [x] Tests for initialization

This is related to #500 for improving the onboarding experience.


jnunemaker:
This is exciting! 

I think I'd need to see a few examples to be able to decide on API for this. 

Like maybe what each of these wold look like:

* single adapter 
* single adapter with cache
* single adapter with multiple levels of cache
* single adapter with read only 
* cloud with same permutations as above

Random thoughts: 

* I don't think I want to use `use` as I do like the idea of allowing for rack-like middleware for adapters at some point -- something that could provide pre and post check operations without needing to build a fully fledged adapter.
* I'm fine with changing namespaces as long as we are backwards compatible for a while.
* I don't like `Store` but don't have a better suggestion. `Storage` feels kind of long but more accurate. `Store` and `Cache` are the same length which I like. For some reason `Store` makes me think store like target or walmart or somewhere you'd buy something more than where you'd _store_ something.
* If people use more than one level of caching, they could just use `default`. My only concern is that could be confusing to require two cache adapters and have the last require _win_. We could make cache build on top of each other (aka each cache adapter is configured automatically and added to caches array), but people would need to require in the correct order then. 
* `config.default` really just becomes the dsl instance. Makes me wonder about `config.dsl { |adapter| ... }`. Or something named better than that.

bkeepers:
@jnunemaker I messed around with a few different incantations of changing how the configuration works, and I don't really like any of them. 

So I'm thinking I want to let it simmer for a while, skip default configuration for the cache adapters for now, and maybe make another attempt later. They're advanced usage anyway, so I don't think it's a problem to require manual configuration.

This is ready for your review.

docs/active_record/README.md
```diff
@@ -23,47 +23,45 @@ Or install it yourself with:
...
+Note that the active record adapter requires the database tables to be created in order to work; failure to run the migration first will cause an exception to be raised when attempting to initialize the active record adapter.
+
+Flipper will be configured to use the ActiveReocrd adapter when `flipper-active_record` is loaded. But if you want to customize the adapter, you can add this to an initializer:
 
 ```ruby
-require 'flipper/adapters/active_record'
-adapter = Flipper::Adapters::ActiveRecord.new
-flipper = Flipper.new(adapter)
-# profit...
+# require 'flipper/adapters/active_record'
```
jnunemaker, line 35: Should this be commented out?

jnunemaker (approved):
Looks great. Thanks for updating the docs too. The initializers being commented out was the only question I had. Was the intent to show not to do it by default and that you'd need to uncomment for it to work?

bkeepers:
> Was the intent to show not to do it by default and that you'd need to uncomment for it to work?

Yes, that was what I was thinking. Thoughts on how to better communicate that?

jnunemaker:
Maybe just bold the "if you want to customize" part and/or put a comment at the top of the block that says if you want to customize?

jnunemaker:
Or a comment that says uncomment this either in the code snippet or before it. I'd like to think it is obvious that you need to uncomment, but you never know. Haha

PR #505, opened 2021-03-18
Add Flipper::Identifier with default flipper_id method

bkeepers:
This is a minor change related to #500 for improving the onboarding experience, but it removes one more small prerequisite for using Flipper: defining `#flipper_id`.  This adds `Flipper::Identifier` module that can be included into any class to define a default implementation, and it is included into `ActiveRecord::Base` and `Sequel::Model` by default.


jnunemaker (approved):
YES.

PR #506, opened 2021-03-18
Add Railties/Engines

bkeepers:
This adds Railties/Engines to remove as much manual configuration as possible when using flipping in a Rails app.

- [x] `require 'flipper'` sets up middleware
- [x] `require 'flipper-cloud'` configures and mounts app
- [x] ~~`require 'flipper-ui` configures and mounts app~~ skipping since it needs some kind of auth
- [x] Doc all the things

This is part of #500 to improve the onboarding experience.


spec/flipper/cloud/engine_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+  end
+
+  describe "routes" do
+    context "sync_method = :webhook" do
+      before { application.config.flipper.cloud.sync_method = :webhook }
+
+      it 'mounts cloud app' do
+        application.initialize!
+        expect(application.routes.recognize_path("/_flipper")).to be(anything)
```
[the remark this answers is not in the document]
bkeepers: Ok, removed the specs for now.

lib/flipper/cloud/routes.rb
```diff
@@ -0,0 +1,16 @@
+# Default routes loaded by Flipper::Cloud::Engine
+Rails.application.routes.draw do
+  config = Flipper::Cloud::Engine.config
+
+  if config.flipper.cloud.sync_method == :webhook
+    # FIXME: does it make sense to provide a Rails-specific config to fetch the instance? Or what should this be?
+    flipper = Flipper.instance
```
bkeepers, line 7: FIXME

lib/flipper/railtie.rb
```diff
@@ -0,0 +1,21 @@
...
+    config.before_configuration do
+      config.flipper = ActiveSupport::OrderedOptions.new
+      config.flipper.memoizer = ActiveSupport::OrderedOptions.new
+      config.flipper.memoizer.preload = true
+    end
+
+    initializer "flipper.memoizer" do |app|
+      if config.flipper.memoizer
+        app.middleware.use Flipper::Middleware::Memoizer, config.flipper.memoizer
+      end
```
bkeepers, line 12: @jnunemaker I'd love your thoughts on configuration.

We already have `Flipper::Configuration`, and this PR will introduce `Rails.application.config.flipper`. I'm tempted to combine them, but I can't decide which direction to go. I'm not sure it makes sense to add `memoizer` and `cloud` to `Flipper::Configuration` since they won't get used outside of Rails.

Would it make sense to deprecate `Flipper::Configuration`?

```ruby
  config.flipper.default = -> { Flipper.new(…) }
end
```

Outside of Rails, you'd be on your own for managing configuration and you would set `Flipper.instance` manually.

🤔

jnunemaker: My preference is to have everything in flipper and rails use flipper stuff. I don't like tying to rails anymore than is convenient. That is just my quick gut reaction. I can read this more today or tomorrow and see if I feel the same.

bkeepers: Ok, I'll take a pass at moving everything to Flipper and see how it feels.

docs/Optimization.md
```diff
@@ -1,52 +1,63 @@
...
+### Skip memoization
 
-# This assumes you setup a default flipper instance using configure.
-Rails.configuration.middleware.use Flipper::Middleware::Memoizer
+Prevent preloading and memoization by setting `unless` to a proc that evaluates to true.
+
+```ruby
+# config/initializers/flipper.rb
+Flipper.configure do |config|
+  config.memoize_unless = ->(request) { request.path.start_with?("/assets") }
```
bkeepers, line 31: Would it make more sense to just sent `config.memoize` to a block, and the result 


```ruby
Flipper.configure do |config|
  # Memoize all requests (default)
  config.memoize = true 

  # Disable memoization
  config.memoize = false 

  # Memoize if not an asset request
  config.memoize = ->(request) { !request.path.start_with?("/assets") }
end
```

This feels cleaner to me than having both `memoize` (boolean) and `memoize_unless` (block)

lib/flipper/middleware/memoizer.rb
```diff
@@ -53,9 +63,9 @@ def memoized_call(env)
         original = flipper.memoizing?
         flipper.memoize = true
 
-        flipper.preload_all if @opts[:preload_all]
-
-        if (preload = @opts[:preload])
+        if @opts[:preload] === true
```
jnunemaker, line 66: TIL. I didn't even know triple equals was a thing.

lib/flipper/railtie.rb
```diff
@@ -0,0 +1,21 @@
+module Flipper
+  class Railtie < Rails::Railtie
+    initializer "flipper.memoizer" do |app|
+      config = Flipper.configuration
+
+      if config.memoize
+        app.middleware.use Flipper::Middleware::Memoizer, {
+          env_key: config.env_key,
+          preload: config.preload,
+          if: config.memoize.is_a?(Proc) ? config.memoize : nil
```
jnunemaker, line 10: Should we just check that it responds to `call`? That would allow:

```ruby
class Blah
  def self.call(request)
    # true or false
  end
end

Flipper.configure do |config|
  config.memoize = Blah
end
```

jnunemaker (approved):
I think this looks great. Solid work. 

Prior to a new gem version release we'll need to test this with different configurations options in a real rails app just to double check that things work. I'd be happy to help with that. 

For some reason I'm having a hard time picturing how Flipper.configure in an initializer would get things setup soon enough for the rails engine to pickup the changes. But I don't know about load order for stuff like that. Have you tried it in a rails app with different memoize options and stuff?

docs/Optimization.md
```diff
@@ -1,52 +1,63 @@
...
-# setup default instance (perhaps in config/initializers/flipper.rb)
+# config/initializers/flipper.rb
 Flipper.configure do |config|
-  config.default do
-    Flipper.new(...)
-  end
+  # Load specific features that are used on most requests
+  config.preload = [:stats, :search, :some_feature]
+
+  # Disable preloading
```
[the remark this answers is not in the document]
bkeepers: Good call. I went with your second suggestion.

lib/flipper/railtie.rb
bkeepers: Great idea. Fixed!

bkeepers:
> For some reason I'm having a hard time picturing how Flipper.configure in an initializer would get things setup soon enough for the rails engine to pickup the changes. But I don't know about load order for stuff like that. Have you tried it in a rails app with different memoize options and stuff?

Thanks for bringing this up. I thought based these [docs in rails/application.rb](https://github.com/rails/rails/blob/84643885cf722b29bb5c0fef7bc3678d1117258d/railties/lib/rails/application.rb#L56) that the Railtie initializers would run _after_ `config/initializers/*`:


> The application is also responsible for setting up and executing the booting
> process. From the moment you require "config/application.rb" in your app,
> the booting process goes like this:
>
>   1)  require "config/boot.rb" to set up load paths
>   2)  require railties and engines
>   3)  Define Rails.application as "class MyApp::Application < Rails::Application"
>   4)  Run config.before_configuration callbacks
>   5)  Load config/environments/ENV.rb
>   6)  Run config.before_initialize callbacks
>   7)  Run Railtie#initializer defined by railties, engines and application.
>       One by one, each engine sets up its load paths, routes and runs its config/initializers/* files.
>   8)  **Custom Railtie#initializers added by railties, engines and applications are executed**
>   9)  Build the middleware stack and run to_prepare callbacks
>   10) Run config.before_eager_load and eager_load! if eager_load is true
>   11) Run config.after_initialize callbacks

However, I tested it in a real app and that was not the case. So I've updated the Railtie to explicitly run after `config/initializers/*`, and I've updated the cloud engine to mount the route after initialize.

bkeepers:
@jnunemaker I tested this in a real app and had to rework configuration a little. Before 2bf2e82, `Cloud::Engine` would instantiate `Flipper.instance` to decide if the webhooks app should be mounted. This would cause it to attempt an HTTP request to sync on app boot, which is not ideal because it not only slows down the boot process, but could prevent the app from booting if an error occurred.

So as a result of cleaning some of that up, I ended up going full circle on the discussion about configuration and moved a few things back to `Rails.application.config`. These are only configuration that relate to this Rails integration (`env_key`, `memoize`, `preload`, and `cloud_path`). Configuring Flipper adapters still happens in `Flipper.configuration`. This feels better to me since they really only apply for the Rails integration.

jnunemaker:
This AMAZING. Very excited. 

I don't want to block shipping this, but some other things that would be nice to wrap up in rails config eventually:

* enabling instrumentation with ActiveSupport::Notifications for cloud and default. I usually use these. Probably good on by default and config to disable ala memoize.
* log flipper calls in rails log `require "flipper/instrumentation/log_subscriber" if ENV.fetch("FLIPPER_LOG", "0").to_i == 1`. This is usually something that I protect with env var but it could be on in dev or on based on config. It would need to set instrumented for dsl instance as well.

I should be done kicking the tires today. Tested a couple apps. I just want to check all the config with the [flipper cloud default rails app](https://github.com/fewerandfaster/flipper-rails-demo) to ensure things are working.

bkeepers:
ooh, good call. 7040006c sets `instrumentor` in both the `Railtie` and `Cloud::Engine`. `LogSubscriber` already checks if `logger.debug?`, so seems to me like it's fine to just require in the railtie without any configuration. If someone _really_ wanted to change or disable it, I think they could set `Flipper::Instrumentation::LogSubscriber.logger = Logger.new(…)` .

PR #509, opened 2021-04-09
Skip adapter tests if connection fails

bkeepers:
My soul can't handle having failing tests locally. This skips the tests that depend on local services if they fail to connect.

### DIFF
--- a/spec/flipper/adapters/active_support_cache_store_spec.rb
+++ b/spec/flipper/adapters/active_support_cache_store_spec.rb
@@ -9,7 +9,7 @@ RSpec.describe Flipper::Adapters::ActiveSupportCacheStore do
   let(:memory_adapter) do
     Flipper::Adapters::OperationLogger.new(Flipper::Adapters::Memory.new)
   end
-  let(:cache) { ActiveSupport::Cache::DalliStore.new(ENV['MEMCACHED_URL']) }
+  let(:cache) { ActiveSupport::Cache::MemoryStore.new }
   let(:adapter) { described_class.new(memory_adapter, cache, expires_in: 10.seconds) }
   let(:flipper) { Flipper.new(adapter) }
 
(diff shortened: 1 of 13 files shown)


jnunemaker:
What would happen in CI if it couldn't connect? Should local behavior be skip and CI raise?

spec/flipper/adapters/active_support_cache_store_spec.rb
```diff
@@ -9,7 +9,7 @@
   let(:memory_adapter) do
     Flipper::Adapters::OperationLogger.new(Flipper::Adapters::Memory.new)
   end
-  let(:cache) { ActiveSupport::Cache::DalliStore.new(ENV['MEMCACHED_URL']) }
+  let(:cache) { ActiveSupport::Cache::MemoryStore.new }
```
bkeepers, line 12: DalliStore is deprecated (https://github.com/petergoldstein/dalli/issues/573) in favor of MemCacheStore, but I swapped it with `MemoryStore` instead to avoid the dependency on memcached.

spec/flipper/middleware/memoizer_spec.rb
```diff
@@ -342,7 +342,7 @@
     it 'eagerly caches known features for duration of request' do
       memory = Flipper::Adapters::Memory.new
       logged_memory = Flipper::Adapters::OperationLogger.new(memory)
-      cache = ActiveSupport::Cache::DalliStore.new(ENV['MEMCACHED_URL'])
+      cache = ActiveSupport::Cache::MemoryStore.new
```
bkeepers, line 345: The adapter being used here doesn't seem to matter, and since DalliStore is deprecated, I just swapped it with MemoryStory.

bkeepers:
> What would happen in CI if it couldn't connect? Should local behavior be skip and CI raise?

Good point. Added `unless ENV['CI']` to all the skips

spec/flipper/adapters/active_support_cache_store_spec.rb
jnunemaker: 👍

test/adapters/mongo_test.rb
```diff
@@ -9,13 +9,15 @@ def setup
     port = '27017'
     logger = Logger.new('/dev/null')
     client = Mongo::Client.new(["#{host}:#{port}"],
-                               server_selection_timeout: 1,
+                               server_selection_timeout: 0.01,
```
jnunemaker, line 12: 👍 🙇

PR #511, opened 2021-04-13
Deprecate superflous sync_method setting

bkeepers:
`sync_secret` always needs to be set if this is changed, so I don't see a need for both settings.

This change was extracted from #506 to try to reduce the scope of that PR.


jnunemaker (approved):
This would be a good one to add as a note in the changelog, but we'll have to update the changelog for next release anyway since there are several changes. I'll probably just go back through PRs and update it.

PR #512, opened 2021-04-14
Add `write_through` to ActiveSupport cache adapter

smcgivern:
The current approach with caching adapters does something like this:

1. On write, delete the entry from the cache.
2. On the next read, repopulate the cache.

Unfortunately this can be subject to subtle race conditions if the
underlying storage (in our case, a replicated database) gives a stale
read for item 2 - which may be in a different process / thread / host
entirely. More details on our specific case are available at
https://gitlab.com/gitlab-org/gitlab/-/issues/325452.

The new `write_through` option to the ActiveSupportCacheStore adapter's
initialiser allows the user to opt in to write-through caching, where
the new value is immediately written to the cache in step 1.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,7 @@
 
 * Allow actors and time gates to deal with decimal percentages (https://github.com/jnunemaker/flipper/pull/492)
 * Change Flipper::Cloud::Middleware to receive webhooks at / in addition to /webhooks.
+* Add `write_through` option to ActiveSupportCacheStore adapter to support write-through caching (https://github.com/jnunemaker/flipper/pull/512)
 
 ## 0.20.3
 
--- a/lib/flipper/adapters/active_support_cache_store.rb
+++ b/lib/flipper/adapters/active_support_cache_store.rb
@@ -25,12 +25,13 @@ module Flipper
       attr_reader :name
 
       # Public
-      def initialize(adapter, cache, expires_in: nil)
+      def initialize(adapter, cache, expires_in: nil, write_through: false)
         @adapter = adapter
         @name = :active_support_cache_store
         @cache = cache
         @write_options = {}
         @write_options[:expires_in] = expires_in if expires_in
+        @write_through = write_through
       end
 
       # Public
@@ -49,7 +50,13 @@ module Flipper
       def remove(feature)
         result = @adapter.remove(feature)
         @cache.delete(FeaturesKey)
-        @cache.delete(key_for(feature.key))
+
+        if @write_through
+          @cache.write(key_for(feature.key), {}, @write_options)
+        else
+          @cache.delete(key_for(feature.key))
+        end
+
         result
       end
 
@@ -88,14 +95,26 @@ module Flipper
       ## Public
       def enable(feature, gate, thing)
         result = @adapter.enable(feature, gate, thing)
-        @cache.delete(key_for(feature.key))
+
+        if @write_through
+          @cache.write(key_for(feature.key), @adapter.get(feature), @write_options)
+        else
+          @cache.delete(key_for(feature.key))
+        end
+
         result
       end
 
       ## Public
       def disable(feature, gate, thing)
         result = @adapter.disable(feature, gate, thing)
-        @cache.delete(key_for(feature.key))
+
+        if @write_through
+          @cache.write(key_for(feature.key), @adapter.get(feature), @write_options)
+        else
+          @cache.delete(key_for(feature.key))
+        end
+
         result
       end
 
--- a/lib/flipper/spec/shared_adapter_specs.rb
+++ b/lib/flipper/spec/shared_adapter_specs.rb
@@ -205,7 +205,7 @@ RSpec.shared_examples_for 'a flipper adapter' do
 
     expect(subject.remove(feature)).to eq(true)
 
-    expect(subject.get(feature)).to eq(subject.default_config)
+    expect(subject.get(feature).values).to all(be_blank)
   end
 
   it 'can clear all the gate values for a feature' do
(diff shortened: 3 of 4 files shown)


lib/flipper/spec/shared_adapter_specs.rb
```diff
@@ -205,7 +205,7 @@
 
     expect(subject.remove(feature)).to eq(true)
 
-    expect(subject.get(feature)).to eq(subject.default_config)
+    expect(subject.get(feature).values).to all(be_blank)
```
smcgivern, line 208: This is because the value with this change can be:

```
{:actors=>#<Set: {}>, :boolean=>nil, :groups=>#<Set: {}>, :percentage_of_actors=>nil, :percentage_of_time=>nil}
```

And that is because if we delete, but a stale reader is reading old data, the stale reader might decide to write the old data. If we insert an empty hash on delete then we guarantee that can't happen.

lib/flipper/adapters/active_support_cache_store.rb
```diff
@@ -88,14 +88,14 @@ def get_all
       ## Public
       def enable(feature, gate, thing)
         result = @adapter.enable(feature, gate, thing)
-        @cache.delete(key_for(feature.key))
+        @cache.write(key_for(feature.key), @adapter.get(feature), @write_options)
```
smcgivern, line 91: This does mean we have to go back to the underlying adapter immediately to read the new value, which might not be ideal.

jnunemaker:
Ugh! Subtle problems like that are so hard to track down too. 😄 

Write through caching is definitely something to be careful with. I'm not sure how I feel about this. 

It will probably be fine for most people, so I'm inclined to merge. But my worry is expectations. I don't believe I would expect a caching adapter to do write through. I'd expect read through. Changing this would change peoples "mental model without looking at the code". 

## Option 1

At GitHub, we had replicas as well. We also had mechanisms for forcing reads from the primary instead of a replica. I believe that is what we did for flipper feature data reads -- always read from primaries. 

Is that something that you could do? I'd suspect the # of reads to be very low since caching would handle 99% or more of reads. 

## Option 2

Actually, another option that just occurred to me would be to have an adapter option. Perhaps the adapter could have `write_through: true` or something as an option. It would default to false to retain expectation, but allow tweaking for those that want it. 

Any thoughts on this option if reading from primaries is a complete no go for you? 

```ruby
Flipper::Adapters::ActiveSupport.new(adapter, cache, write_through: true)

# in adapter code
if options[:write_through]
  @cache.write(...)
else
  @cache.delete(...)
end
```

smcgivern:
Thanks for your thoughts @jnunemaker, that's very helpful!

> At GitHub, we had replicas as well. We also had mechanisms for forcing reads from the primary instead of a replica. I believe that is what we did for flipper feature data reads -- always read from primaries.
>
> Is that something that you could do? I'd suspect the # of reads to be very low since caching would handle 99% or more of reads.

We could, and I think that's totally reasonable. The reason I picked this approach (changing the caching layer) in our codebase is largely due to the flip side of the last sentence I quoted there. While a very high proportion of reads are handled by caching, a very high proportion of reads don't strictly _need_ the primary. It's only those that are racing with a write happening at the same time. In practice, this isn't likely to add much additional load to our primaries, but to me it feels better (for what that's worth) to change the caching layer.

> Actually, another option that just occurred to me would be to have an adapter option. Perhaps the adapter could have `write_through: true` or something as an option. It would default to false to retain expectation, but allow tweaking for those that want it.

This would work fine for us :+1: Equally fine for us is you rejecting this change entirely, though: the interface of these adapters appears to be very stable, so us having our own overrides for some of these methods is not a big headache for us maintenance-wise.

Obviously if that _does_ work for you, I'm happy to go ahead and add it. The closer we can stay to upstream the better. I just don't want you to feel forced into adding an option you don't really want just because of our use-case 😃

jnunemaker:
I don't feel forced. Happy to accept a write through option. That keeps it the same for everyone but allows those that need it to be different. 

If you are willing to add it, I'll happily accept. Also no need to add to any adapters you aren't using. You can just add it where you need it and go back to work as normal. If you have any problems or would prefer I do it I can try to whip something together soonish.

Changelog.md
```diff
@@ -4,6 +4,7 @@
 
 * Allow actors and time gates to deal with decimal percentages (https://github.com/jnunemaker/flipper/pull/492)
 * Change Flipper::Cloud::Middleware to receive webhooks at / in addition to /webhooks.
+* Add `write_through` option to ActiveSupportCacheStore adapter to support write-through caching (https://github.com/jnunemaker/flipper/pull/512)
```
smcgivern, line 7: Not sure on the etiquette here, I can remove this if you prefer these to happen outside PRs.

spec/flipper/adapters/active_support_cache_store_spec.rb
```diff
@@ -22,11 +23,25 @@
...
       adapter.remove(feature)
-      expect(cache.read(described_class.key_for(feature))).to be(nil)
+    end
+
+    it 'expires feature by deleting the cache' do
+      expect(cache.read(described_class.key_for(feature))).to be_nil
+      expect(feature).not_to be_enabled
+    end
+
+    context 'with write-through caching' do
```
smcgivern, line 38: Would you like me to add similar tests for `#enable` and `#disable`?

smcgivern:
@jnunemaker I've updated this PR now, could you take a look please?

spec/flipper/adapters/active_support_cache_store_spec.rb
smcgivern: Sure thing, I've done that now.

Changelog.md
jnunemaker: That's fine! You are an exemplary contributor. :)

jnunemaker:
I'll do my best to look at this tonight. On my phone and with both kids or I'd do it now. Kids win. :)

Appreciate the work.

PR #515, opened 2021-04-21
docs: fix Flipper::Identifier.flipper_id example

choznerol:
The [default implementation][1] of `Flipper::Identifier` uses a semicolon (instead of a colon) to separate class and id.

[1]: https://github.com/jnunemaker/flipper/blob/f4a50467a6da84f4d886b8bec1df56127446002b/lib/flipper/identifier.rb#L14


jnunemaker:
Thanks for noticing!

PR #520, opened 2021-05-04
Spread cache keys when using `Redis#mget`

ianks:
The [signature of `Redis#mget` actually requires a splatted array of keys to fetch](https://www.rubydoc.info/gems/redis/3.1.0/Redis%3Amget). Currently, the implementation works fine for the default `redis` gem, but the current behavior (non-spreading) does not work with `redis-store`. This patch should fix that.

### DIFF
--- a/lib/flipper/adapters/redis_cache.rb
+++ b/lib/flipper/adapters/redis_cache.rb
@@ -148,7 +148,7 @@ module Flipper
         return [] if keys.empty?
 
         cache_keys = keys.map { |key| key_for(key) }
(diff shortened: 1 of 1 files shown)


bkeepers (approved):
Thanks @ianks! Confirmed in the docs for [Redis#mget](https://www.rubydoc.info/gems/redis/3.1.0/Redis%3Amget) that it does indeed take a splat.

bkeepers:
Tests are just failing because the examples require a secret to run.

PR #521, opened 2021-05-06
Fix small typo in README.md in installing flipper-active_record part.

philiplambok:
I found a typo when trying to install the gem, especially when installing the flipper with an active record adapter. So this PR is just to fix the typo :)


jnunemaker:
@philiplambok thanks!

Also there is going to be a new release out very soon that reflects these docs. The latest gem and the docs here don't line up currently. Just an FYI. 

PR #523, opened 2021-05-07
Only perform memoization and preloading once

bkeepers:
When upgrading to the next release of flipper, anyone using the Memoizer middleware will need to make some changes. Specifically, they will need to either remove the mounting of the middleware and update the `preload` config, if they are using it…

```diff
- Rails.application.middleware.use Flipper::Middleware::Memoizer, preload: [:some, :features]

+ Rails.application.configure do
+  config.flipper.preload = [:some, :features]
+ end
```

…or disable memoizing from the Railtie:

```diff
+ Rails.application.configure do
+  config.flipper.memoize = false
+ end
```

This pull request updates the Memoizer middleware to be a noop if `flipper.memoizing?` is already true.  This should allow for the first instance of the middleware (manually mounted in `config/initializers/*`) to continue working, and the second one (added by the Railtie after `config/initializers/*`) to be a noop.

### DIFF
--- a/lib/flipper/middleware/memoizer.rb
+++ b/lib/flipper/middleware/memoizer.rb
@@ -61,9 +61,11 @@ module Flipper
         reset_on_body_close = false
         flipper = env.fetch(@env_key) { Flipper }
 
-        # Already memoizing. Maybe the memization middleware is mounted twice?
-        # This instance does not need to do anything.
-        return @app.call(env) if flipper.memoizing?
+        # Already memoizing. This instance does not need to do anything.
+        if flipper.memoizing?
+          warn "Flipper::Middleware::Memoizer appears to be running twice. Read how to resolve this at https://github.com/jnunemaker/flipper/pull/523"
+          return @app.call(env)
+        end
 
         flipper.memoize = true
 
--- a/lib/flipper/middleware/memoizer.rb
+++ b/lib/flipper/middleware/memoizer.rb
@@ -60,7 +60,11 @@ module Flipper
       def memoized_call(env)
         reset_on_body_close = false
         flipper = env.fetch(@env_key) { Flipper }
-        original = flipper.memoizing?
+
+        # Already memoizing. Maybe the memization middleware is mounted twice?
+        # This instance does not need to do anything.
+        return @app.call(env) if flipper.memoizing?
+
         flipper.memoize = true
 
         case @opts[:preload]
@@ -70,12 +74,12 @@ module Flipper
 
         response = @app.call(env)
         response[2] = Rack::BodyProxy.new(response[2]) do
-          flipper.memoize = original
+          flipper.memoize = false
         end
         reset_on_body_close = true
         response
       ensure
-        flipper.memoize = original if flipper && !reset_on_body_close
+        flipper.memoize = false if flipper && !reset_on_body_close
       end
     end
   end
--- a/spec/flipper/middleware/memoizer_spec.rb
+++ b/spec/flipper/middleware/memoizer_spec.rb
@@ -239,13 +239,18 @@ RSpec.describe Flipper::Middleware::Memoizer do
       end.to_app
     end
 
+    def get(uri, params = {}, env = {}, &block)
+      silence { super(uri, params, env, &block) }
+    end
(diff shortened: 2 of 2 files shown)


lib/flipper/middleware/memoizer.rb
```diff
@@ -60,7 +60,11 @@ def memoize?(request)
       def memoized_call(env)
         reset_on_body_close = false
         flipper = env.fetch(@env_key) { Flipper }
-        original = flipper.memoizing?
+
+        # Already memoizing. Maybe the memization middleware is mounted twice?
+        # This instance does not need to do anything.
+        return @app.call(env) if flipper.memoizing?
```
bkeepers, line 66: Should this warn or do something to log that it's skipping memoization?

jnunemaker: I like the idea of warning.

jnunemaker (approved):

Linuus:
I don't fully understand how to mitigate this.

Before, we had this initializer:

```ruby
Flipper.configure do |config|
  config.default do
    adapter = Flipper::Adapters::ActiveRecord.new

    Flipper.new(adapter)
  end
end

Rails.configuration.middleware.use(Flipper::Middleware::Memoizer)
```

Now, I have removed all this code, since I believe it should be the default(?), and now I get the warning `Flipper::Middleware::Memoizer appears to be running twice. Read how to resolve this at https://github.com/jnunemaker/flipper/pull/523`.

How am I supposed to configure it now? I want to preload all and use memoization 🤔

rbclark:
This error is showing up for me on an upgrade from Rails 6.1 to Rails 7 even though there is no instances of `config.flipper` or `Flipper::Middleware::Memoizer` in the application. The application is just pulling in the `flipper`, `flipper-ui`, and `flipper-active_record` gems. This error was not showing up previously when the application was running Rails 6.1 with the same configuration.

jnunemaker:
Can you share what versions of flipper, flipper-ui and flipper-active_record you are using?

rbclark:
Sure no problem:

```
    flipper (0.23.1)
    flipper-active_record (0.23.0)
    flipper-ui (0.23.0)
```

jnunemaker:
@Linuus sorry I didn't respond to you. I've been not getting some emails via github, not sure why. 

What version of rails, flipper and flipper-active_record are you using?

rbclark:
@jnunemaker Sorry the versions I posted last were from the wrong copy of the app. I have one version running on Rails 6.1 with the versions I posted above, the version that is acting up on Rails 7 is running `0.23.1` across the board:

```
    flipper (0.23.1)
    flipper-active_record (0.23.1)
    flipper-ui (0.23.1)
```

With the mixed up versions things are working correctly although I'm not sure if that's due to Rails 6.1 or due to the mixed up versions. I'll have to do some more testing to track that down.

coreyaus:
In case it's helpful, we're seeing the same issue with the latest gem release:
```
flipper (0.23.1)
flipper-active_record (0.23.1)
```

I've posted the full notes on what we're seeing in the following new issue in case that's helpful (though obviously feel free to close if not!): https://github.com/jnunemaker/flipper/issues/604

coreyaus:
Do the changes in this PR mean some of the docs need to be updated, e.g. the Optimization page: https://www.flippercloud.io/docs/optimization#advanced, which includes this example code block:

```ruby
use Flipper::Middleware::Memoizer,
  preload: true,
  unless: ->(request) { request.path.start_with?("/assets") }
```

bkeepers:
@coreyaus the docs should be up to date. is there something wrong with that example? It is for advanced usage, but should not be needed for a normal Rails app.

MrChrisW:
@bkeepers Not to drag up an only thread, but I think @coreyaus is pointing out that the [advanced examples](https://www.flippercloud.io/docs/optimization#advanced) of using `Flipper::Middleware::Memoizer` now output the warning added in this PR. 

Should the advanced documentation include the need to disable memoizing from the railtie (`config.flipper.memoize = false`) if manually mounting `Flipper::Middleware::Memoizer` to prevent this warning?

PR #525, opened 2021-05-08
Enable log subscriber by default in Rails

bkeepers:
Requires `flipper/instrumentation/log_subscriber` in the Railtie unless logging is explicitly disabled. The log subscriber already does a check for `logger.debug?`, so I think it makes sense to require this regardless of environment.

### DIFF
--- a/lib/flipper/railtie.rb
+++ b/lib/flipper/railtie.rb
@@ -5,7 +5,8 @@ module Flipper
         env_key: "flipper",
         memoize: true,
         preload: true,
-        instrumenter: ActiveSupport::Notifications
+        instrumenter: ActiveSupport::Notifications,
+        logging: true
       )
     end
 
@@ -27,6 +28,10 @@ module Flipper
           if: config.memoize.respond_to?(:call) ? config.memoize : nil
         }
       end
+    end
+
+    initializer "flipper.logging", after: :load_config_initializers do |app|
+      require "flipper/instrumentation/log_subscriber" if app.config.flipper.logging
     end
 
     initializer "flipper.identifier" do
--- a/lib/flipper/railtie.rb
+++ b/lib/flipper/railtie.rb
@@ -31,7 +31,10 @@ module Flipper
     end
 
     initializer "flipper.log", after: :load_config_initializers do |app|
-      require "flipper/instrumentation/log_subscriber" if app.config.flipper.log
+      config = app.config.flipper
+      if config.log && config.instrumenter == ActiveSupport::Notifications
+        require "flipper/instrumentation/log_subscriber"
+      end
     end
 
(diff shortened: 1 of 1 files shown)


lib/flipper/railtie.rb
```diff
@@ -5,7 +5,8 @@ class Railtie < Rails::Railtie
         env_key: "flipper",
         memoize: true,
         preload: true,
-        instrumenter: ActiveSupport::Notifications
+        instrumenter: ActiveSupport::Notifications,
+        logging: true
```
jnunemaker, line 9: Truly minor thing: memoize, preload, log maybe are more consistent. Or memoizing, preloading, logging, instrumenting. Perhaps `log: true` or `logger: true` would be more consistent? 

I also think I'd default to `false`. Then people can easily turn it on if they like. 

Last thought: this would only work if instrumenter is set to ActiveSupport::Notifications so I wonder if those settings need to be tied together somehow. You have any preference or thoughts on that?

bkeepers: > Truly minor thing: memoize, preload, log maybe are more consistent. Or memoizing, preloading, logging, instrumenting. Perhaps `log: true` or `logger: true` would be more consistent?

hah, I had the same argument in my head. My eyes liked that `logging` is the same length as all the other options, but my frontal lobe thought that was a pretty ridiculous reason to chose a name for something. Apparently my eyes won that argument.

`logger` makes it seem like you can set the logger instance. I guess we could support that too.

> I also think I'd default to `false`. Then people can easily turn it on if they like.

I personally really like seeing flipper in the logs. I get that it can get noisy in apps with a ton of flags, but in smaller apps, it's really nice.

> Last thought: this would only work if instrumenter is set to ActiveSupport::Notifications so I wonder if those settings need to be tied together somehow. You have any preference or thoughts on that?

Ah ok, we can add a check to make sure instrumenter it's AS::Notifications

jnunemaker: > hah, I had the same argument in my head.

great minds...

> I personally really like seeing flipper in the logs.

I can go either way so you can make the call.

lib/flipper/railtie.rb
```diff
@@ -30,8 +30,11 @@ class Railtie < Rails::Railtie
       end
     end
 
-    initializer "flipper.logging", after: :load_config_initializers do |app|
-      require "flipper/instrumentation/log_subscriber" if app.config.flipper.logging
+    initializer "flipper.log", after: :load_config_initializers do |app|
+      config = app.config.flipper
+      if config.log && config.instrumenter == ActiveSupport::Notifications
```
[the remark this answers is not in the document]
bkeepers: Changing instrumenter seems very rare, so I can go either way.

Issue #526, opened 2021-05-10
Memoizer warning


morgoth:
After upgrading to flipper 0.21.0, I'm getting a warning in the logs when visiting flipper UI.
`Flipper::Middleware::Memoizer appears to be running twice. Read how to resolve this at https://github.com/jnunemaker/flipper/pull/523`
On the regular app requests, it looks fine (there is no warning).

I don't have any custom configuration of this middleware.

jnunemaker:
Ah! Good catch. Thanks for reporting. 

@bkeepers we can remove the memoizer middleware from the UI and API apps. But if we do we lose the block functionality. 

@bkeepers Trying to decide if I should just add functionality to pass block or instance to UI and API or just remove block/instance and warn... any preferences?

bkeepers:
Doh, sorry, I didn't even think about this side effect.

> @bkeepers Trying to decide if I should just add functionality to pass block or instance to UI and API or just remove block/instance and warn... any preferences?

What are the use cases where people would use it, instead of just relying on the default instance? Either way, it seems very rare, so I would be in favor of removing block/instance, or at least moving it to a named argument instead of the first argument.

jnunemaker:
> Doh, sorry, I didn't even think about this side effect.

Me either. 

> What are the use cases where people would use it, instead of just relying on the default instance?

I guess people who haven't switched to the default instance. I actually have no concept of how many have switched verse how many are still using the old way of just creating a new dsl instance. 

So so so wish I could get automatic analytics on stuff like this in a non creepy way. 

> Either way, it seems very rare, so I would be in favor of removing block/instance, or at least moving it to a named argument instead of the first argument.

K.

jnunemaker:
This is fixed in the unreleased version of flipper. I'll probably cut a release soon.

PR #527, opened 2021-05-10
Remove memoizer from UI

jnunemaker:
Default rails and flipper setups will use memoizer so having ui include it causes the double memoize warning message, which can be confusing.

This removes the memoizer from UI and adds the backwards compatibility stuff we need to make it all work the same.

We can add a doc on how to add your own memoizer for just the ui:

```ruby
run Flipper::UI.app { |builder|
  builder.use Flipper::Middleware::SetupEnv, some_flipper, env_key: "some_flipper"
  builder.use Flipper::Middleware::Memoizer, env_key: "some_flipper"
  # etc
}
```

We can also add a doc on how to add your own memoizer for just the api: 

```ruby
run Flipper::Api.app { |builder|
  builder.use Flipper::Middleware::SetupEnv, api_flipper, env_key: "api_flipper"
  builder.use Flipper::Middleware::Memoizer, env_key: "api_flipper"
  # etc
}
```

I'm open to doing something different. Just figured I'd throw out a first example to react to.

href #526

### DIFF
--- a/lib/flipper/api/middleware.rb
+++ b/lib/flipper/api/middleware.rb
@@ -9,15 +9,9 @@ end
 module Flipper
   module Api
     class Middleware
-      def initialize(app, flipper = nil, options = {})
+      def initialize(app, options = {})
         @app = app
         @env_key = options.fetch(:env_key, 'flipper')
-
-        if flipper.respond_to?(:call)
-          @flipper_block = flipper
-        else
-          @flipper = flipper || Flipper
-        end
 
         @action_collection = ActionCollection.new
         @action_collection.add Api::V1::Actions::PercentageOfTimeGate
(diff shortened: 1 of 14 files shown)


lib/flipper/api/middleware.rb
```diff
@@ -9,10 +9,16 @@
...
-      def initialize(app, options = {})
+      def initialize(app, flipper = nil, options = {})
         @app = app
         @env_key = options.fetch(:env_key, 'flipper')
 
+        if flipper.respond_to?(:call)
+          @flipper_block = flipper
+        else
+          @flipper = flipper || Flipper
+        end
```
bkeepers, line 20: Per discussion in #526, I would be in favor of either passing this in `options`, or removing it entirely and relying on `SetupEnv` in all the middleware/rack apps.

jnunemaker:
@bkeepers updated. Is this more like what you were picturing?

PR #536, opened 2021-06-15
fix(active-record): arelize get_all query

### LINKED ISSUE #535: Incorrect syntax near the keyword 'key' due to SqlServer reserved word use in get_all query
https://github.com/jnunemaker/flipper/blob/d52f7bb88feca6f0463ed655c31c8b482dadad14/lib/flipper/adapters/active_record.rb#L103-L109

The above usage of `key` and `value` appears to be causing a SQL query syntax error in 0.21.0 - specifically 0.21.0.rc2. This query appears to have been there for quite some time, but something in 0.21.0.rc2 caused the break. `key` and `value` are both reserved words in SQL Server. `value` is also a reserved word in MySQL, so curious if it's an issue there as well.

We're using the active record adapter with an MS SqlServer database in a Rails 5.0 app. We use the activerecord-sqlserver-adapter gem.

We do have a module that adjusts our model `table_name`s depending on db technology, which we include in the Feature and Gate Flipper AR classes. It doesn't muck with Flipper, only models' `table_name` attributes. That's the extent of our customization. The rest of the Flipper config/setup is very basic.

```ruby
class Flipper::Adapters::ActiveRecord::Feature
  include WbCommon::AdminData
end

class Flipper::Adapters::ActiveRecord::Gate
  include WbCommon::AdminData
end
```

gradybarrett:
Arel and Sequel know how to handle reserved words in terms of escape characters for the given underlying database technology. This way we don't have to assume that DB-specific knowledge.

Closes #535

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ gem 'minitest-documentation'
 gem 'webmock', '~> 3.0'
 gem 'climate_control'
 gem 'redis-namespace'
+gem 'webrick'
 
 group(:guard) do
   gem 'guard', '~> 2.15'
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -101,11 +101,13 @@ module Flipper
       end
 
       def get_all
-        rows = ::ActiveRecord::Base.connection.select_all <<-SQL.tr("\n", ' ')
-          SELECT ff.key AS feature_key, fg.key, fg.value
-          FROM #{@feature_class.table_name} ff
-          LEFT JOIN #{@gate_class.table_name} fg ON ff.key = fg.feature_key
-        SQL
+        features = ::Arel::Table.new(@feature_class.table_name.to_sym)
+        gates = ::Arel::Table.new(@gate_class.table_name.to_sym)
+
+        rows_query = features.join(gates, Arel::Nodes::OuterJoin).on(features[:key].eq(gates[:feature_key]))
+                             .project(features[:key].as('feature_key'), gates[:key], gates[:value])
+        rows = ::ActiveRecord::Base.connection.select_all rows_query
+
         db_gates = rows.map { |row| Gate.new(row) }
         grouped_db_gates = db_gates.group_by(&:feature_key)
         result = Hash.new { |hash, key| hash[key] = default_config }
--- a/lib/flipper/adapters/sequel.rb
+++ b/lib/flipper/adapters/sequel.rb
@@ -97,11 +97,17 @@ module Flipper
       end
 
       def get_all
-        db_gates = @gate_class.fetch(<<-SQL).to_a
-          SELECT ff.key AS feature_key, fg.key, fg.value
-          FROM #{@feature_class.table_name} ff
-          LEFT JOIN #{@gate_class.table_name} fg ON ff.key = fg.feature_key
-        SQL
+        # TODO: make this more succinct
+        features_sql =
+          @feature_class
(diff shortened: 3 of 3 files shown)


gradybarrett:
@jnunemaker I think this is ready for you to take a look at. I left some detail in the commit for the Sequel change. Happy to make any changes.

jnunemaker:
@gradybarrett you rock! 🤘 Thank you so much for fixing this and I'm sure making many people's lives better.

jnunemaker:
@gradybarrett oh and I almost forgot. I've been giving www.flippercloud.io credits to people who contribute, so if you want to check it out beyond the trial, let me know and I'll spot you several months or more. 

gradybarrett:
My pleasure. Glad I could help!

Cool, thank you! I'll let you know. At some point, we'll be re-evaluating feature flagging software and flipper cloud is in the running. So far we've only used open-source, and to be honest, we haven't really taken advantage of the feature set offered. We use Flipper in 3 Rails apps and really only flip features fully on or off. We do this primarily to facilitate testing and help limit the need for long-lived feature branches through silent releases. It's been great for those use cases. We haven't had any issues that I can recall, setup was quite straightforward, and the framework is sufficiently extensible for the limited customization that we have.

I think the only thing that might knock Flipper out of the running is how the enterprise feature set stacks up against competitors - primarily service integration w/ Rollbar and CI/CD tools as well as polyglot client support.

As we modernize our CI/CD tools we open the door for more complex workflows, automated gradual feature rollouts, A/B testing, etc. All that and more is where we'd like to go eventually.

Hope that helps, and thanks for this awesome library. It'll likely continue to be my go-to lib for Ruby projects in the future.

jnunemaker:
> So far we've only used open-source, and to be honest, we haven't really taken advantage of the feature set offered. We use Flipper in 3 Rails apps and really only flip features fully on or off. We do this primarily to facilitate testing and help limit the need for long-lived feature branches through silent releases.

This is exactly what taking advantage of flipper to the fullest looks like. Its not about using all the gates. Its about getting code into production faster so you can kick the tires without causing damage. Love it.

> primarily service integration w/ Rollbar

I saw LD or someone had an integration like this. What would that even mean or do? I definitely use error tracking (currently https://honeybadger.io) but I'm not sure how they would integrate together. I'd love to hear more. 

> CI/CD tools

I'm also curious what else would be helpful here. I usually use the memory adapter for flipper to keep things snappy in test mode. I should probably write up a post on that as I think I've said it to a few people now. 

> polyglot client support

What languages? This is definitely on our radar.

gradybarrett:
> I saw LD or someone had an integration like this. What would that even mean or do? I definitely use error tracking (currently https://honeybadger.io) but I'm not sure how they would integrate together. I'd love to hear more.

> CI/CD tools

The biggest use case I was thinking of was when we do an A/B or beta feature release using flags, we could have LD disable a feature if we start seeing certain errors being reported. Or we could integrate w/ CI/CD tools to slowly roll a feature out canary-style. There might be more that I'm not thinking of atm.

> I'm also curious what else would be helpful here. I usually use the memory adapter for flipper to keep things snappy in test mode. I should probably write up a post on that as I think I've said it to a few people now.

Nice! That's a great suggestion.

> What languages? This is definitely on our radar.

For us, it's Java, Javascript, and Python - in that order in terms of usage although Javascript variants are quickly gaining on Java.

jnunemaker:
Ah! That makes sense. Interesting idea. Thanks for explaining.

PR #547, opened 2021-07-30
Remove Octicons and replace with a pure CSS status circle 

onesneakymofo:
For me, when I switched from the Rails asset pipeline over to Webpacker, the squirrel icons turned into little squares. I didn't want to have to download Octicons or create some hacky solution so I figured I would create a pull request to remove the icons altogether.

This pull request replaces the squirrels with a simple circle in the CSS and drops the trashcan icon in favor of the word Delete to match the other buttons on the page. It also saves 180kb most of which wasn't used.

Screenshots:
2021-07-30 at 2 01 AM

2021-07-30 at 2 02 AM


thetimbanks:
I would be fine with making this change, but would like @jnunemaker to chime in just to make sure there isn't more to having the squirrel icons.

jnunemaker:
I'll miss them but I'm fine with reducing dependencies. 👍

jnunemaker:
I haven't checked this on mobile but I'm assuming it's fine. I'll merge it soon if @thetimbanks doesnt.

jnunemaker:
I'd maybe go with remove instead of delete. I usually think add and remove or create and delete. Minor and I'd be curious if anyone else has any opinions on that.

onesneakymofo:
@jnunemaker Yep, contextually that makes a lot more sense. I've made the change.

PR #556, opened 2021-08-25
Feature: Passing list of new actors

ivan-garanin:
Ability to pass multiple actors in 1 request. 
Sometimes our product team asks us to turn on some feature for specific users. We just select them using SQL by some condition or our analytics/marketing provides this list of us. Right now nonprogramming guys or developers without prod access are forced to add each user separated. This PR could solve their pain when they need to handle a big list of users.


ivan-garanin:
added `Flipper::UI.configuration.actors_list_separator_symbol`. Default separtor is comma `,`, but you can change in config  using  `Flipper::UI.configuration.actors_list_separator_symbol`

PR #560, opened 2021-09-16
Fix remove button for groups

dgmora:
It seems that when the octicons were removed this was left over. The remove button isn't visible at the moment (it's clickable even if it's invisible)

WhatsApp Image 2021-09-16 at 11 38 57


thetimbanks (approved):

jnunemaker:
Thanks!

Issue #90, opened 2015-11-19
Mongoid support


dmitrypol:
Thank you for creating such a valuable gem.  I see that there is a support for Mongo generic Ruby driver.  Is there a way to make Flipper work with Mongoid?  Especially 4x version since it still uses the previous Moped driver vs Mongoid 5 (which uses Mongo Ruby driver)

jnunemaker:
Yes, there definitely is. You can easily create an adapter of your own for any data store. The docs are here: https://github.com/jnunemaker/flipper/blob/9ea8a7df5aad222bd3332f09c082071a1898c835/docs/Adapters.md. I'm not interested in personally supporting a mongoid adapter, but you can create your own gem or even just an adapter for only your application. I'd be happy to link to it if you do make one as a gem.

phstc:
Just for the reference for those using (newer) versions of Mongoid that uses `mongo`:

```ruby
Flipper.configure do |config|
  config.default do
    collection = Mongoid.default_client['flipper']
    adapter = Flipper::Adapters::Mongo.new(collection)
    Flipper.new(adapter)
  end
end
```

derekclee:
I ended up creating a mongoid model.

```
class FlipperGate
  include Mongoid::Document

  store_in collection: "flipper"

  def self.flipper
    adapter = Flipper::Adapters::Mongo.new(collection)
    Flipper.new(adapter)
  end
end
```

jclusso:
> Just for the reference for those using `mongoid >= 5.0.0`
> 
> ```ruby
> Flipper.configure do |config|
>   config.default do
>     collection = Mongoid.default_client['flipper']
>     adapter = Flipper::Adapters::Mongo.new(collection)
>     Flipper.new(adapter)
>   end
> end
> ```

This would be great to have in the readme @jnunemaker.

PR #561, opened 2021-09-30
Added Mongoid example

jclusso:
I think this example will save others the time I wasted messing around. Eventually @phstc's example in #90.


bkeepers:
@jclusso thanks for the doc update! I made a little tweak to use `#adapter` instead of `#default`, but otherwise looks great!

PR #563, opened 2021-10-06
Fix railtie initialization to avoid altering middleware order

### LINKED ISSUE #562: Upgrading Flipper to any version >= 0.21.0 Leads to Railtie/Middleware Ordering Issues
So I've been going down a bit of a rabbit hole trying to upgrade some apps that utilize flipper to the most recent version (0.22.1  at time of writing) These are apps that used very old versions of flipper (from 0.14 to 0.17). (Note: This is all being done in Rails 5) 

However, after doing this upgrade, we ran into new errors with our specs. While some of these were fixed by disabling preloading as detailed in the changelog, there were other errors that we could not immediately diagnose. 

After further digging, we found that the upgrade to version 0.21.0 of Flipper led to the order of Railtie initialized middleware being changed. This led to errors in situations where the app expected one middleware to be before another. 

For example, pre-version 0.21 the middleware stack looked like:

```
use Rack::Sendfile
use ActionDispatch::Static
use Rack::Lock
use ActionDispatch::Executor
use ActiveSupport::Cache::Strategy::LocalCache::Middleware
use Rack::Runtime
use Rack::MethodOverride
use ActionDispatch::RequestId
use ActionDispatch::RemoteIp
use Rails::Rack::Logger
use ActionDispatch::ShowExceptions
use MetaRequest::Middlewares::Headers
use ActionDispatch::DebugExceptions
use ActionDispatch::Callbacks
use ActionDispatch::Cookies
use ActionDispatch::Session::CookieStore
use ActionDispatch::Flash
use ActionDispatch::ContentSecurityPolicy::Middleware
use Rack::Head
use Rack::ConditionalGet
use Rack::ETag
use Rack::TempfileReaper
use CustomMiddleware1
use MetaRequest::Middlewares::MetaRequestHandler
use MetaRequest::Middlewares::AppRequestHandler
use Pinglish
use Pinglish
use CustomMiddleware2
use Warden::Manager
``` 

CustomMiddleware2 depends on CustomMiddleware1 running first to work as intended. (A bit of a nono since it can lead to issues like we're seeing now, but I digress). 

After upgrading to a version of flipper >= 0.21, the middleware stack now looks like this: 

```
... Same middleware as above down through TempfileReaper 
...
...
use Pinglish
use Pinglish
use CustomMiddleware2
use Warden::Manager
use Flipper::Middleware::Memoizer
use CustomMiddleware1
use MetaRequest::Middlewares::MetaRequestHandler
use MetaRequest::Middlewares::AppRequestHandler
```

The only thing that is changing is the upgrading of the flipper gem, but as can be seen, the entire order of this middleware has been changed. Most importantly CustomMiddleware2 has been moved to be above CustomMiddleware1. 

I've done some digging through Flipper's Railtie implementation, but could not discern what could be causing this change in middleware ordering. But this issue persists whether memoizing is enabled/disabled ( the presence of the memoizer middleware does not change the ordering issue at hand). 

We can proceed with the upgrade by manually reordering the Railties, but I decided to post this issue to document some of this digging and to see if there was any insight as to what could be causing this problem.

bkeepers:
The railtie was using `after: : load_config_initializers` to ensure app initializers were run before mounting the middleware. As described in #562, this appears to cause the middleware stack to be initialized prematurely and changing the order of middleware. That option is not needed here and was only added to be explicit because we have another initializer that is `before: :load_config_initializers`. Looking closer at the [rails booting process](https://api.rubyonrails.org/classes/Rails/Application.html#class-Rails::Application-label-Booting+process), initializers from railties run after `config/initializers`.

Fixes #562

### DIFF
--- a/lib/flipper/railtie.rb
+++ b/lib/flipper/railtie.rb
@@ -18,7 +18,7 @@ module Flipper
       end
     end
 
-    initializer "flipper.memoizer", after: :load_config_initializers do |app|
+    initializer "flipper.memoizer" do |app|
       config = app.config.flipper
 
       if config.memoize
@@ -30,7 +30,7 @@ module Flipper
       end
     end
 
-    initializer "flipper.log", after: :load_config_initializers do |app|
+    initializer "flipper.log" do |app|
       config = app.config.flipper
       if config.log && config.instrumenter == ActiveSupport::Notifications
         require "flipper/instrumentation/log_subscriber"


jdelStrother:
For what it's worth, this does mean that the [suggestion](https://github.com/jnunemaker/flipper/blob/master/Changelog.md#upgrading) to configure preloading by adding something like

```ruby
  config.flipper.preload = [:foo]
end
```

in config/initializers/flipper.rb is now ignored. You'd have to put it in, eg, config/application.rb or config/environments/production.rb for it to get picked up early enough.

PR #564, opened 2021-10-08
railtie spec

dpep:
Hey, I'm a big user of Flipper and was looking through the new Flipper middleware.  I saw a few opportunities to simplify this spec. Open to any of this?

### DIFF
--- a/spec/flipper/railtie_spec.rb
+++ b/spec/flipper/railtie_spec.rb
@@ -4,29 +4,23 @@ require 'flipper/railtie'
 
 RSpec.describe Flipper::Railtie do
   let(:application) do
-    app = Class.new(Rails::Application).new(
+    Class.new(Rails::Application).new(
       railties: [Flipper::Railtie],
-      ordered_railties: [Flipper::Railtie]
-    )
-    app.config.eager_load = false
-    app.config.logger = ActiveSupport::Logger.new($stdout)
-    app.run_load_hooks!
+    ).tap do |app|
+      app.config.eager_load = false
+      app.run_load_hooks!
+    end
   end
 
-  before do
-    Rails.application = nil
-  end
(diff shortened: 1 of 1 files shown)


jnunemaker:
@bkeepers hey when you get a sec can you look at this? @dpep said it should be good. Just did a call with him. Just want any gut feelings you have.

spec/flipper/railtie_spec.rb
```diff
@@ -3,30 +3,29 @@
...
-  let(:application) do
-    app = Class.new(Rails::Application).new(
-      railties: [Flipper::Railtie],
-      ordered_railties: [Flipper::Railtie]
-    )
-    app.config.eager_load = false
-    app.config.logger = ActiveSupport::Logger.new($stdout)
-    app.run_load_hooks!
+  let(:config) do
+    Rails.application.config.flipper
```
bkeepers, line 7: I don't think we run the tests multi-threaded yet, but if we do switch to that (which I'd like to do eventually to make sure we don't have any threading issues), this could fail by relying on the `Rails.application` singleton. What would you think about assigning `let(:application)`, in the before block setting `Rails.application = application`, and having `config` delegate to `application` instead of `Rails.application`?

bkeepers (commented):
Hey @dpep, sorry I missed the initial ping on this.

Thanks for taking time to work on this. I remember writing this test and feeling like I was performing a magic trick to get the railtie set up right for testing. My only concern is relying on `Rails.application`, but I don't feel strongly about it if that's the only way to do it. We can deal with any threading issues later.

spec/flipper/railtie_spec.rb
dpep: further simplified, given your suggestions

PR #570, opened 2021-10-26
feat: allow HTML in banner and feature descriptions

irphilli:
Hi!

I was looking for a way to format the banner / descriptions a little bit more, but it's a bit tricky since Flipper UI is Rack middleware. Would you be open to something like this where HTML is allowed to render in the ERBs?

There _shouldn't_ be any user-inputted text here since this would all come from config values, but also happy to make it an option (possibly that just changes the `escape` option passed to Erubi) if that's preferred.

Thanks!

### DIFF
--- a/lib/flipper/ui/views/feature.erb
+++ b/lib/flipper/ui/views/feature.erb
@@ -9,7 +9,7 @@
     <div class="col">
       <div class="card">
         <div class="card-body">
-          <%== @feature.description %>
+          <%== Sanitize.fragment(@feature.description, Sanitize::Config::BASIC) %>
         </div>
       </div>
     </div>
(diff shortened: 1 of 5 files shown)


irphilli:
@jnunemaker That's more than reasonable! Will update this once get the chance - thanks!

lib/flipper/ui/views/feature.erb
```diff
@@ -9,7 +9,7 @@
     <div class="col">
       <div class="card">
         <div class="card-body">
-          <%= @feature.description %>
+          <%== Sanitize.fragment(@feature.description, Sanitize::Config::BASIC) %>
```
irphilli, line 12: Seems like [Sanitize::Config::BASIC](https://github.com/rgrove/sanitize#sanitizeconfigbasic) is sensible. Can potentially make this an options, but this seems good enough for now.

irphilli:
Updated the PR to use sanitize!

jnunemaker:
@irphilli Thanks!

PR #572, opened 2021-11-03
feat(api): include configuration option to omit feature gate data

irphilli:
Hi again!

Had a scenario where I only needed feature state data returned by the API, and it was preferable to strip out the additional gate data (i.e. prevent exposing actor IDs).

It's easy enough to monkey-patch, but figured someone else may want simplified data returned as well, so here's a PR to make it an option!

Thanks!


jnunemaker:
Any thoughts on making this a param in the url (say ?exclude_gates=true or something) instead of a config option?

irphilli:
> Any thoughts on making this a param in the url (say ?exclude_gates=true or something) instead of a config option?

Awesome idea - that makes it much more flexible 👍 . Will do!

irphilli:
Updated this since docs have been moved out the repo. Not sure how to update docs now, but this should be good to go as far as code goes.

PR #576, opened 2021-11-25
tap-it-good

dpep:
I discovered `.tap` a while back and it was a game changer.  shall we simplify?

https://apidock.com/ruby/Object/tap

### DIFF
--- a/lib/flipper/adapters/dual_write.rb
+++ b/lib/flipper/adapters/dual_write.rb
@@ -34,33 +34,27 @@ module Flipper
       end
 
       def add(feature)
-        result = @remote.add(feature)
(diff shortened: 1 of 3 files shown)


bkeepers (approved):
I'm a sucker for one-liners

jnunemaker:
I intentionally didn't do this because there is no way a block yield is faster than a variable assignment (old school and lame I know). But honestly perf at this spot probably isn't a big deal so I'll merge it. 😄

dpep:
yeah, performance _must_ be worse with a block.  this optimizes for dev efficiency - both better readability and less prone to inadvertently changing the return value

PR #578, opened 2021-11-25
pstore thread_safe fix

dpep:
not sure the thread_safe attribute is working.  should we save the value, extract it out of @store via `.instance_variable_get`, or delete the accessor since it may not be needed?  (I also sorted the variables to keep things tidy)

### DIFF
--- a/lib/flipper/adapters/pstore.rb
+++ b/lib/flipper/adapters/pstore.rb
@@ -22,9 +22,10 @@ module Flipper
 
       # Public
       def initialize(path = 'flipper.pstore', thread_safe = false)
+        @name = :pstore
         @path = path
         @store = ::PStore.new(path, thread_safe)
-        @name = :pstore
+        @thread_safe = thread_safe
       end
 
       # Public: The set of known features.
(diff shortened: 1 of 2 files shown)


bkeepers (changes requested):
It doesn't look like the attribute reader is used, and I can't think of a scenario when someone would even need to check if it's thread safe or not. I am in favor of just removing it.

I also think the default value should be `true`. [The docs](https://ruby-doc.org/stdlib-3.0.2/libdoc/pstore/rdoc/PStore.html) say there is a minor performance impact, flipper is not write heavy enough to worry about it, and defaulting to `false` could lead to obscure bugs in production.

cc https://github.com/jnunemaker/flipper/pull/334

jnunemaker:
@bkeepers fixed your stuff in 4048704fefe41b716015294a19a0b94546637630

dpep:
yeah, I agree that removing the attribute and making pstore thread safe by default are good moves.  it makes Flipper more resilient by default.  we may want to update the changelog since these are technically breaking changes and there's a slight chance of causing trouble for folks downstream

jnunemaker:
Yep I added a note here: https://github.com/jnunemaker/flipper/blob/6ee7d6309e4d2f30a4380f0a4ef41ffab9fba37a/Changelog.md#L9-L9 That look good?

dpep:
👍🏻   we may also want to add language around the removal of the `.thread_safe` attribute reader in case folks happen to be using it.  eg https://github.com/jnunemaker/flipper/pull/582

PR #579, opened 2021-11-25
synchronizer now

dpep:
it seems like `Synchronizer.now` is only needed for testing.  can we de-duplicate this with the instance method?


jnunemaker (approved):
I dont remember why I did it that way. I'm good with your changes.

PR #581, opened 2021-11-28
noop spec cleanup

dpep:
use rspec's `expect {}.to yield_control` syntax to ensure block is being called

https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/yield-matchers


jnunemaker (approved):

jnunemaker:
I love changes like this. Thanks @dpep!

PR #582, opened 2021-11-28
changelog update

dpep:
improve messaging around breaking change


jnunemaker (approved):
Yep, this is fine, though before it didn't work so I doubt anyone is using it.

PR #587, opened 2021-12-04
Improve http error message

jnunemaker:
The API returns nice error messages but our errors weren't showing them.

Now we are showing them. Much easier to figure out what went wrong and fix it on your own.


jnunemaker:
@jonewalt this should improve the error you hit regarding actor limit. Instead of just saying 422, it now says the exact thing that caused the validation error.

PR #589, opened 2021-12-16
Support ActiveRecord 7.0

nhocki:
Rails 7.0 was just released and having the dependency `< 7` prevents us from installing flipper with a new rails app.

I'm not sure _what_ needs to change or why the restriction on `< 7` so opening the easiest PR for this, but let me know if there are things that worry you and I'll work on them!

One thing that might hold this back is that Rails 7 requires Ruby 2.7.0+ and I see tests with ruby 2.5 and 2.6.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         ruby: ['2.5', '2.6', '2.7']
-        rails: ['5.2', '6.0.0', '6.1.0']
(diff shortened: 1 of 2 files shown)


jnunemaker:
I'm close on this. I excluded 2.5 and 2.6 for rails 7. I also updated active support to < 8 in one spot. Now just seeing why a couple specs are failing about modifying frozen array.

jnunemaker:
Your changes here were merged as part of another branch. Thanks again!

PR #603, opened 2022-01-25
Fix deprecated uses of `Redis#pipelined`

casperisfine:
Context: https://github.com/redis/redis-rb/pull/1059

The following is deprecated
```ruby
redis.pipelined do
  redis.get(key)
end
```

And should be rewritten as:
```ruby
redis.pipelined do |pipeline|
  pipeline.get(key)
end
```

Functionally it makes no difference.
This API is available since Redis 3.0.

### DIFF
--- a/lib/flipper/adapters/redis.rb
+++ b/lib/flipper/adapters/redis.rb
@@ -125,14 +125,14 @@ module Flipper
       # Private: Gets a hash of fields => values for the given feature.
       #
       # Returns a Hash of fields => values.
-      def doc_for(feature)
-        @client.hgetall(feature.key)
+      def doc_for(feature, pipeline: @client)
+        pipeline.hgetall(feature.key)
       end
 
       def docs_for(features)
-        @client.pipelined do
+        @client.pipelined do |pipeline|
           features.each do |feature|
-            doc_for(feature)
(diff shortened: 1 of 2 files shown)


lib/flipper/adapters/redis.rb
```diff
@@ -125,14 +125,14 @@ def read_feature_keys
       # Private: Gets a hash of fields => values for the given feature.
       #
       # Returns a Hash of fields => values.
-      def doc_for(feature)
+      def doc_for(feature, _pipeline: @client)
         @client.hgetall(feature.key)
```
bkeepers, line 129: Should this use `_pipeline`?

casperisfine: 🤦 , yes absolutely...

casperisfine: Fixed. Also since I noticed the doc mark the method as private, I droped the leading `_`.

fabn:
Is there an ETA for this to be released? I hate to put github deps in Gem file. Thanks.

jnunemaker:
Personally, I don't mind locking to a ref for a bit, but I just released 0.24.0 for you. :)

fabn:
Thank you very much

PR #606, opened 2022-02-08
Call to_app before returning the rack app

matheussilvasantos:
If we return a `Rack::Builder` instance, it will call `to_app` for every request. That's not performant, but more importantly, it makes apps that want to use `Rack::Session::Pool` not work. `Rack::Session::Pool` uses `@pool` to store the sessions; if we rebuild the Rack application for every request, it can't keep track of sessions because it resets the variable.

The following application wouldn't work without this change:

```ruby
require "rack"
require "flipper"
require "flipper-ui"

flipper = Flipper.new(Flipper::Adapters::Memory.new)

run Rack::URLMap.new({
  "/flipper" => Flipper::UI.app(flipper) do |builder|
    builder.use Rack::Session::Pool
  end
})
```


jnunemaker:
Thanks! Sorry it took me a few days to run this through my apps to ensure it works everywhere. :)

matheussilvasantos:
No worries, thanks for the merging!

PR #612, opened 2022-03-13
feat(dsl): make it possible to disable internal memoization

### LINKED ISSUE #611: Custom Memoizable adapter
First of all, I'd like to thank you for such a good open-source project.

# Context
I have 2 applications, one is user-facing and is the one that defines the feature flags, and the other is for developers to update these feature flags state.

image

The thing is that **Application1** uses [Event Machine](https://github.com/eventmachine/eventmachine) to handle user requests, which makes it a single-thread application and because of that it ends up with a single instance of Flipper for the time that the application lives.

This is a problem in this scenario because `Flipper` uses `Flipper::Adapters::Memoizable` as its internal `memoization` strategy and the way it works makes the in-memory cache not be updated in the **Application1** runtime.

> A simple solution is to just create a new `Flipper` instance every time we change the Event Machine context to handle a different request, but it also would not work for us because we also have an in-memory cache strategy where we define a TTL in order to reduce the pressure on Redis.

# Suggestion
Allow `Flipper` to be created passing a custom `Memoizable` adapter:

```ruby
Flipper.configure do |config|
  config.default do
    Flipper.new(redis_adapter, options, memoize_adapter: my_custom_memoize_adapter)
  end
end
```

douglasgabriel:
Closes #611

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,7 @@
 * Add (alpha) client side instrumentation of events to cloud (https://github.com/jnunemaker/flipper/pull/602)
 * Fix deprecated uses of Redis#pipelined (https://github.com/jnunemaker/flipper/pull/603). redis-rb >= 3 now required.
 * Fix Flipper UI Rack application when `Rack::Session::Pool` is used to build it (https://github.com/jnunemaker/flipper/pull/606).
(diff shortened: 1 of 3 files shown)


jnunemaker:
Thanks! Sorry this slipped past me last week.

PR #616, opened 2022-03-23
Add Flipper::Actor#hash

jdelStrother:
This allows them to be used as hash-keys - eg

```ruby
features = {
  Flipper::Actor.new("abc") => [:foo, :bar],
  Flipper::Actor.new("def") => [:baz],
}
features[Flipper::Actor.new("abc")] #=> [:foo, :bar]
```


jnunemaker:
Awesome. Thanks! 

Issue #557, opened 2021-09-01
Expressions


jnunemaker:
I'll try to communicate more soon. But the tldr is @orderedlist, @bkeepers and I are working on a new feature (for www.flippercloud.io and open source) we're calling "rules" ([branch](https://github.com/jnunemaker/flipper/compare/learn-the-rules)). This issue is just so we can keep track of the work together on the open source side.

## ActiveRecord Migration

Those using the `flipper-active_record` adapter will want to migrate the database so it can store JSON expressions:

```
$ rails generate migration change_flipper_gates_value_to_text
```

```ruby
  def up
    change_column :flipper_gates, :value, :text
  end

  def down
    change_column :flipper_gates, :value, :string
  end
```

## TODO 

* [x] Support single `Rule`
* [x] Support multiple rules through `And` rule
* [x] Support multiple rules through `Any` rule
* [x] Refactor `from_hash` and other case statements
* [x] Get all specs and tests passing
* [x] Add enable_* and disable_* methods to `Flipper`, `Flipper::DSL` and `Flipper::Feature`
* [x] Add Rule equality like `==` and `eql?`
* [x] Make real ruby classes for types in left, operator and right
* [x] Add examples for `Rule`, `Any`, and `And`
* [x] Make all types consistent in capitalization (currently Condition and property, string, etc.)
* [x] Add rule shortcuts to make it easier to use them
* [x] Add new shared adapter spec(s) to shared adapter tests
* [x] Update feature synchronizer to include :json/rules
* [x] Update api docs
* [x] Update gate docs
* [x] Error in AR/Sequel if `value` column is not `text`
* [x] Add `Flipper.add_rule` and `Flipper.remove_rule` for easy adding and removing
* [x] What about floats? Should we use number instead of integer as the type and support floats? Or add float as type?  
* [x] Document AR/Sequel migration to convert `value` from `string` to `text`

## After releasing beta

* [ ] Enforce types for operators. in requires array on right. percentage requires string on left and integer/number on right, etc. Make this easy to pass to javascript front end for @orderedlist. 
* [ ] Enforce string, number, null, boolean, etc. or array of former for flipper_properties somehow (maybe skip invalid and warn to avoid blowing up? or blow up in dev but skip in prod?)
* [x] Ensure that time based enablements are easy (enable when now epoch > int or now > date time)
* [x] Make it possible to enable/disable rules in www.flippercloud.io
* [ ] Update readme to include rules

## Someday 

[...]

PR #618, opened 2022-03-30
Deny actor gate

### LINKED ISSUE #514: Allow explicitly disabling actors
This has come up a few times. It would be great if we could come up with a good way of explicitly enabling a feature with exceptions. For example, this feature is enabled for everyone but john. Or this feature is enabled for everyone but customers on the basic tier. 

The main problem currently is that flipper is about gates and letting people through the gates. Flipper.disable is about clearing those enablements, not explicitly disabling the gate for the actor which is different. 

https://github.com/jnunemaker/flipper/issues/82#issuecomment-759626499 talks more about this. Storage of disabled is easy but the interface changes could cause quite a ripple for existing flipper users so we'd want to be careful. 

* **enable** - enables feature for gate
* **un-enable** - not disable, just removes the enabled gate if it is enabled else nothing, if actor enabled then this would remove actor, if group is enabled, it would remove group, if percentage is set it would unset it
* **disable** - disables feature for gate, disable actor would store actor id in set of disabled actors, disable group would store group in set of disabled groups, flipper feature enabled? checks would likely need throw catch or some other mechanism to explicitly say "even though this feature is enabled for this gate, it is explicitly disabled for this actor or group or other gate and should return false 

## Workarounds 

* **Create a negative feature**. If feature is :redesign, you create new feature :redesign_disabled. Each code point needs to check `Flipper.enabled?(:redesign, actor) && !Flipper.enabled?(:redesign_disabled, actor)`. A bit painful but can be wrapped up in another application method (e.g. `Redesign.enabled?(actor)` which can check both and return value). **Con**: Involves checking two features. Also involves knowing what/how the negative feature works in this context.
* **Use group + application storage**. If you have group and group checks attribute you can enable the group and have that group check an attribute on the actor. :redesign_enabled group, checks if redesign_enabled? (some attribute stored on actor). If that is false it returns false for group. If that is true it returns true for group. Then you just need to avoid using other gates. **Con**: Would break if someone fully enabled. Also involves application storage instead of all baked into flipper.
* **Use application storage + pre-feature flag check**. Check something with application logic in addition to Flipper feature check. Can also do this via Redesign.enabled?(actor) or something. That can check `return false if actor.redesign_disabled?` or whatever. **Con**: Involves you storing stuff on your own instead of baked into flipper.

rafaeelaudibert:
This PR adds a new DeniedActor gate, responsible for handling the case where we want to forbid (deny!) access of an actor to the resource in question.

When calling `Flipper.deny_actor(actor)` we want to be able to deny access of an actor to a resource, overriding any permission this actor might have to see a feature. If we want to remove this, we can use the `Flipper#reinstate_actor` method.

This takes precedence over the already existing gates, so if you enable a feature to everyone (or even to a single actor), if an actor was denied, it won't have access to the feature.

Closes #514


rafaeelaudibert:
@jnunemaker This is an experimental PR, and I am open to hearing your thoughts about the approach I took.

I definitely think there should be better approaches, but I didn't want to break backward compatibility, so I kept every public-facing method with the same behavior as before (except the #enabled? one, for obvious reasons), and just added some new methods on top of the already existing DSL.

Also, I read #514 and decided to go with deny/reinstate, but this is clearly open to changes.

At last, I probably didn't write enough tests or documentation, but I think this is ok for the first review.

jnunemaker:
I'm excited to look through this but I'm on vacation this week. Just wanted you to know that's why I'll probably be slow to respond fully.

rafaeelaudibert:
@jnunemaker No problem! I took the time in my company's Hackathon Week to do this, so I'm actually not 100% available as well, feel free to take the time you need to look at it

bkeepers (commented):
Hey @rafaeelaudibert, thanks for the pull request! We've been talking for a long time about adding the ability to deny and it's something I really want.

The approach you've taken here works great for actors. I think we also need to support block/deny on boolean, groups, percentage of time, and percentage of actors. Basically, any gate should have the ability to halt.

We have talked internally about how Rules in #557 will make this easier to implement, but we don't currently have a timeline for launching rules.

I'm going to leave this open so we can keep discussing if it makes sense to go forward with adding deny support to existing gates, but just wanted to give you a heads up on our current thinking.

rafaeelaudibert:
@bkeepers Sorry for taking this long to get back to you!

I agree with you, and I think it'd be way easier to implement this with the proposed set of rules. I'll keep this PR open just for the sake of it, in case someone wants to take my work and bring it the last mile, but I will not be able to finish it up.

Looking forward to when we have rules!

We are using Flipper in production extensively, and it's been so helpful, thank you for your product :)

rafaeelaudibert:
I don't think this will be useful now that you've implemented expressions - great job!

I'll close this one :)

jnunemaker:
@rafaeelaudibert I'm torn on this one. I think it is still useful to deny individual actors. 

Maybe we build a mechanism into expressions to deny. 
Or adding a `not` expression (if we don't have one, sometimes I forget or don't realize what all we support). 
Or we add a deny expression and if that matches it never checks the allow/current.

PR #620, opened 2022-04-07
Add kwargs for method_missing using ruby 3.0

RobWu:
`kwargs` are not supported for ruby 3.0 for `method_missing`

`method_missing` is called in blocks of flippers groups when calling method of `actor`
(https://www.flippercloud.io/docs/features#enablement-group)


jnunemaker:
@RobWu thanks for caring!

PR #625, opened 2022-04-21
 Added a prompt in Flipper UI for the 'Delete' button to prevent accidental delete of features

ivorpad:
The confirm prompt was not working so I added a prompt to prevent accidental deletes and also deleted the iframe with the video that was linking to an invalid video URL.

### DIFF
--- a/lib/flipper/ui/views/features.erb
+++ b/lib/flipper/ui/views/features.erb
@@ -8,9 +8,6 @@
           <a class="btn btn-primary btn-sm" href="<%= script_name %>/features/new">Add Feature</a>
         </p>
       <%- end -%>
-      <div class="embed-responsive embed-responsive-16by9">
-        <iframe class="embed-responsive-item" width="560" height="315" src="https://www.youtube.com/embed/e-ORhEE9VVg" frameborder="0" allowfullscreen></iframe>
-      </div>
     <% else %>
       <h4>Getting Started</h4>
       <p class="mb-1">You have not added any features to configure yet.</p>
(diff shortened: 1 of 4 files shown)


lib/flipper/ui/views/features.erb
```diff
@@ -8,9 +8,6 @@
           <a class="btn btn-primary btn-sm" href="<%= script_name %>/features/new">Add Feature</a>
         </p>
       <%- end -%>
-      <div class="embed-responsive embed-responsive-16by9">
-        <iframe class="embed-responsive-item" width="560" height="315" src="https://www.youtube.com/embed/e-ORhEE9VVg" frameborder="0" allowfullscreen></iframe>
```
bkeepers, line 12: 😭 Goodbye Taylor. Looks like the video doesn't allow embedding anymore?

bkeepers (approved):
Looks good to me.

bkeepers:
Thanks @ivorpad, we'll get a release out soon.

lib/flipper/ui/views/features.erb
maxjacobson: It looks like it can be. Maybe the HTML changed.

```html
<iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/e-ORhEE9VVg?si=w0z-d34k2yc3_Cty" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" referrerpolicy="strict-origin-when-cross-origin" allowfullscreen></iframe>
```

PR #626, opened 2022-04-25
failsafe adapter

dpep:
Create an adapter that fails safe, ie. closed, when issues arise.  This protects against issues like network hiccups or Redis crashes from taking down an application using Flipper.

### DIFF
--- a/lib/flipper/adapters/failsafe.rb
+++ b/lib/flipper/adapters/failsafe.rb
@@ -1,0 +1,76 @@
+module Flipper
+  module Adapters
+    class Failsafe
+      include ::Flipper::Adapter
+
+      # Public: The name of the adapter.
+      attr_reader :name
+
+      # Public: Build a new Failsafe instance.
+      #
+      # adapter   - Flipper adapter to guard.
+      # options   - Hash of options:
+      #             :errors - Array of exception types for which to fail safe
+
+      def initialize(adapter, options = {})
+        @adapter = adapter
+        @errors = options.fetch(:errors, [StandardError])
+        @name = :failsafe
+      end
+
+      def features
+        @adapter.features
+      rescue *@errors
+        Set.new
+      end
+
+      def add(feature)
+        @adapter.add(feature)
+      rescue *@errors
+        false
+      end
+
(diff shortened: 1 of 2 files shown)


bkeepers (approved):
@dpep nice work! Looks good to me. I'll leave this open for a day or two to see if @jnunemaker has any feedback.

I could see it being nice to have a mechanism for reporting the errors that were rescued, but that can easily be a future addition.

dpep:
yeah, a reporting mechanism was top of mind but I ran out of time this weekend...it'll ship with v2 ;)  I'm thinking an optional block or kwarg?

jnunemaker (approved):
Looks good to me. 

The first thought I had for implementation, though more complex than here, would be to accept primary and fallback adapters. When the primary fails it uses the fallback. 

This would allow people to create chains of fallback steps. Additionally, the fallback adapter could be a memory adapter with a default of false but perhaps one or two features set to true. 

This allows for more flexibility without needing to program it, but also more documentation and thought for implementer as they have to think well i can fallback to an in memory adapter that is already populated.

Other thoughts I had (not blockers): 

* Should `{}` be returned for `get` or should the default values be returned?
* Should `{}` be returned for get_multi or should it be hash of keys => default values?

Things I'd do next (not blockers):

* allow setting the safe value to `true` instead of `false` and maybe even allow setting true/false safe value per feature by key

I'm good to merge this now, but if anything I said above inspires you feel free to roll with it!

dpep:
@jnunemaker : all good points.  Could we have achieved the same functionality with the [Failover adapater](https://github.com/jnunemaker/flipper/blob/master/lib/flipper/adapters/failover.rb) + memory adapter?

supporting default values, or failing open instead of closed could be interesting to explore too.  maybe setting the fallback values in a memory adapter and then using Failover is a better implementation option?

dpep:
upon further investigation: this adapter and Failover differ with regard to how they handle failures during write operations, eg. enable/disable.  Failsafe effectively swallows all errors and no-ops, whereas Failover fails for write events and falls back to a secondary during read failures.

PR #628, opened 2022-04-28
Allow specify a connection class on AR adapter

ceritium:
It fits better on Rails multidb setups.


bkeepers:
Here's what I'm suggesting: https://github.com/jnunemaker/flipper/compare/ar-base

Then you should be able to set the connection with:

```ruby
Flipper::Adapters::ActiveRecord::Model.connects_to database: { writing: :flipper, reading: :flipper_replica }

**Comment by ceritium** [comment]
Hi @bkeepers,

Yes, I had to subclass them. Your approach works even better, at least for my current scenario. Do you plan to merge it and release it soon? 

Instead of `connects_to`, I included a module with more behavior, but it's basically the same.

```ruby
Flipper::Adapters::ActiveRecord::Model.include(Global::Concerns::Connection)
```

I am going to close this PR.

PR #629, opened 2022-04-30
ActiveRecord: base class for internal models

bkeepers:
Alternative approach to #628 to ease connecting to alternate database. This defines an abstract base class for Flipper's internal ActiveRecord models, which can be used to configure connection handling and probably other options. It also switches from using `ActiveRecord::Base` to the internal `Feature` class (or the `feature_class` option passed when the adapter is initialized) to get a new connection.

This allows configuring connection options:

```ruby

Flipper::Adapters::ActiveRecord::Model.connects_to database: {
  writing: :flipper, reading: :flipper_replica
}
```

cc @ceritium

### DIFF
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -7,6 +7,7 @@ module Flipper
     class ActiveRecord
       include ::Flipper::Adapter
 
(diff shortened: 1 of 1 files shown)


ceritium:
It is great! Many thanks.

PR #630, opened 2022-05-01
CLI

bkeepers:
This adds a `flipper` CLI, which exposes basic flipper features over the command line.

```
$ flipper enable -g admin -a "User;1" foo
"foo" is enabled for 1 actor, 1 group

$ flipper disable foo
"foo" is disabled

$ flipper enable -p 50 foo
"foo" is enabled for 50% of actors

$ flipper enable foo
"foo" is fully enabled
```

Here's what's currently only my list to implement for the initial version:

- [x] `enable feature` / `disable feature`
  - [x] no args (boolean)
  - [x] -a, --actor
  - [x] -g, --group
  - [x] -p, --percentage-of-actors
  - [x] -t, --percentage-of-time
- [x] `-h`, `--help` to show CLI help
  - [x] list available commands
  - [x] `help [command]`
- [x] Initialize via local Rails app
- [x] `list`
- [x] `show feature`
- [ ] Better error messages (for wrong number of args, unknown option)
- [ ] Docs

Here's stuff that would be really nice to have eventually, but not a priority for this first version:

- Standalone mode
  - via `FLIPPER_CLOUD_TOKEN` env var
  - via config file in project or home directory
  - with browser workflow to authenticate
- `-e`, '--environment` option for all other commands to update features in remote environments
- `audit`
- `undo` to revert last feature change
- `environments`
  - list with no args
  - `-c`, `--create`
  - `-d`, `--delete`
- `webhooks`
- `tokens`
- `actors`


jnunemaker:
This looks good to me. I'd say finish up whatever you want and get it merged.

jnunemaker:
I was going to say it feels a little weird to have feature after the enablements. Not a blocker though and maybe it does work either way but the examples were just with feature key last.

bkeepers:
> I was going to say it feels a little weird to have feature after the enablements. Not a blocker though and maybe it does work either way but the examples were just with feature key last.

Fixed. Now either of these work:

```
$ flipper enable foo -p 50
$ flipper enable -p 50 foo
```

Issue #604, opened 2022-01-30
Latest release with default config throws "Flipper::Middleware::Memoizer appears to be running twice" warning


coreyaus:
Firstly, massive thanks for an excellent open-source gem! 🙏   And obviously let me know if this is some misunderstanding on my part rather than a bug (and apologies if that's the case!)

## Background on the 2 issues we experienced upgrading to `v0.23.1`

We've started to use Flipper recently with a very basic configuration and only a few features at this stage. We use the `flipper-active_record` gem, and don't have any custom configuration in any initializers or anything like that (e.g. no calls to `middleware.use` anywhere in our code). The only extra configuration we had in place was the suggested testing config for RSpec (to use the memory adapter in our specs): https://www.flippercloud.io/docs/testing#starting-fresh

Everything worked fine for us on v0.22.2 and v0.23.0

We started to see two issues when upgrading from v0.23.0 to v0.23.1 (see the code comparison for those 2 releases [here](https://github.com/jnunemaker/flipper/compare/v0.23.0...v0.23.1)):
- The following warning appeared frequently across lots of specs (essentially any specs that loaded `Flipper.enabled?(:some_feature)` conditionals somewhere in the executed code):
    > Flipper::Middleware::Memoizer appears to be running twice. Read how to resolve this at [#523](https://github.com/jnunemaker/flipper/pull/523)
- We started to see failures in some specs that were directly testing behaviour determined by `Flipper.enabled?(:some_feature)` conditionals. The tests followed the same simple structure as [these great demos](https://www.flippercloud.io/docs/testing#test-both-sides)

By putting a `binding.pry` in our test suite with our original configuration we were also seeing that `Flipper::Middleware::Memoizer` was auto-loaded in our middleware stack (potentially out of sequence, given it was throwing that "appears to be running twice" warning above):
```ruby
Rails.configuration.middleware.middlewares.map(&:inspect).include?('Flipper::Middleware::Memoizer')
=> true
```

## Temporary fix

We had to remove the following [suggested testing config](https://www.flippercloud.io/docs/testing#starting-fresh) from our spec helper file:

```ruby
# in spec_helper.rb
RSpec.configure do |config|
  config.before(:each) do
    Flipper.instance = Flipper.new(Flipper::Adapters::Memory.new)
  end
end
```

And instead add the equivalent code to `/config/environments/test.rb`:

```ruby
# in test.rb
Rails.application.configure do
[...]

PR #631, opened 2022-05-07
Remove use of Rack::BodyProxy

bkeepers:
Rack::BodyProxy allows deferring execution until after the response has sent to the client. It's intended for expensive cleanup (after the response is rendered, but before it's sent) that could delay delivering the response to the client. Cleaning up memoization is not an expensive operation, and I suspect this use of BodyProxy is the source for warnings in https://github.com/jnunemaker/flipper/issues/604.

I haven't studied it closely, but my guess is that after this middleware returns the rendered response body, puma (or possibly  any threaded web server) delivers the response to the client in another thread, returning execution of the app thread to Rails to start the next request. If the next request starts before the previous one is finished sending, then the next request will produce the warning.

_Updated to added more details._

### DIFF
--- a/lib/flipper/middleware/memoizer.rb
+++ b/lib/flipper/middleware/memoizer.rb
@@ -74,14 +74,9 @@ module Flipper
         when Array then flipper.preload(@opts[:preload])
         end
 
-        response = @app.call(env)
-        response[2] = Rack::BodyProxy.new(response[2]) do
-          flipper.memoize = false
-        end
-        reset_on_body_close = true
-        response
+        @app.call(env)
       ensure
-        flipper.memoize = false if flipper && !reset_on_body_close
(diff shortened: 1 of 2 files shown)


jnunemaker (approved):
Love it. Let's get this merged and out. I can do a release today or tomorrow for sure.

bkeepers:
@canderson-activatecare It looks like this does not fix the underlying issue in #604, so I'm not sure there's a good reason to merge it right now.

bkeepers:
After thinking about this some more, I think this should still be merged. It may not be the _only_ cause of #604, but I don't see how it can't be one cause. The Rails app should be done with execution by the time the block returns, so I think it's safe to stop memoization at that time.

Merging!

PR #636, opened 2022-06-08
Global instrumenter configuration

bkeepers:
While working on some changes for a different PR, I was running into some friction with having to pass the instrumenter option around everywhere. This doesn't seem like something that needs configured per object. This PR adds a global `Flipper.instrumenter` config and a `Flipper.instrument` method.

```diff
# config/initializers/flipper.rb
  Flipper.configure do |config|
    config.adapter do
-     Flipper::Adapters::Instrumented.new(my_adapter, instrumenter: MyInstrumenter)
+     Flipper::Adapters::Instrumented.new(my_adapter)
    end
+   config.instrumenter MyInstrumenter
  end
```

### DIFF
--- a/lib/flipper/cloud/configuration.rb
+++ b/lib/flipper/cloud/configuration.rb
@@ -41,14 +41,6 @@ module Flipper
       #  configuration.debug_output = STDOUT
       attr_accessor :debug_output
 
-      # Public: Instrumenter to use for the Flipper instance returned by
-      #         Flipper::Cloud.new (default: Flipper::Instrumenters::Noop).
-      #
-      #  # for example, to use active support notifications you could do:
-      #  configuration = Flipper::Cloud::Configuration.new
-      #  configuration.instrumenter = ActiveSupport::Notifications
-      attr_accessor :instrumenter
-
       # Public: Local adapter that all reads should go to in order to ensure
       # latency is low and resiliency is high. This adapter is automatically
       # kept in sync with cloud.
@@ -88,14 +80,11 @@ module Flipper
         @adapter_block = ->(adapter) { adapter }
         self.url = options.fetch(:url) { ENV.fetch("FLIPPER_CLOUD_URL", DEFAULT_URL) }
 
-        instrumenter = options.fetch(:instrumenter, Instrumenters::Noop)
-
         # This is alpha. Don't use this unless you are me. And you are not me.
         cloud_instrument = options.fetch(:cloud_instrument) { ENV["FLIPPER_CLOUD_INSTRUMENT"] == "1" }
-        @instrumenter = if cloud_instrument
-          Instrumenter.new(brow: brow, instrumenter: instrumenter)
-        else
-          instrumenter
+        if cloud_instrument
+          # FIXME: replace with a subscriber to AS::Notifications
+          Flipper.instrumenter Instrumenter.new(brow: brow, instrumenter: Flipper.instrumenter)
         end
       end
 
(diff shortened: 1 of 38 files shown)


lib/flipper/cloud/configuration.rb
```diff
@@ -88,14 +82,15 @@ def initialize(options = {})
...
-          Instrumenter.new(brow: brow, instrumenter: instrumenter)
-        else
-          instrumenter
+        if cloud_instrument
+          require 'flipper/instrumentation/cloud_subscriber'
+
+          Flipper.instrumenter.subscribe(
+            Flipper::Feature::InstrumentationName,
+            Flipper::Instrumentation::CloudSubscriber.new(brow)
+          )
```
bkeepers, line 93: Just a heads up that I rewrote this to use a subscriber instead of wrapping the instrumenter. I think it's fair to say that anyone that wants to use this should be using `AS::Notifications` (default in Rails) or some other instrumentation framework.

Doesn't seem like there were any tests for this, and I have not tested it manually yet.

jnunemaker: The new telemetry stuff needs to work as a wrapper so that and this will conflict.

lib/flipper/configuration.rb
```diff
@@ -1,8 +1,9 @@
 module Flipper
   class Configuration
     def initialize(options = {})
-      @default = -> { Flipper.new(adapter) }
-      @adapter = -> { Flipper::Adapters::Memory.new }
+      adapter { Flipper::Adapters::Memory.new }
```
jnunemaker, line 4: Wow! That looks nice. I know its a little thing, but sweet.

spec/flipper/adapters/failover_spec.rb
```diff
@@ -84,21 +84,19 @@
   context 'when primary is instrumented and fails' do
     before do
       allow(memory_adapter).to receive(:get).and_raise(Net::ReadTimeout)
+      Flipper.instrumenter = instrumenter
```
jnunemaker, line 87: This is ok because it's actually updating config and config is reset for every spec, right? Just did a double take because global changes without resetting always make me shiver. :)

spec/flipper/instrumentation/statsd_subscriber_spec.rb
```diff
@@ -6,16 +6,16 @@
   let(:statsd_client) { Statsd.new }
   let(:socket) { FakeUDPSocket.new }
   let(:adapter) do
-    memory = Flipper::Adapters::Memory.new
-    Flipper::Adapters::Instrumented.new(memory, instrumenter: ActiveSupport::Notifications)
+    Flipper::Adapters::Instrumented.new(Flipper::Adapters::Memory.new)
```
jnunemaker, line 9: Now stuff like this can use `use` right?

jnunemaker (approved):
I'm good with this. It does simplify things and no one in their right mind would use different instrumenters for different parts of flipper. Left one note about the cloud instrumenter. Feel free to hit me up about that when looking to merge this.

lib/flipper/cloud/configuration.rb
bkeepers: > The main reason is that AS notifications are super slow. I don't want people (including me) to have to use those in order to get analytics.

FWIW, the engine configures AS Notifications by default in all Rails apps, so every app you manage right now is probably already using it.

jnunemaker: K. I thought we turned that off for some reason but I can see it is not. I'm sure my apps aren't sensitive to it but I've helped a handful of people debug perf issues and AS::Notification was the easiest win each time.

Magkoooh (approved):

Magkoooh (approved):

PR #638, opened 2022-06-08
Refactor memoization

bkeepers:
This adds `#memoize(&block)` method to perform memoization. It removes conditional memoization logic and mutating adapter/DSL in favor of just returning new DSL instance with wrapped adapter.

My best guess as to the cause of #604 is just that there is a lot of internal state mutation happening when enabling memoization. This branch is an attempt to reduce the state mutation down to just setting `Flipper.instance` to a new DSL with a wrapped adapter.

TODO:
- [ ] Make this work with `flipper-cloud`


bkeepers:
Punting on this for now.

PR #639, opened 2022-06-21
Add Dependabot for GitHub Actions

petergoldstein:
There's an outdated checkout action in the test configuration, so once this is merged we should expect to see Dependabot open a PR to update it.


jnunemaker (approved):
I know nothing about this but let's do it.

Issue #659, opened 2022-08-23
Redis::Deprecated:   Redis#sadd will always return an Integer in Redis 5.0.0. Use Redis#sadd? instead


bf4:
in our app [Bump redis from 4.7.1 to 4.8.0](https://github.com/redis/redis-rb/pull/1134) we started getting errors from Flipper here

- Introduce sadd? and srem? as boolean returning versions of sadd and srem.
- Deprecate sadd and srem returning a boolean when called with a single argument. To enable the redis 5.0 behavior you can set Redis.sadd_returns_boolean = true.

PR #660, opened 2022-08-23
Flipper fails with Redis >= 4.8.0

bf4:
closes https://github.com/jnunemaker/flipper/issues/659

### DIFF
--- a/lib/flipper/adapters/redis.rb
+++ b/lib/flipper/adapters/redis.rb
@@ -28,7 +28,7 @@ module Flipper
 
       def self.redis_sadd_returns_boolean?
         return @redis_sadd_returns_boolean if defined?(@redis_sadd_returns_boolean)
-        @redis_sadd_returns_boolean = true
+        @redis_sadd_returns_boolean =
           if ::Redis.respond_to?(:sadd_returns_boolean)
             ::Redis.sadd_returns_boolean
           else
--- a/lib/flipper/adapters/redis.rb
+++ b/lib/flipper/adapters/redis.rb
@@ -26,19 +26,9 @@ module Flipper
         read_feature_keys
       end
 
-      def self.redis_sadd_returns_boolean?
-        return @redis_sadd_returns_boolean if defined?(@redis_sadd_returns_boolean)
-        @redis_sadd_returns_boolean =
-          if ::Redis.respond_to?(:sadd_returns_boolean)
-            ::Redis.sadd_returns_boolean
-          else
-            false
-          end
-      end
-
       # Public: Adds a feature to the set of known features.
       def add(feature)
-        if self.class.redis_sadd_returns_boolean?
+        if redis_sadd_returns_boolean?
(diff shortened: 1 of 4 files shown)


lib/flipper/adapters/redis.rb
```diff
@@ -26,15 +26,33 @@ def features
...
+            ::Redis.sadd_returns_boolean
+          else
+            false
+          end
+      end
+
       # Public: Adds a feature to the set of known features.
       def add(feature)
-        @client.sadd FeaturesKey, feature.key
+        if self.class.redis_sadd_returns_boolean?
```
bf4, line 41: Not sure how you might want to address this so the conditional is evaluated the minimal number of times.

lib/flipper/adapters/redis.rb
```diff
@@ -26,15 +26,33 @@ def features
         read_feature_keys
       end
 
+      def self.redis_sadd_returns_boolean?
+        return @redis_sadd_returns_boolean if defined?(@redis_sadd_returns_boolean)
+        @redis_sadd_returns_boolean = true
+          if ::Redis.respond_to?(:sadd_returns_boolean)
```
mikdiet, line 32: it will return `true` on the second and so on call.

You probably meant

```suggestion
        @redis_sadd_returns_boolean =
          if ::Redis.respond_to?(:sadd_returns_boolean)
```

bf4: Lol, yes. That's embarrassing

lib/flipper/adapters/redis.rb
bkeepers: `Redis.sadd_returns_boolean` is just an `attr_accessor`, so I don't think it's that important to cache the lookup.

bkeepers:
Thanks @bf4! 0.25.2 is released

lib/flipper/adapters/redis.rb
bf4: well, it was really the `respond_to?` that I wanted to avoid calling a lot, and I didn't want to introduce any meta programming since that wasn't a pattern I saw used to evaluate fewer times. On the other hand, enable/disable isn't really a hotspots, so `¯\_(ツ)_/¯`

Thanks for the attention!

lib/flipper/adapters/redis.rb
```diff
@@ -28,13 +28,21 @@ def features
...
+          @client.sadd FeaturesKey, feature.key
+        end
         true
       end
 
       # Public: Removes a feature from the set of known features.
       def remove(feature)
-        @client.srem FeaturesKey, feature.key
+        if redis_sadd_returns_boolean?
+          @client.srem? FeaturesKey, feature.key
```
bf4, line 42: looking at https://github.com/mperham/sidekiq/issues 5484 I guess we could have 

```diff
-          @client.srem FeaturesKey, feature.key
+          @client.srem FeaturesKey, [feature.key]
```

I'm no redis expert, though

PR #672, opened 2022-10-09
add method disabled?to feature

italomatos:
This PR has a goal to add ```disabled?``` method to Feature.

This method is opposite to ```enabled?``` method


italomatos:
I saw in other PR that no make sense create this method :) That why I closed

PR #677, opened 2022-11-02
Check sadd_returns_boolean on the actual client class rather than `::Redis`

Intrepidd:
Because the client might not be a `::Redis` instance, for example it could be a [`MockRedis` ](https://github.com/sds/mock_redis/) instance

### DIFF
--- a/lib/flipper/adapters/redis.rb
+++ b/lib/flipper/adapters/redis.rb
@@ -118,7 +118,7 @@ module Flipper
       private
 
       def redis_sadd_returns_boolean?
-        ::Redis.respond_to?(:sadd_returns_boolean) && ::Redis.sadd_returns_boolean
+        @client.class.respond_to?(:sadd_returns_boolean) && @client.class.sadd_returns_boolean
       end
 
       def read_many_features(features)


Intrepidd:
Thanks !

> I've been giving [www.flippercloud.io](http://www.flippercloud.io/) credits to people who contribute, so if you want to check it out beyond the trial, let me know and I'll spot you several months or more.

I might, thanks a lot :) 

> Lastly, if you don't mind, I would love to know what you are using flipper for as well and how it is working out for you (what went well, what was hard other than what you mentioned here).

I mostly use-it for for pre-releasing some features to some users, or to handle VIP features for some select users.
It's mostly a breeze !

One "issue" I recently had was the following : 
- I had a feature turned for specific actors
- I wanted to enable it to everyone
- But I also wanted to be able to easily roll it back to the previous specific actors in case something went wrong
- My solution was to screenshot the list of actors, which is obviously a pain

But looks like the cloud version has changes history so that would fix the problem, but it would be great in the free version too :)

jnunemaker:
> But looks like the cloud version has changes history so that would fix the problem, but it would be great in the free version too :)

Correct, building this into the open source for every adapter would have been painful. Cloud has audit history, but even better than that it has rollback. You just click a button and it shows you what it would look like to roll back to any point in time and your done.

## Audit Example

Screen Shot 2022-11-03 at 3 23 10 PM

## Rollback Example

Screen Shot 2022-11-03 at 3 23 17 PM

hlascelles:
Note, we are seeing an error `undefined method `sadd?' for #<MockRedis::Database:>` as we are using [redis-namespace](https://github.com/resque/redis-namespace) and [mock_redis](https://github.com/sds/mock_redis). The redis-namespace gem indicates that `sadd?` is supported, but the underlying `MockRedis::Database` does not.

You can solve it in specs with:

```
Redis::Namespace.sadd_returns_boolean = false
```

PR #680, opened 2022-11-10
Flipper UI - Fix issue preventing feature flags being enabled when confirm_fully_enable is on and feature_removal_enabled is off

Lockyy:
I noticed that the confirmation text set up by `confirm_fully_enable` does not work correctly if `feature_removal_enabled` was set to `false`.

This was due to the node with ID `feature_name` only being present within the form included when `feature_removal_enabled` was set to true. This lead to a situation where your UI would display "undefined" as the prompt text though when that or the actual feature name was used the  flag was not toggled since it was comparing the string "undefined" to an undefined variable.

This commit fixes this edge case by decoupling the confirmation text for both buttons from the `feature_removal_enabled` form and moving them into data attributes on the button being clicked. This allows them to be enabled/disabled separately without impacting each other. The appearance of these attributes are controlled by the flags themselves.

| Before | After |
|-|-|
| before | after |

### DIFF
--- a/lib/flipper/ui/views/feature.erb
+++ b/lib/flipper/ui/views/feature.erb
@@ -243,6 +243,9 @@
               <div class="col">
                 <button type="submit" name="action" value="Enable" <% if Flipper::UI.configuration.confirm_fully_enable %>id="enable_feature__button"<% end %> class="btn btn-outline-success btn-block" <% if Flipper::UI.configuration.read_only %>disabled<% end %>>
                   <span class="d-block" data-toggle="tooltip"
+                    <% if Flipper::UI.configuration.confirm_fully_enable %>
+                      data-confirmation-text="<%= feature_name %>"
+                    <% end %>
                     <% if Flipper::UI.configuration.read_only %>
                       title="Fully enable is not allowed in read only mode."
                     <% else %>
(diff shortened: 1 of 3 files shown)


lib/flipper/ui/views/feature.erb
```diff
@@ -289,7 +292,7 @@
             <%== csrf_input_tag %>
             <input type="hidden" id="feature_name" name="_feature" value="<%= feature_name %>">
```
Lockyy, line 293: I expect that this line can be removed now as it was added when the confirmation text box was put in place. I didn't touch it however as I wanted to avoid unintended side-effects. 

Does anyone have any input on whether this can be removed or if it's used anywhere else? Since it is an input and would get sent back to the server.

jnunemaker:
Weird! I never got a notification about this. Checking it now.

jnunemaker:
@Lockyy thanks!

Lockyy:
Thanks for merging this in! Glad to be able to help. :)

jnunemaker:
@Lockyy I've been giving www.flippercloud.io credits to people who contribute, so if you want to check it out beyond the trial, let me know and I'll spot you several months or more.

PR #586, opened 2021-12-01
Fix railtie initialization to mount middleware after config/intializers/*

bkeepers:
After #563, Flipper configuration in `config/initializers/*` is ignored because the Memoization middleware gets mounted before initializers are run. This reverts 713e9d1 and fixes the specs to expose this issue.

The tests pass here, but it's going to cause issues that #563 was meant to fix, so I need some help.

I've been spelunking on Rails initialization process all day, and I can't quite figure out the order of initializers. Some initializers run after `load_config_initializers` without declaring a `before`/`after` dependency (see all of `action_view` below), while most run earlier in the process.

### What else I tried

* Running `flipper.memoizer` before `build_middleware_stack` initializer, but that appears to be defined on a `Finalizer` module outside of the app's initializers, so you can't order anything before/after it.
* Removing all `before`/`after` dependencies to see if it would just run in the right order automagically 🤷 

### Intializer order

Here's the order for a few different examples:

<details>
<summary>Initializers for new Rails 6.1.4.1 app</summary>
<pre>
>> puts Rails.application.initializers.tsort.map(&:name).uniq
load_environment_config
load_environment_hook
load_active_support
set_eager_load
initialize_logger
initialize_cache
initialize_dependency_mechanism
set_load_path
set_autoload_paths
set_eager_load_paths
bootstrap_hook
set_secrets_root
active_support.set_authenticated_message_encryption
active_support.reset_all_current_attributes_instances
active_support.deprecation_behavior
active_support.initialize_time_zone
active_support.initialize_beginning_of_week
active_support.require_master_key
active_support.set_configs
active_support.set_hash_digest_class
action_dispatch.configure
active_model.secure_password
active_model.i18n_customize_full_message
action_controller.assets_config
action_controller.set_helpers_path
action_controller.parameters_config
action_controller.set_configs
action_controller.compile_config_methods
action_controller.request_forgery_protection
action_controller.eager_load_actions
active_record.initialize_timezone
active_record.logger
active_record.backtrace_cleaner
active_record.migration_error
active_record.database_selector
Check for cache versioning support
active_record.check_schema_cache_dump
active_record.define_attribute_methods
active_record.warn_on_records_fetched_greater_than
active_record.set_configs
active_record.initialize_database
active_record.log_runtime
active_record.set_reloader_hooks
active_record.set_executor_hooks
active_record.add_watchable_files
active_record.clear_active_connections
active_record.set_filter_attributes
active_record.set_signed_id_verifier_secret
global_id
active_job.logger
[...]

PR #681, opened 2022-11-10
Add ENV vars to railtie

dudo:
We've been having some issues with load order of this railtie. We have a gem that loads several other gems (including flipper, obviously) that has its own railtie. 

The load order of these railties in railties in railties is fragile, and we feel that adding some ENV vars would alleviate this fragility.

If this is something you're ok with, I'll add documentation and tests around this change.


jnunemaker (commented):
@bkeepers any thoughts? Most seem fine. The only part I'm in the fence about is the constantize.

lib/flipper/railtie.rb
```diff
@@ -2,11 +2,11 @@ module Flipper
...
     config.before_configuration do
       config.flipper = ActiveSupport::OrderedOptions.new.update(
-        env_key: "flipper",
-        memoize: true,
-        preload: true,
-        instrumenter: ActiveSupport::Notifications,
-        log: true
+        env_key: ENV.fetch('FLIPPER_ENV_KEY', 'flipper'),
+        memoize: ENV.fetch('FLIPPER_MEMOIZE', 'true') == 'true',      
+        preload: ENV.fetch('FLIPPER_PRELOAD', 'true') == 'true',
```
bkeepers, line 7: I'd like to see the casting here be a little more robust. Maybe use `ActiveModel::Type::Boolean.new.cast(value)`?

bkeepers (commented):
I love ENV vars and this seems like a reasonable approach to me, however, I'd also like to hear more about the issue you're having in case other people are having the same issue.

lib/flipper/railtie.rb
dudo: Well... this is a railtie. Probably safe to assume ActiveModel is loaded? That being said string comparison is pretty robust, no? No need to over complicate things.

dudo:
@jnunemaker `constantize` feels like the safest bet, since we're in a railtie. The default is loading `ActiveSupport`, so if we expect that to be ok, then we should be able to lean on its helpers?

dudo:
Sounds like ya'll are ok with the change, though - I'll update the docs and add specs.

dudo:
The readme doesn't feel like the right place for this config (as that mainly applies to Flipper usage, not config). Is the documentation open source? And while we're talking about it... can we add search to the docs?!

lib/flipper/railtie.rb
dudo: For `preload`, we can provide an array... do we want to support that via the ENV?

dudo: Bumped to `casecmp` to catch `False`

dudo:
@jnunemaker Any more suggestions to get this merged in?

jnunemaker:
Sorry! I’ve been swamped. I’ll try to get this merged this week.
Appreciate the work!

On November 30, 2022, David Grevink ***@***.***> wrote:
> @jnunemaker <https://github.com/jnunemaker> Any more suggestions to
> get this merged in?
>

—
Reply to this email directly, view it on GitHub
<https://github.com/jnunemaker/flipper/pull/681#issuecomment-
1332416430>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAAAB25MQORG2VYJDBK2FFTWK547NANCNFSM6AAAAAAR4ZU5TM>.
You are receiving this because you were mentioned.Message ID:
***@***.***>

jnunemaker:
> The readme doesn't feel like the right place for this config (as that mainly applies to Flipper usage, not config). Is the documentation open source? 

This is the downside of our doc move. The upside is now people realize that cloud exists which helps us grow that side of things and thus flipper. If you want to just drop a few paragraphs here i can move them into the docs.

> And while we're talking about it... can we add search to the docs?!

I just went through the process with algolia to add search. Should be out soon.

jnunemaker:
@dudo https://www.flippercloud.io/docs has search now powered by algolia. We've noticed a few things we need to tweak but its good enough to release.

dpep:
thanks for this, @dudo !

@jnunemaker : this relates to the work we did in https://github.com/jnunemaker/flipper/pull/586 regarding railtie load order issues.  unfortunately, it bit my company again during a flipper upgrade

bkeepers:
@dpep can you explain what happened?

dpep:
@bkeepers: essentially, we have an internal gem that configures Flipper for internal services.  it does so within a rails initializer, which itself gets configured when the gem is loaded.  we noticed that the services using this gem didn't have Flipper configured as expected and we suspect a load order issue.  it's entirely possible that our convoluted setup is to blame here, but after a few failed attempts to correct the issue it feels safer to switch tactics eg. @dudo's ENV variable bypass in this PR

PR #682, opened 2022-11-10
Cloud Background Polling

jnunemaker:
Polling in the foreground is never a good idea as it ties availability to whatever is being polled. That's why we recommend syncing with webhooks. But, sometimes people can't use webhooks for whatever reason, so this provides a polling mechanism that updates in a background thread. The foreground threads just check to see if any updates are possible and attempt a sync if so. 

I'm no expert, but I think we have this in a good place. Its a similar pattern to brow which I've been using in production to send events for half a year or more. 

That said, if anyone wants to dig into the threaded part and make suggestions, I'd love that.


sirwolfgang:
I think the [docs](https://www.flippercloud.io/docs/adapters) need to be updated to include this adaptor.

bkeepers:
@sirwolfgang yep, the docs need updated, but we haven't fully finalized the API and implementation of the background poller (see #699, #700, #701), so for now it's considered and internal API. We'll get it documented as soon as we're confident that the API is stable.

Issue #686, opened 2022-12-06
Executing arbitrary logic when a feature flag status gets retrieved from storage


nicolasiensen:
I checked [Bugsnag's new feature](https://docs.bugsnag.com/product/features-experiments/), where we can now include the enabled feature flags to be part of the error reports.

They updated their Ruby SDK to allow developers to include feature flags as part of the error reports like so:

```ruby
Bugsnag.add_feature_flag('New checkout flow')
```

On my team, we disabled Flipper's preload due to performance implications, so we would like to call `Bugsnag.add_feature_flag` every time we interact with Flipper to retrieve the status of a particular feature flag. Can I do that already?

One way to achieve this would be to make the `Feature` class observable and register observers in the library like the example below:

```ruby
class FeatureCheckObserverForBugsnag < Flipper::FeatureCheckObserver
  def call(feature, enabled)
    if enabled
      Bugsnag.add_feature_flag(feature.name)
    else
      Bugsnag.clear_feature_flag(feature.name)
    end
  end
end

Flipper.add_feature_check_observer(FeatureCheckObserverForBugsnag.new)

# Flipper will now include the new observer on every new instance of Feature that it instantiates
```

Let me know if there is already an approach to get the outcome I'm looking for and if you have any comments about my proposal. I will happily implement it if you think it is worth it.

Cheers!

bkeepers:
@nicolasiensen This should be possible with Flipper's existing support for [Instrumentation](https://www.flippercloud.io/docs/instrumentation) (assuming you're using Rails, it everything will already be configured for you). I'll update the docs based on this discussion because that link doesn't show you how to do custom stuff, but try adding this to an initializer in your app:

```ruby
ActiveSupport::Notifications.subscribe('feature_operation.flipper') do |name, start, finish, id, payload|
  if payload[:operation] == :enabled?
    puts "#{payload[:feature_name]} enabled? #{payload[:result]}"
  end
end
```

Now when I run this code…

```ruby
Flipper.enabled? :instrument
Flipper.enable :instrument
Flipper.enabled? :instrument
```

…I see this:

```
my_feature enabled? false
my_feature enabled? true
```

Other operations besides `:enabled?` are `:enable`, `:disable`, `:add`, `:remove`, `:exist?`, and `:clear`.

Let me know if this doesn't solve your problem and I'll reopen the issue.

PR #687, opened 2022-12-16
Bugsnag integration

bkeepers:
#686 got me thinking about flipper integration with other libraries and I wanted to see what it would like if we just automatically loaded integrations.

This adds an instrumentation subscriber for Bugsnag which uses it's support for [feature flags & experiments](https://docs.bugsnag.com/product/features-experiments/) to mark features as enabled/disabled whenever they are checked.

I'm guessing eventually someone will want to customize the integration. We could just provide a way to disable automatic loading and they can add their own subscriber.

Thoughts/concerns about this approach?

### DIFF
--- a/lib/flipper/instrumentation/bugsnag.rb
+++ b/lib/flipper/instrumentation/bugsnag.rb
@@ -1,0 +1,21 @@
+module Flipper
+  module Instrumentation
+    class Bugsnag
+      def call(event)
+        operation, feature_name, result = event.payload.values_at(:operation, :feature_name, :result)
+        return unless operation == :enabled?
+
+        if result
+          ::Bugsnag.add_feature_flag(feature_name)
+        else
+          ::Bugsnag.clear_feature_flag(feature_name)
+        end
+      end
+    end
+
+    # Register the subscriber if using ActiveSupport::Notifications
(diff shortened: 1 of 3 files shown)


lib/flipper/instrumentation/bugsnag.rb
```diff
@@ -0,0 +1,21 @@
+module Flipper
+  module Instrumentation
+    class Bugsnag
+      def call(name, start, finish, id, payload)
+        operation, feature_name, result = payload.values_at(:operation, :feature_name, :result)
+        return unless operation == :enabled?
+
+        if result
```
jnunemaker, line 8: Neat! Any thoughts on if a feature flag returns true on a page in one spot and then false in another? For example, % of time can do this. It seems like this implementation would leave it with whatever was last. Correct?

bkeepers: Correct, this implementation means last flag wins. I think that is the behavior you would want so any bugs reported show what the last value was of a feature check.

bkeepers:
There hasn't been a lot of interest in this. I'll publish this as a blog post for now and reconsider doing this in the future.

PR #693, opened 2023-01-16
Remove executables directive, this gem has 0 exe

olleolleolle:
This PR cleans up two small things in the gemspec:

- Drop reference to non-existent bin/ directory
- Drop Travis reference


jnunemaker:
@olleolleolle Thanks! Every little bit helps.

PR #696, opened 2023-02-15
CI: Use bundler-cache: true to simplify the installation parts

olleolleolle:
This PR tries to use `bundler-cache: true`, and in order to do that, moves around some instructions.

- [x] workflows/ci.yml
- [x] workflows/examples.yml


olleolleolle:
@jnunemaker There! This seems like it does the thing.

jnunemaker:
Neat!

PR #700, opened 2023-03-05
Rename Flipper::Adapters::Poll::Poller to Flipper::Poller

bkeepers:
I'm working on a another branch that refactors some of the other synchronization and memoization code, all of which use `Poller`, so I propose we hoist it to a top level `Flipper::Poller` constant.


jnunemaker (approved):
Should be fine, though I'd rather we just get a poll adapter that can do whatever we need rather than having people reach for poller (long term at least).

PR #701, opened 2023-03-06
Refactor Sync and Poll adapters to remove duplication

bkeepers:
They are effectively doing the same thing with just slightly different logic. This makes `Poll < Sync` and updates `Sync` to use `DualWrite` internally.

I don't think `Sync` is used anymore internally, so it might make sense to just depreciate and remove it, but not sure if there are other use cases for it.


bkeepers:
Skipping this for now since it caused some test to fail unexpectedly. I think we should eventually just deprecate the `Sync` adapter.

PR #702, opened 2023-03-08
Make Memory adapter threadsafe

bkeepers:
Addresses https://github.com/jnunemaker/flipper/pull/699#discussion_r1126465923

~~This uses `Concurrent::Hash` (which just uses `Hash` on MRI since any C function is thread safe and Hash is implemented in C) as the backend source for the memory adapter and replaces any non-threadsafe uses (like `||=`).~~

See [update](https://github.com/jnunemaker/flipper/pull/702#issuecomment-1462629962)

### DIFF
--- a/lib/flipper/adapters/memory.rb
+++ b/lib/flipper/adapters/memory.rb
@@ -1,5 +1,3 @@
-require 'set'
-
 module Flipper
   module Adapters
     # Public: Adapter for storing everything in memory.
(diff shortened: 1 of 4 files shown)


bkeepers:
Update: After experimenting `Concurrent::Hash` and several other classes from concurrent-ruby, I settled on just using a `Mutex` to wrap all adapter operations. It makes it a lot cleaner, and since this is all data in memory it should have minimal performance impact.

PR #708, opened 2023-03-16
Speed up skipped adapter specs

bkeepers:
Minor tweak to speed up skipping of adapter specs in local dev if underlying sever is not available. Currently each example has to wait for the error, which could be ~1 second per example. This sets a global and just short-circuits the skip if one of the pervious examples in the example group fail.

Before:
```
Finished in 50.18 seconds (files took 5.08 seconds to load)
1719 examples, 0 failures, 69 pending
```

After:
```
Finished in 10.86 seconds (files took 2.63 seconds to load)
1719 examples, 0 failures, 69 pending
```

### DIFF
--- a/spec/support/skippable.rb
+++ b/spec/support/skippable.rb
@@ -1,0 +1,18 @@
+RSpec.configure do |config|
+  config.before(:all) do
+    $skip = false
+  end
+
+  def skip_on_error(error, message, &block)
+    # Premptively skip if we've already skipped
+    skip(message) if $skip
+    block.call
+  rescue error
+    if ENV["CI"]
+      raise
+    else
+      $skip = true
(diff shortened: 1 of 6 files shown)


spec/support/skippable.rb
```diff
@@ -0,0 +1,18 @@
+RSpec.configure do |config|
+  config.before(:all) do
+    $skip = false
```
bkeepers, line 3: Tried using `@@skip` here, but can't use class variables in rspec. I couldn't figure out any other way to share state across examples.

jnunemaker: Probably not Luke have to make a Skip class or module with top level var. I’m fine either way.

PR #709, opened 2023-03-17
Easy Import/Export

jnunemaker:
I'd like to make it easy to backup flipper using just flipper (instead of backing up the external storage the adapter is talking to). This is a first pass. I'll look it over again in the morning, but it adds the concept of exporting the current state of the world. 

## TODO

* [x] Make it so import can accept an export (in addition to the flipper instance or adapter it currently supports in places)
* [x] Add import of an export to the UI

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,10 +10,10 @@ gem 'debug'
 gem 'rake', '~> 12.3.3'
 gem 'statsd-ruby', '~> 1.2.1'
 gem 'rspec', '~> 3.0'
-gem 'rack-test', '~> 0.6.3'
+gem 'rack-test'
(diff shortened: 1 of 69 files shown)


jnunemaker:
@bkeepers I'm not all the way done but feel free to look this over and suggest when you have time.

bkeepers (approved):
This looks good! I'm still surprised that `Set` requires manual conversion to an `Array` to serialize as JSON

jnunemaker:
> This looks good! I'm still surprised that `Set` requires manual conversion to an `Array` to serialize as JSON

Yeah, me too. It doesn't in Rails. I should see what they do... likely the `as_json` stuff.

PR #710, opened 2023-03-19
Allow passing multiple actors to Flipper.enabled?

jnunemaker:
Before: 

```ruby
[user, team, org].any? { |actor| Flipper.enabled?(:some_feature, actor) }
```

After:

```ruby
Flipper.enabled?(:some_feature, [user, team, org])
```

The biggest benefit is not re-running all the gates and instrumentation for every actor and every gate. Instead this runs the boolean and & of time once. Then it just checks multiple actors in the actor, % of actors and group gates.

I'm going to give this another pass to see if anything needs to change back or not.

### DIFF
--- a/lib/flipper/gates/percentage_of_actors.rb
+++ b/lib/flipper/gates/percentage_of_actors.rb
@@ -31,11 +31,8 @@ module Flipper
       # Returns true if gate open for any actors, false if not.
       def open?(context)
         return false unless context.actors?
-
-        context.actors.any? do |actor|
-          id = "#{context.feature_name}#{actor.value}"
-          Zlib.crc32(id) % (100 * SCALING_FACTOR) < context.values.percentage_of_actors * SCALING_FACTOR
-        end
+        id = "#{context.feature_name}#{context.actors.map(&:value).sort.join}"
+        Zlib.crc32(id) % (100 * SCALING_FACTOR) < context.values.percentage_of_actors * SCALING_FACTOR
       end
 
       def protects?(thing)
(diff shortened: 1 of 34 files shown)


lib/flipper/gates/percentage_of_actors.rb
```diff
@@ -26,14 +26,16 @@ def enabled?(value)
...
       def open?(context)
-        return false if context.thing.nil?
+        return false unless context.actors?
 
-        id = "#{context.feature_name}#{context.thing.value}"
-        Zlib.crc32(id) % (100 * SCALING_FACTOR) < context.values.percentage_of_actors * SCALING_FACTOR
+        context.actors.any? do |actor|
+          id = "#{context.feature_name}#{actor.value}"
+          Zlib.crc32(id) % (100 * SCALING_FACTOR) < context.values.percentage_of_actors * SCALING_FACTOR
+        end
```
bkeepers, line 38: I think this will change the probability. For example, if this is enabled for 50% and there are 6 actors, it would be the same as rolling a dice 6 times and seeing if **any** of the dice rolls are < 4. Each additional actor increases the probability (6x in this example).

More accurately would be to calculate the CRC once and compare that to a percentage:

```ruby
id = "#{context.feature_name}#{context.actors.map(&:value).join(';')}"
Zlib.crc32(id) % (100 * SCALING_FACTOR) < context.values.percentage_of_actors * SCALING_FACTOR
```

jnunemaker: Hmm. I’m not sure what is right. I was trying to make it the same probability as checking each actor individually would be. 

If we change this then how will people migrate to get the array of actors benefit?

jnunemaker: Resolved by 63f4f695253ded4b138825f808c754257eac5a1d

jnunemaker:
Forgot to drop here that checking 5 actors at once vs 5 enabled checks with 1 actor each is like 5x faster so that's good!

bkeepers (approved):

PR #715, opened 2023-03-27
Refactor expressions

bkeepers:
This is a fairly significant refactor of Expressions (#692).

1. Adds `Constant` (which quacks like an `Expression`) to represent explicit string, number, and boolean values. Note that there are still `String`, `Number`, and `Boolean` _expressions_, which will perform casting of values passed to them, but those expressions are no longer used by default to represent constant values when serialized:
    ```diff
    {
      "Equal" => [
        {"Property" => ["plan"]},
    -     {"String" => ["basic"]},
    +     "basic",
      ]
    }    ```
2. A new expression is now a simple object that responds to `#call` and receives any arguments instead of subclassing `Expression`. All evaluating of arguments is done internally in `Expression` before being passed to the `#call` method. If the `context: ` key is part of the method signature, it will be passed in.
    ```diff
    - class Flipper::Expressions::Property < Expression
    -   def initialize(args)
    -     super Array(args)
    -   end
    - 
    -   def evaluate(context = {})
    -     key = evaluate_arg(0, context)
    -     context.dig(:properties, key.to_s)
    -     end
    -   end
    - end
    + class Flipper::Expressions::Property
    +   def self.call(key, context:)
    +     context.dig(:properties, key.to_s)
    +   end
    + end
    ```

This greatly simplifies the process of adding new expressions internally, and paves the way for making it easy for users to create expressions of their own.


lib/flipper/expression/builder.rb
```diff
@@ -0,0 +1,73 @@
...
+      def any?
+        is_a?(Expression) && function == Expressions::Any
+      end
+
+      def all?
+        is_a?(Expression) && function == Expressions::All
+      end
+
+      def group?
+        any? || all?
```
jnunemaker, line 69: I'd be tempted to do function.group? so the function can determine if it is a group or not. Then new functions don't require changes to the builder and this method. Just a thought, not a blocker.

lib/flipper/expression/constant.rb
```diff
@@ -0,0 +1,26 @@
...
+      include Expression::Builder
+
+      attr_reader :value
+
+      # Override initialize to avoid trying to build args
+      def initialize(value)
+        @value = value
+      end
+
+      def evaluate(context = {})
```
jnunemaker, line 16: If we do context = nil would that save a hash allocation?

lib/flipper/expressions/duration.rb
```diff
@@ -0,0 +1,28 @@
+module Flipper
+  module Expressions
+    class Duration
+      SECONDS_PER = {
+        "second" => 1,
+        "minute" => 60,
+        "hour" => 3600,
+        "day" => 86400,
+        "week" => 604_800,
+        "month" => 26_29_746,  # 1/12 of a gregorian year
```
jnunemaker, line 10: 2_629_746?

lib/flipper/expressions/now.rb
```diff
@@ -0,0 +1,9 @@
+module Flipper
+  module Expressions
+    class Now
+      def self.call
+        ::Time.now
```
jnunemaker, line 5: Should this be `.utc`?

lib/flipper/typecast.rb
```diff
@@ -36,17 +36,25 @@ def self.to_float(value)
...
-    def self.to_percentage(value)
-      result_to_f = value.to_f
-      result_to_i = result_to_f.to_i
-      result_to_f == result_to_i ? result_to_i : result_to_f
+    def self.to_number(value)
+      case value
+      when Numeric
+        value
+      when String
+        value.include?('.') ? to_float(value) : to_integer(value)
```
jnunemaker, line 48: We may want to bench this against the to_f/to_i shenanigans we had.

bkeepers: main:
```
Typecast.to_number 1      7.863M (± 0.3%) i/s -     40.021M in   5.090105s
Typecast.to_number 1.1
                          7.879M (± 0.3%) i/s -     39.979M in   5.074375s
Typecast.to_number '1'
                          7.004M (± 0.8%) i/s -     35.206M in   5.026764s
Typecast.to_number '1.1'
                          6.919M (± 0.8%) i/s -     35.179M in   5.085068s
Typecast.to_number nil
                          7.833M (± 0.1%) i/s -     39.786M in   5.079194s
Typecast.to_number Time.now
                          6.579M (± 0.4%) i/s -     33.543M in   5.098330s
```

this branch:
```
Typecast.to_number 1     11.578M (± 1.1%) i/s -     58.624M in   5.064082s
Typecast.to_number 1.1
                         11.503M (± 0.7%) i/s -     57.944M in   5.037778s
Typecast.to_number '1'
                          4.986M (± 0.4%) i/s -     25.126M in   5.039757s
Typecast.to_number '1.1'
                          4.805M (± 0.4%) i/s -     24.340M in   5.065364s
Typecast.to_number nil
                          7.612M (± 0.2%) i/s -     38.168M in   5.014263s
Typecast.to_number Time.now
                          5.895M (± 0.4%) i/s -     29.987M in   5.086817s
```


So with a `Numeric` it is significantly faster since it doesn't need to do anything, but definitely slower with a Sting. The problem I was trying to solve is that the current function converts `1.0` or `"1.0"` to the integer `1`. That feels wrong. I I can't think of a better way to solve it.

How often in the normal course of events are strings cast to numbers as it currently stands? Either way, I'm fine to revert this and deal with it later.

lib/flipper/expressions/percentage.rb
```diff
@@ -1,19 +1,8 @@
...
-      def evaluate(context = {})
-        value = super
-
-        value = 0 if value < 0
-        value = 100 if value > 100
-
-        value
+    class Percentage
+      def self.call(value)
+        value.to_f.clamp(0, 100)
```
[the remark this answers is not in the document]
bkeepers: Me too. Just happened to notice it in the docs the other day, so it was fresh in my mind

lib/flipper/expression/builder.rb
bkeepers: It would have to be `function.respond_to?(:group?) && function.group?` since each function is just a plain ol' class. 

With this PR I was hoping we could avoid functions being anything more than a single method so all the "expression" logic is condensed to a single class, but as we talked IRL earlier, it might not be achievable since we still need some kind of input validation or types. Adding to the builder here is another where I'd love to figure out how to make it extensible.

lib/flipper/expression/constant.rb
```diff
@@ -0,0 +1,25 @@
...
+    class Constant
+      include Expression::Builder
+
+      attr_reader :value
+
+      def initialize(value)
+        @value = value
+      end
+
+      def evaluate(_ = nil)
```
jnunemaker, line 15: Super minor: this could just be `-` right? `-` is usually to denote unused, so assigning something that is unused seems pointless.

lib/flipper/expressions/time.rb
```diff
@@ -0,0 +1,9 @@
+module Flipper
+  module Expressions
+    class Time
+      def self.call(value)
+        ::Time.parse(value)
```
jnunemaker, line 5: This will assume local time, not utc, won't it? I stumbled on [this stack overflow](https://stackoverflow.com/questions/4262550/how-do-i-get-ruby-to-parse-time-as-if-it-were-in-a-different-time-zone) when googling.

spec/flipper/expressions/now_spec.rb
```diff
@@ -0,0 +1,11 @@
+RSpec.describe Flipper::Expressions::Now do
+  describe "#call" do
+    it "returns current time" do
+      expect(described_class.call.round).to eq(Time.now.round)
```
jnunemaker, line 4: I usually use expect be_within here for when the second passes. 

```suggestion
      expect(described_class.call).to be_within(2).of(Time.now.utc)
```

jnunemaker (approved):
Looks great. I left a few minor things. 

My only concern is the time stuff. I think it's too liberal right now for multi-language. We'll probably want to lock it down and only accept strings in a particular format or formats.

lib/flipper/expression.rb
```diff
@@ -1,114 +1,66 @@
...
     private
 
-    def evaluate_arg(index, context = {})
-      object = args[index]
-
-      if object.is_a?(Flipper::Expression)
-        object.evaluate(context)
-      else
-        object
+    def call_with_context?
```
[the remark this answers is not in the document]
bkeepers: Yeah, I had the same thought.

lib/flipper/expressions/time.rb
bkeepers: Only if the string passed in excludes the timezone. In the [schemas I'm working on](https://github.com/fewerandfaster/expressions.js/blob/main/schemas/Time.schema.json), I explicitly defined `Time` to take a `date-time`, which JSON Schema defines as an ISO8601 formatted string. Here are the tests:

```
    Time
      valid
        ✓ {"Time":"2021-01-01T00:00:00Z"}
        ✓ {"Time":["2021-01-01T00:00:00Z"]}
        ✓ {"Time":"2021-01-01T00:00:00-05:00"}
        ✓ {"Time":["2021-01-01T00:00:00-05:00"]}
        ✓ {"Time":{"Property":"created_at"}}
        ✓ {"Time":[{"Property":"created_at"}]}
      invalid
        ✓ {"Time":"2021-01-01"}
        ✓ {"Time":"January 1, 2021 10:00"}
        ✓ {"Time":null}
        ✓ {"Time":false}
        ✓ {"Time":[{"Property":"created_at"},{"Property":"updated_at"}]}
```

lib/flipper/typecast.rb
jnunemaker: Sweet!

lib/flipper/expression/constant.rb
jnunemaker: Ah. Didn’t think of that. Ignore me.

PR #718, opened 2023-03-30
Schema for Expressions

bkeepers:
This introduces a schema based on [JSON Schema](https://json-schema.org) ([draft-07](https://json-schema.org/specification-links.html#draft-7) since it is the latest draft supported by Ruby libraries) for validating expressions (#692) and the arguments to expression functions. It also moves all of the tests into JSON file format so we they can be used when we add support for other clients.

- [ ] Document how to add a new function
- [ ] Document expression example format used in `packages/expressions/test/examples`
- [ ] Scripts/docs to release `@flippercloud.io/expressions` npm package
- [ ] Add method to validate an expression
- [ ] Validate expression before saving, e.g. `Flipper.enable_expression :feature, expression`


flipper-expressions-schema.gemspec
```diff
@@ -0,0 +1,34 @@
...
+  gem.homepage      = 'https://www.flippercloud.io/docs/expressions'
+
+  gem.files         = [
+    'package.json',
+    'package-lock.json',
+    'lib/flipper-expressions-schema.rb',
+    'lib/flipper/expression/schema.rb',
+    'lib/flipper/version.rb',
+  ] + Dir['node_modules/@flippercloud.io/expressions/schemas/*.json']
+  gem.name          = 'flipper-expressions-schema'
```
bkeepers, line 28: Should this be `flipper-expression-schemas`? The ruby class is `Flipper::Expression::Schema` to avoid conflicts with functions declared in `Flipper::Expressions::*`, but maybe those should be renamed to `Flipper::Function::*`. The JS package is [`@flippercloud.io/expressions`](https://github.com/flippercloud/expressions.js)

jnunemaker: Are you just wondering if it should be plural? I think matching to ruby is fine even if singular.

jnunemaker (approved):
I understand this at a high level. I'm not sure what it will feel like moving forward but the best way is to merge and find out.

Issue #719, opened 2023-04-03
Can't delete flags due to redis-namespace bug


jdelStrother:
I use Redis::Namespace to keep my flipper flags separate from the rest of my redis database, but this is currently introducing a bug where I can't remove any feature flags - it tries to remove the `flagname` key from redis rather than `flipper:flagname`.

This is due to a bug in redis-namespace - https://github.com/resque/redis-namespace/pull/223 - rather than being anything that Flipper is doing wrong, but thought it was worth highlighting here for anyone else running into the same problem.

---

As an aside, Flipper is the only reason we use redis-namespace. Would you be supportive of a PR that adds a `key_prefix` option into flipper-redis, so we can drop redis-namespace?

jnunemaker:
I'd be open to this.

jnunemaker:
redis namespace is no longer needed thanks to #722.

PR #722, opened 2023-04-16
Add a key_prefix option to the Redis adapter

jdelStrother:
We currently use on redis-namespace to keep Flipper's keys separate from the rest of our redis db, though that's led to some bugs (https://github.com/jnunemaker/flipper/issues/719). How about a `key_prefix` option on the redis adapter?

My original attempt at this was more wide-reaching, adding a KeyValueStore module that could be used by all the KV adapters and adding the key_prefix option to all of the simultaneously - WIP here: https://github.com/jnunemaker/flipper/compare/main...jdelStrother:flipper:key-lookups-v1
Backwards compatibility is a pain, though - the redis adapter would need to override a bunch of the shared module methods to make sure its keys stay the same, and I decided to switch to the much simpler approach in this PR. See which you prefer.


bkeepers (approved):
Looks good to me, just one bit of feedback and I think this can be merged.

jnunemaker (approved):
@bkeepers you want me to merge this or is there anything else?

jdelStrother:
Anything I can do to help with this?

bkeepers:
Thanks. We’ll get a release out soon

PR #723, opened 2023-04-19
Address `color` deprecation warning

seanpdoyle:
While using Flipper with `rails>7.0`, the following deprecation warning is output:

```
DEPRECATION WARNING: Bolding log text with a positional boolean is deprecated and will be removed in Rails 7.2. Use an option hash instead (eg. `color(:red, "my text", bold: true)`).
```

To address this issue, this commit makes the suggested changes to the `LogSubscriber` class.


jimmypoulsen:
Fixes #733.

PR #726, opened 2023-05-03
Use new method of making logs bold

piperswe:
Using a positional boolean is deprecated in Rails 7.1, and will be removed in Rails 7.2:

```
DEPRECATION WARNING: Bolding log text with a positional boolean is deprecated and will be removed in Rails 7.2. Use an option hash instead (eg. `color("my text", :red, bold: true)`). (called from feature_operation at /usr/local/bundle/ruby/3.2.0/gems/flipper-0.28.0/lib/flipper/instrumentation/log_subscriber.rb:35)
```


jnunemaker:
Will this work on older versions of rails or do we need to do this for new and true for old?

piperswe:
Sadly, looks like the new form was added in Rails 7.1. I'll make it change functionality based on Rails version.

Issue #730, opened 2023-05-17
Bundling bootstrap and poppler with the library ?


dhruvCW:
I am prototyping flipper for my organisation, we are running into a bit of a issue with our Content Security Policy. Would it be possible for the library to bundle poppler, bootstrap and jquery ? This will then not require us to extend our CSP. Our CSP is configured site wide and we currently are not in a position to setup per route CSP right now, if the dependencies are bundled we won't need to extend the CSP.

jnunemaker:
@bkeepers would know more about this. Seems ok if it’s easy. We’d just need a middleware to serve static files, right?

dhruvCW:
I think the existing `UI::Action::File` module can be utilized. Infact just dropping the files under the right public folder should be enough. If this is something that the project would like to do, I can create a PR with the suggested changes.

dhruvCW:
Would a new release containing this change be created soon ?

PR #731, opened 2023-05-19
Bundle bootstrap, jquery and poppler with the library.

dhruvCW:
fixes https://github.com/jnunemaker/flipper/issues/730


thetimbanks:
One recommendation I would have on this is to include the version number in the filenames.

jnunemaker:
Looks good. I'll pull it down and test it soon so I can get it merged. Thanks!

PR #736, opened 2023-06-06
Improve performance of `color_name` function in log subscriber

natematykiewicz:
This fixes a performance degradation introduced in https://github.com/jnunemaker/flipper/pull/726

Move runtime version comparison into file-load comparison, since the gem version does not change while the process is running.

This removes two Gem::Version.new allocations (which in turn does some regex scans and string allocations), as well as a Hash allocation (`{ bold: true }` on Rails 7.1+) per `color_name` call.

### DIFF
--- a/lib/flipper/instrumentation/log_subscriber.rb
+++ b/lib/flipper/instrumentation/log_subscriber.rb
@@ -73,14 +73,17 @@ module Flipper
 
       private
 
+      # Rails 7.1 changed the signature of this function.
+      # Checking if > 7.0.99 rather than >= 7.1 so that 7.1 pre-release versions are included.
+      COLOR_OPTIONS = if Rails.gem_version > Gem::Version.new('7.0.99')
+        { bold: true }.freeze
+      else
+        true
+      end
+      private_constant :COLOR_OPTIONS
+
       def color_name(name)
-        # Rails 7.1 changed the signature of this function.
-        # Checking if > 7.0.99 rather than >= 7.1 so that 7.1 pre-release versions are included.
-        if Rails.gem_version > Gem::Version.new('7.0.99')
-          color(name, CYAN, bold: true)
-        else
-          color(name, CYAN, true)
-        end
+        color(name, CYAN, COLOR_OPTIONS)
       end
     end
   end


jnunemaker:
Thanks! 

I’ve been giving www.flippercloud.io credits to people who contribute, so if you want to check it out beyond the trial, let me know and I’ll spot you several months or more.

Lastly, if you don’t mind, I would love to know what you are using flipper for as well and how it is working out for you (what went well, what was hard other than what you mentioned here).

If you aren’t comfortable leaving that information here in a public comment, feel free to email me directly john@fewerandfaster.com. I’m just always curious. :)

natematykiewicz:
Hey John!

[Dwell](https://dwellapp.io) is using Flipper to do a site redesign (different colors and fonts, switching from Turbolinks to Turbo, etc).

So I've got 2 Flipper Features:

One determines who gets to see a button on their account page to toggle between this pink site you see here and the new design, and the second one determines who currently has the new design enabled. Using Devise and some new controllers, I'm actually able to handle the feature Flipping inside the router.

```ruby
authenticated :user, ->(user) { Flipper.enabled?(:show_dwell_3_layout, user) } do
  scope module: :dwell3 do
    resources :foos
  end

  resources :foos
end
```

Devise has an `authenticate` method and `authenticated` method. That subtle `d` is a big deal. `authenticate` will prompt for login and 404 if you are signed in but don't match the block, which we use on admin pages (such as the flipper dashboard). `authenticated` simply falls through the route if you either aren't signed in or are don't match the block. So here we're simply defining a higher priority route that you may or may not match depending on whether you're signed in and have the flipper feature enabled.

Now, `FoosController` is the pink layout which uses Turbolinks / Coffeescript / jQuery / SCSS. `Dwell3::FoosController` (namespacing the redesign and all controllers from here on out 😅) is the new layout, which uses Turbo / StimulusJS / Tailwind. 

Flipper is making it really easy to manage this redesign project.

We've also slightly used it to test out changes that might have a negative performance impact (basically A/B testing), since it can be hard to measure the full impact of a change on anything other than the production database.

I should really look into Flipper Cloud more. We've used feature flags so infrequently so far that I hadn't really considered it. We should probably use more feature flags.

The only other thing that's been tricky was something I mentioned (and you saw) in the Rails Performance Slack. I found that Flipper's memoizer preloads Features on all requests, and so I needed an initializer to cut down on unnecessary SQL queries from routes that for sure don't use Flipper, such as `/rails/active_storage`, `/cable`, our health check, and the routes that I've mounted Flipper, Sidekiq, PgHero, and Lookbook at. For those last 4 the big reason to skip was because the preloaded ran on asset requests (js, css, images) for those 4 Rails Engines. I wouldn't have really cared so much if it were just one on the page.

```ruby
#
# Skip preloading Flipper Gates/Features for certain paths:
# This avoids an unused query like:
#   SELECT "flipper_features"."key" AS feature_key, "flipper_gates"."key", "flipper_gates"."value"
#   FROM "flipper_features"
#   LEFT OUTER JOIN "flipper_gates" ON "flipper_features"."key" = "flipper_gates"."feature_key"
  skippable_paths = %w[
    /cable
    /rails/active_storage
    /lookbook-assets

    # health check path

    # Flipper mount path
    # Sidekiq mount path
    # Pg Hero mount path
    # Lookbook mount path
  ].freeze

  config.flipper.memoize = -> (request) do
    skippable_paths.none? do |prefix|
      request.path.start_with?(prefix)
    end
  end
end
```

PR #737, opened 2023-06-15
Human readable actor names

jevin:
This makes actor names more readable. Made sure there were no N+1 queries like @jnunemaker suggested!

I also wrote a deserializer to convert actor identifiers to ActiveRecord fields:
```ruby
def self.humanize_identifiers(keys, field)
  grouped_keys = {}
  actor_names = {}

  keys.each do |key|
    parts = key.split(";")
    grouped_keys[parts.first] ||= []
    grouped_keys[parts.first] << parts.second
  end

  grouped_keys.each do |model, ids|
    model.constantize.where(id: ids).find_each do |resource|
      actor_names["#{resource.class.name};#{resource.id}"] = resource.send(field)
    end
  end

  actor_names
end
```

I'm not sure how that can fit in the project, apart from adding it to the README. Suggestions appreciated.


jnunemaker:
@jevin this looks great! Nice work. I'll get it merged and released soon.

PR #740, opened 2023-07-17
Avoid any possibility of sync writing back to remote

jnunemaker:
The goal is for sync to trust the remote and overwrite local with remote. 

The problem is that using dual write adapter for poll adapter in cloud means that a sync could write back to remote (cloud). If we use local adapter for poll then a sync can only affect local and never touch remote. 

Then we can use dual write with poll adapter as the read and cloud as the remote to ensure that local Flipper.enable calls (and similar) will still go to both local and write.


bkeepers (approved):
👍 🧷

PR #743, opened 2023-07-18
Merge flipper-cloud into flipper gem

bkeepers:
This merges `flipper-cloud` gem into `flipper` and only configures cloud if the `FLIPPER_CLOUD_TOKEN` is set.

Alternative to #739


spec/flipper/engine_spec.rb
```diff
@@ -106,6 +108,82 @@
...
+      { "FLIPPER_CLOUD_TOKEN" => "test-token" }
+    end
+
+    # App for Rack::Test
+    let(:app) { application.routes }
+
+    it "initializes cloud configuration" do
+      stub_request(:get, /flippercloud\.io/).to_return(status: 200, body: "{}")
+
+      ENV.update(env)
```
jnunemaker, line 122: Does this need to be rolled back after spec is done?

jnunemaker: I've been using https://github.com/thoughtbot/climate_control for modifying env of late. Its nice.

jnunemaker:
I think overall this looks good. We can release as rc1 for 0.29 or something to gain confidence that we did it right without messing people up. I can update boxoutsports.com, thewatcharchive.com and southbendwatch.co to see how it works.

spec/flipper/engine_spec.rb
bkeepers: Good call. I added climate_control and updated everywhere else that we were modifying ENV in 4e4decb

PR #755, opened 2023-08-30
Add #enabled? to ActiveRecord models

bkeepers:
Adds `#enabled?` to all ActiveRecord models as shorthand for `Flipper.enabled?(:feature_name, model)`…with a bonus…

One thing that has always felt awkward is that I'm never quite sure who/what to use as an actor when building a new feature. Is it user-centric? Or some other other model like `Organization` or `Project`? In #710 we added the ability to check multiple actors at once, which is great, but it's still awkward to always have to have `Flipper.enabled?(:feature, current_user, *current_user&.organizations)` sprinkled throughout the code.

With this change, you can define `#flipper_actors` on any model, and calling `#enabled?` will check the feature against the relevant actors.

```ruby
class User < ApplicationRecord
  has_many :memberships
  has_many :organizations, through: :memberships

  def flipper_actors
    [self] + organizations
  end
end
```

So now `current_user.enabled?(:new_feature)` will check if it's enabled for a specific user or any organization that they are part of.

### DIFF
--- a/lib/flipper/model/active_record.rb
+++ b/lib/flipper/model/active_record.rb
@@ -7,6 +7,25 @@ module Flipper
       def flipper_properties
         {"type" => self.class.name}.merge(attributes)
       end
+
+      # Check if a feature is enabled for this record or any associated actors
+      # returned by `#flipper_actors`
+      def enabled?(feature_name)
+        Flipper.enabled?(feature_name, *flipper_actors)
+      end
+
+      # Returns the set of actors associated with this record. Override to
+      # return any other records that should be considered actors.
+      #
+      #   class User < ApplicationRecord
+      #     # …
+      #     def flipper_actors
+      #       [self, company]
+      #     end
+      #   end
+      def flipper_actors
+        [self]
+      end
     end
   end
 end
--- a/spec/flipper/model/active_record_spec.rb
+++ b/spec/flipper/model/active_record_spec.rb
@@ -45,4 +45,34 @@ RSpec.describe Flipper::Model::ActiveRecord do
       })
     end
   end
+
+  describe "enabled?" do
+    subject { User.create!(name: "Test") }
+
+    module Friendable
+      attr_accessor :friends
+
+      def flipper_actors
+        [self] + Array(friends)
+      end
+    end
+
+    it "returns false if feature is disabled" do
+      expect(subject.enabled?(:stats)).to be(false)
+    end
+
+    it "returns true if feature is enabled for actor" do
+      Flipper.enable :stats, subject
+      expect(subject.enabled?(:stats)).to be(true)
+    end
+
+    it "returns true if feature is enabled for associated actor" do
+      friend = User.create!(name: "Friend")
+      subject.extend Friendable
+      subject.friends = [friend]
+
+      Flipper.enable :stats, friend
+      expect(subject.enabled?(:stats)).to be(true)
+    end
+  end
 end


lib/flipper/model/active_record.rb
```diff
@@ -7,6 +7,25 @@ module ActiveRecord
       def flipper_properties
         {"type" => self.class.name}.merge(attributes)
       end
+
+      # Check if a feature is enabled for this record or any associated actors
+      # returned by `#flipper_actors`
+      def flipper_enabled?(feature_name)
```
jnunemaker, line 13: One interesting thing with this, you could have user and org. In user you override flipper_actors but in org you can't because you don't know the specific user. I think that is part of my concern with adding flipper_actors and this method name (even though it is now scoped). 

I want the convenience but also don't want it to be too *sharp* for users.

bkeepers: > One interesting thing with this, you could have user and org. In user you override flipper_actors but in org you can't because you don't know the specific user.

I'm not sure I understand. If you have access to a user, you would check against that instead of the organization.

```ruby
current_user.flipper_enabled?(:thing)
```

If not, then you check against just the organization:

```ruby
current_org.flipper_enabled?(:thing)
```

Either way, enabling the feature for the organization would make both checks pass:

```ruby
Flipper.enable(:thing, current_org)
```

Every feature check would benefit from the user knowing it's related actors.

> I want the convenience but also don't want it to be too _sharp_ for users.

By default the behavior doesn't change. It's only if you explicitly overwrite `#flipper_actors`.

jnunemaker:
> I'm not sure I understand.

I just meant if you have this:

```ruby
class User < ApplicationRecord
  has_many :memberships
  has_many :organizations, through: :memberships

  def flipper_actors
    [self] + organizations
  end
end

class Organization < ApplicationRecord
  def flipper_actors
    [self] # no way to add user here so its the same multi actor as user
  end
end

user = User.find(...) # user is member of org
org = Org.find(...)
Flipper.enable(:foo, org)

user.feature_enabled?(:foo) # => true
org.feature_enabled?(:foo) # => true
```

Both of those work. But they are not symmetrical. If you enable user instead of org the result is different. 

```ruby
Flipper.disable(:foo, org)
Flipper.enable(:foo, user)

user.feature_enabled?(:foo) # => true
org.feature_enabled?(:foo) # => false
```

Looking at it, it only works as hierarchy and when using the lowest level. But if you check from higher levels you'll have to really know what's going on. 

Maybe because users are a very common actor and usually the lowest level in a given hierarchy this will end up fine?

bkeepers:
> Looking at it, it only works as hierarchy and when using the lowest level. But if you check from higher levels you'll have to really know what's going on.

Gotcha, I guess I see that as desired behavior. I often want to enable a feature for a`User` (myself or specific users) before other actor models like `Project` or `Organization`. I also expect that enabling a feature for an `Organization` will enable it for all users in that organization.

jnunemaker:
Interesting! Let’s decide on it next week. Only other concern is one more thing to search for when finding flag usage.

jnunemaker:
Seeing #783 made me think if we do this PR we could also add flipper_features to return all enabled features for an actor or a Hash of all features with key pointed at true/false.

jnunemaker:
@bkeepers I was going through old PRs today. I say we merge this and see if people like it.

lib/flipper/model/active_record.rb
pbstriker38: If the feature is enable for 1 of the users organizations, then wouldn't they see it enabled on all of their organizations? That doesn't seem correct.

PR #759, opened 2023-09-07
Allow head requests to api

jnunemaker:
@bkeepers @garrettdimon any downside to doing it this way? I mostly just want to make `curl -Is` work with api requests.


bkeepers (approved):
Not that I can think of.

PR #760, opened 2023-09-16
Add Strict adapter to ensure feature exists on get

bkeepers:
First pass at adding a strict adapter to ensure only features that exist can be called.

```ruby
Flipper.configure do |config|
  config.adapter { Flipper::Adapters::Strict.new(Flipper::Adapters::ActiveRecord.new) }
end

Flipper.enabled?(:featuere_name) # raises Flipper::Adapters::Strict::NotFound

Flipper.add(:feature_name)

Flipper.enabled?(:feature_name) # => false
```

In a follow up PR I'm hoping to use this by default in Rails apps in development/test.


lib/flipper/adapters/strict.rb
```diff
@@ -0,0 +1,44 @@
...
+        end
+      end
+
+      HANDLERS = {
+        raise: ->(feature) { raise NotFound.new(feature.key) },
+        warn: ->(feature) { warn NotFound.new(feature.key).message },
+        noop: ->(_) { },
+      }
+
+      DEFAULT_NOT_FOUND =
```
jnunemaker, line 21: Is this magic or missing?

lib/flipper/adapters/strict.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      def_delegators :@adapter, :features, :get, :get_multi, :get_all, :add, :remove, :clear, :enable, :disable
```
jnunemaker, line 23: Since get is defined below, it shouldn't be in here, right? 

Also, should get_multi be defined below so reads for it also get strictly checked?

lib/flipper/adapters/strict.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      def initialize(adapter, name: :strict, handler: :raise)
```
jnunemaker, line 25: I don't usually allow changing the adapter name. But seeing it makes me wonder if we should allow this for people as I could see scenarios where someone wants to override for instrumentation purposes.

jnunemaker (approved):
Looks good. Left a couple questions. I'd think get_multi should be in so there is a raise/warning for preloading the wrong feature name.

lib/flipper/adapters/strict.rb
bkeepers: Doh, that is unused now. Funny that it wasn't raising an error.

lib/flipper/adapters/strict.rb
bkeepers: Removed name param for now. I actually had some changes locally that adds `Flipper::Adapter#name`, which just defaults to class name symbolized. Seems like most adapters just explicitly assign `@name` and an `attr_reader`

lib/flipper/adapters/strict.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      HANDLERS = {
+        raise: ->(feature) { raise NotFound.new(feature.key) },
+        warn: ->(feature) { warn NotFound.new(feature.key).message },
+        noop: ->(_) { },
+      }
+
+      def_delegators :@adapter, :features, :get_all, :add, :remove, :clear, :enable, :disable
+
+      def initialize(adapter, handler: :raise)
```
bkeepers, line 23: I don't love `handler`, but can't come up with anything better. `on_missing`? `behavior`? Drop the named arg entirely and just have it be the 2nd arg?

garrettdimon: I like `on_missing` because that's what the Strict adapter is entirely about. Then it defaults to raise, or if we include it by default, it can default to warn?

garrettdimon (commented):
Are you envisioning that this would be used like in the example or would it also have some level of configuration? i.e. Is the configuration happening purely from how the initializer is called?

bkeepers:
> Are you envisioning that this would be used like in the example or would it also have some level of configuration? i.e. Is the configuration happening purely from how the initializer is called?

Still in progress, but this is what I am imagining: https://github.com/flippercloud/flipper/pull/763

lib/flipper/adapters/strict.rb
bkeepers: I decided to just remove the keyword argument and use the second param as the handler name, or a block passed to the initialize method. My reasoning is that I think it's unlikely that this adapter will take other params, and we can always add this as keyword params if needed.

PR #761, opened 2023-09-18
Use Rails global constant

RStankov:
When I updated my application to `Ruby 3.2.2` I start getting the following error

```
 undefined method `gem_version' for Flipper::Rails:Module (NoMethodError)
```

From this line

```ruby
COLOR_OPTIONS = if Rails.gem_version > Gem::Version.new('7.0.99')
```

The reason is that `Rails` in this context is mistaken for `Flipper::Rails` instead of the global `Rails`.

### DIFF
--- a/lib/flipper.rb
+++ b/lib/flipper.rb
@@ -200,4 +200,4 @@ require 'flipper/types/percentage_of_time'
 require 'flipper/typecast'
 require 'flipper/version'
 
-require "flipper/engine" if defined?(Rails)
+require "flipper/engine" if defined?(::Rails)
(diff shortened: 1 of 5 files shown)


jnunemaker:
Weird that it would do that. Doesn't seem to be happening for me on 3.2.2. That said, I don't think this can hurt anything and maybe it helps. 

@bkeepers any thoughts?

bkeepers:
@RStankov Are you by chance using the old `flipper-rails` gem? We don't define `Flipper::Rails` anywhere, so I'm not sure where else that would be coming from.

RStankov:
Yes, I'm using `flipper-rails`. 

Is it not needed anymore?

bkeepers:
@RStankov Great! `flipper-rails` was not an official gem and is deprecated.

bkeepers:
Closing this since I don't think it's needed, but let me know if anyone feels otherwise.

PR #762, opened 2023-09-20
Define default Adapter#name

bkeepers:
I noticed while working on #760 that every adapter defines `#name` and it's just a variable set in initialize. This PR added a default `Adapter#name` method, which is just the `snake_case` version of the `ClassName`.

In https://github.com/flippercloud/flipper/pull/760#discussion_r1328013132, @jnunemaker said:

> I don't usually allow changing the adapter name. But seeing it makes me wonder if we should allow this for people as I could see scenarios where someone wants to override for instrumentation purposes.

Currently, the ActiveRecord and Sequel adapters are the only ones that allow setting the name. Those will still work with this change. Should we allow others?

### DIFF
--- a/lib/flipper/adapter.rb
+++ b/lib/flipper/adapter.rb
@@ -64,6 +64,11 @@ module Flipper
     def default_config
       self.class.default_config
     end
+
+    # Public: default name of the adapter
+    def name
+      @name ||= self.class.name.split('::').last.split(/(?=[A-Z])/).join('_').downcase.to_sym
+    end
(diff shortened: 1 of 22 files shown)


jnunemaker (approved):
I think this should work. As long as the names are the same for anyone instrumenting we should be fine. If the magic changed any names, we'd want to let people know in changeling as metric names would change and they'd have to update graphs.

bkeepers:
👍 All names should be the same.

PR #768, opened 2023-10-05
Use more-efficient Redis `sismember` command to check if feature exists

owst:
We have a fairly large number of features backed by a Redis adapter and use something like that in the discussion in https://github.com/flippercloud/flipper/issues/354 to set default feature values.  I spotted that the `exist?` check `smembers` command was showing up in our [Redis slowlog](https://redis.io/commands/slowlog-get/).

We can use [`sismember` ](https://redis.io/commands/sismember/) instead of pulling back all feature names with [`smembers`](https://redis.io/commands/smembers/) and checking for inclusion. This should go from O(N) (N is the number of features) to O(1), for a modest performance improvement.

### DIFF
--- a/lib/flipper/adapter.rb
+++ b/lib/flipper/adapter.rb
@@ -43,6 +43,13 @@ module Flipper
       result
     end
 
+    # Public: Does a feature with this key exist.
+    #
+    # Returns true if a feature exists with this key, else false.
+    def exist?(key)
+      features.include?(key)
+    end
+
     # Public: Ensure that adapter is in sync with source adapter provided.
     #
     # source - The source dsl, adapter or export to import.
--- a/lib/flipper/adapters/redis.rb
+++ b/lib/flipper/adapters/redis.rb
@@ -32,6 +32,11 @@ module Flipper
         read_feature_keys
       end
 
+      # Public: Does a feature with this key exist.
+      def exist?(key)
+        @client.sismember(features_key, key)
+      end
+
       # Public: Adds a feature to the set of known features.
       def add(feature)
         if redis_sadd_returns_boolean?
--- a/lib/flipper/feature.rb
+++ b/lib/flipper/feature.rb
@@ -79,7 +79,7 @@ module Flipper
     #
     # Returns true if exists in adapter else false.
     def exist?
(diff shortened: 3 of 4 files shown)


lib/flipper/adapter.rb
```diff
@@ -43,6 +43,13 @@ def get_multi(features)
       result
     end
 
+    # Public: Does a feature with this key exist.
+    #
+    # Returns true if a feature exists with this key, else false.
+    def exist?(key)
+      features.include?(key)
```
owst, line 50: Add a default implementation on the base adapter, that we can override in the Redis adapter

bkeepers:
@owst thanks for the pull request!

One question I have just for context: is there a use case where you're often checking if a feature exists and _not_ evaluating the result of `enabled?`, or checking for a lot of features that don't exist? Just curious, because in what I presume is the more common case of checking if a feature exists and then evaluating it, this will add two network calls where there was just one previously.

I'm not opposed to merging this, I just want to understand the usage patterns.

owst:
@bkeepers thanks for taking a look!

In short: no, I think that we would always evaluate `enabled?`.  I think our only use case is like this:
```ruby
def feature_enabled?(feature_name)
  feature = flipper[feature_name]
  
  return true if feature.enabled?
  
  # ... is it false because the feature is undefined, or false because the feature is disabled?
  return false if feature.exists? 
  
  enabled_by_default = feature_enabled_by_default?(feature_name) 
  
  enabled_by_default ? feature.enable : feature.disable
  
  enabled_by_default
end
```

which has the intended behaviour of setting the feature default the first time it is accessed.

Does this make sense? And does the PR make sense in this context?

dpep:
my two cents here is that enabling a non-existent flag as a default behavior is novel and perhaps dangerous.  defaulting to false/off creates a safer, failed-closed approach

it might be better (and simpler) to simply invert your flags when appropriate, eg.

```ruby
return if Flipper[:disable_feature].enabled?

# do feature thing
```

which would default to showing the feature

jnunemaker:
Closing this because I think this will help one use case but hurt others. But keep communicating here as I'll make sure we get you all settled and performing well.

PR #771, opened 2023-11-05
preload via block

jnunemaker:
This enhances preload to support customizing preloading per request. Should solve the problems in https://github.com/flippercloud/flipper/discussions/729#discussioncomment-5876084.


bkeepers (approved):
👍 looks good

PR #772, opened 2023-11-13
Hide UI elements when Flipper is in read-only mode

radar:
We have a particular business-critical application that uses Flipper flags, and we would like to provide a completely read-only UI with Flipper. I know of a way to do this, which is to set `Flipper::UI`'s configuration to be read only. However, this same application is a bit a "nexus" for some smaller apps, and therefore we have multiple instances of `Flipper::UI` at play here. The different instances each control the flipper flags for the different applications. 

I was able to see that there's a `ReadOnly` adapter however:

```
readonly_app = Flipper::UI.app(Flipper.new(Flipper::Adapters::ReadOnly.new(connection)))
```

But this doesn't hide all the UI buttons in Flipper UI. Here's what I see in that app by default:

image

Having all the buttons and forms present makes our security people uncomfortable.

---

So what I've tried to do in this PR is to cut out all of the UI that performs a write action, which then cuts the page down to this:

image

And the homepage to this:

image

I've also updated the "nooooope" error screen to show that the `ReadOnly` adapter might be in use:

image

### DIFF
--- a/lib/flipper/ui/actions/actors_gate.rb
+++ b/lib/flipper/ui/actions/actors_gate.rb
@@ -23,7 +23,7 @@ module Flipper
         end
 
         def post
-          read_only if Flipper::UI.configuration.read_only
+          read_only if read_only?
 
           feature = flipper[feature_name]
           value = params['value'].to_s.strip
--- a/lib/flipper/ui/actions/actors_gate.rb
+++ b/lib/flipper/ui/actions/actors_gate.rb
@@ -23,7 +23,7 @@ module Flipper
         end
 
         def post
-          read_only if read_only?
+          render_read_only if read_only?
 
           feature = flipper[feature_name]
           value = params['value'].to_s.strip
--- a/lib/flipper/ui/views/features.erb
+++ b/lib/flipper/ui/views/features.erb
@@ -3,7 +3,7 @@
     <% if Flipper::UI.configuration.fun %>
       <h4>But I've got a blank space baby...</h4>
       <p>And I'll flip your features.</p>
-      <%- if !Flipper::UI.configuration.read_only && Flipper::UI.configuration.feature_creation_enabled -%>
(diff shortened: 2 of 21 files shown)


radar:
Thanks for making the Flipper code so easy to read and reason about. It was a pleasure to contribute :)

lib/flipper/ui/actions/actors_gate.rb
```diff
@@ -23,7 +23,7 @@ def get
         end
 
         def post
-          read_only if Flipper::UI.configuration.read_only
+          read_only if read_only?
```
jnunemaker, line 26: Should this be render_read_only also?

lib/flipper/ui/views/features.erb
```diff
@@ -3,15 +3,16 @@
     <% if Flipper::UI.configuration.fun %>
       <h4>But I've got a blank space baby...</h4>
       <p>And I'll flip your features.</p>
-      <%- if !Flipper::UI.configuration.read_only && Flipper::UI.configuration.feature_creation_enabled -%>
+      <%= read_only? %>
```
jnunemaker, line 6: Did you mean to leave this?

spec/flipper/ui/actions/add_feature_spec.rb
```diff
@@ -39,4 +39,24 @@
       expect(last_response.body).to include('Feature creation is disabled.')
     end
   end
+
+  describe 'GET /features/new when an adpter is set to readonly' do
```
jnunemaker, line 43:
```suggestion
  describe 'GET /features/new when an adapter is set to readonly' do
```

jnunemaker (commented):
Looks great. I left a couple questions. If you have time to answer/address go for it. If you don't, just let me know and I'll get this over the finish line.

radar:
1. Removed the leftover view code.
2. Fixed up the add feature spec which wasn't actually using the `read_only?` query method.
3. Added a regression test for ActorsGate, and then fixed the code.

PR #776, opened 2023-11-30
Remove support for EOL Ruby 2.6/2.7, Rails 5.2/6.0

bkeepers:
I propose we drop support for EOL Ruby and Rails versions. ~~Any reservations about doing this now before the upcoming 1.1 release?~~ We will wait until the 2.0 release.

EOL dates:
Ruby 2.6 - 31 Mar 2022 - 20 months ago
Rails 5.2 - 01 Jun 2022 - 18 months ago
Ruby 2.7 - 31 Mar 2023 - 8 months ago
Rails 6.0 - 01 Jun 2023 - 6 months ago


dpep:
I'm a big fan of cleanups and simplifications!  Ruby 2.6 and Rails 5 can definitely go.  I'm more on the fence about Ruby 2.7

can we add the constraint to [the gemspec](https://github.com/flippercloud/flipper/blob/main/flipper.gemspec) (`gem.required_ruby_version = ">= 3"`) and do a major version bump to communicate the breaking change?

bkeepers:
> can we add the constraint to [the gemspec](https://github.com/flippercloud/flipper/blob/main/flipper.gemspec) (`gem.required_ruby_version = ">= 3"`) and do a major version bump to communicate the breaking change?

Yes, good suggestion. I think I'll let this sit for now because we do still have some Flipper Cloud customers using Ruby 2.7. I added framework stats in #780 so we can see if anyone is using old Rails versions, and that will go out in 1.1 hopefully this week.

jnunemaker:
Closing this for now. I'll come back to it at some point when these versions are causing pain.

PR #778, opened 2023-12-02
statsd cleanup

dpep:
I was poking around and discovered the statsd subscriber.  In addition to being excited to test it out in prod, I noticed a couple small improvements we might make to simplify - using the `&.` operator to guard against nil objects, and adding test coverage...  what'd you think?


dpep:
yeah, if someone explicitly plugs in the statsd subscriber but doesn't provide a client, failing loudly makes sense to me.  I just didn't want to change the functionality on my first pass...

we could also expose something in the config dsl that makes setup easier, eg.
```
Flipper.configure do |conf|
  conf.statsd = my_client
end
```

PR #779, opened 2023-12-04
statsd config

dpep:
following up on our discussion in https://github.com/flippercloud/flipper/pull/778 to enable

```ruby
Flipper.configure do |conf|
  conf.statsd = my_client
end
```

this would simplify statsd integration and abstract away the subscriber logic.  I don't love the embedded requires and have an idea to rework that if we'd like.  we may want to put `ActiveSupport::Notifications.subscribe` into a method and pair it with a way to unsubscribe, so that it can be done multiple times (especially for testing)

### DIFF
--- a/lib/flipper/configuration.rb
+++ b/lib/flipper/configuration.rb
@@ -54,5 +54,15 @@ module Flipper
         @default.call
       end
     end
+
+    def statsd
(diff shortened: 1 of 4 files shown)


jnunemaker (approved):
Works for me. Thanks for making this easier for people.

PR #780, opened 2023-12-04
Include client framework versions in API requests

bkeepers:
Sends Rails, Sinatra, and Hanami versions so we can prioritize which versions to support.

cc #776


jnunemaker (approved):
Love it. 

Only thing I’d maybe change is add rescue around the method and constants inside the block in case they change but I imagine they are pretty stable.

PR #782, opened 2023-12-06
Load Cloud secrets from Rails credentials

bkeepers:
This allows setting Flipper Cloud secrets via [Rails credentials](https://edgeguides.rubyonrails.org/security.html#custom-credentials)

```
$ rails credentials:edit
```

```
flipper:
  cloud_token: <your-cloud-token>
  cloud_sync_secret: <your-webhook-secret>
```

### DIFF
--- a/lib/flipper/engine.rb
+++ b/lib/flipper/engine.rb
@@ -23,6 +23,10 @@ module Flipper
     end
 
     initializer "flipper.default", before: :load_config_initializers do |app|
+      # Load cloud secrets from Rails credentials
+      ENV["FLIPPER_CLOUD_TOKEN"] ||= app.credentials.dig(:flipper, :cloud_token)
+      ENV["FLIPPER_CLOUD_SYNC_SECRET"] ||= app.credentials.dig(:flipper, :cloud_sync_secret)
+
       require 'flipper/cloud' if cloud?
 
       Flipper.configure do |config|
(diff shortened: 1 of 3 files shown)


lib/flipper/engine.rb
```diff
@@ -23,6 +23,10 @@ class Engine < Rails::Engine
     end
 
     initializer "flipper.default", before: :load_config_initializers do |app|
+      # Load cloud secrets from Rails credentials
+      ENV["FLIPPER_CLOUD_TOKEN"] ||= app.credentials.dig(:flipper, :cloud_token)
```
garrettdimon, line 27: I assume we're comfortable assuming all instances of `app` will have `credentials`? I couldn't find any official details on which versions of Rails are supported. Maybe worth making this more resilient just to be safe?

garrettdimon (approved):
Great idea to have it proactively check credentials!

lib/flipper/engine.rb
bkeepers: I updated the test to actually write the credentials instead of stubbing them just to make sure.

jnunemaker:
Do app credentials already namespace by environment? Just curious.

bkeepers:
> Do app credentials already namespace by environment? Just curious.

Yep! There are separate credentials files for each environment, e.g. `rails credentials:edit -e test`

jnunemaker:
That's awesome.

jnunemaker:
I guess the only issue would be how people get their personal environment configured? Or is that handled somehow too? Not a blocker at all just curious because I haven't used credentials.

PR #783, opened 2023-12-07
Add #features_for_actor to return all enabeld features for a given actor

MagoMathew:
Adds #features_for_actor

With this change, you can get all features for the given actor as a param.

I've found we need that feature on my team,  where we just load all the features once.

Please feel free to share your comments and thoughts,
Probably there's a better way to do it.

Thanks for the great work.

### DIFF
--- a/spec/flipper/dsl_spec.rb
+++ b/spec/flipper/dsl_spec.rb
@@ -171,6 +171,30 @@ RSpec.describe Flipper::DSL do
       end
     end
   end
+  describe '#features_for_actor' do
+    actor = Flipper::Actor.new(5)
+    context 'with no features' do
+      it 'defaults to empty set' do
+        expect(subject.features_for_actor(actor)).to eq(Set.new)
+      end
+    end
+
+    context 'with features enabled and disabled' do
+      before do
+        subject.enable_actor(:stats, actor)
+        subject.enable_actor(:cache, actor)
+        subject[:search].disable
+      end
+
+      it 'returns set of feature instances' do
+        expect(subject.features_for_actor(actor)).to be_instance_of(Set)
+        subject.features_for_actor(actor).each do |feature|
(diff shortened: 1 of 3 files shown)


MagoMathew:
@jnunemaker
My approach could not be the most performant way to do what the PR does, so please feel free to suggest/apply any changes to make it work better.

I'm sure there's a way to get all the enabled features for a given actor with a single call to the DB, instead of calling one by one the features with the `enabled?` method.

spec/flipper/dsl_spec.rb
```diff
@@ -171,6 +171,30 @@
       end
     end
   end
+  describe '#features_for_actor' do
```
jnunemaker, line 174:
```suggestion

  describe '#features_for_actor' do
```

jnunemaker:
@MagoMathew thanks! This looks good. I've been mostly off for the holidays. Sorry for slow response. 

Only thing I'm trying to decide on is if we should roll with what you have or maybe something like `Flipper.features.enabled?(...)`. I'd have to make a proxy for that, so probably not worth it. 

The other thought is maybe something that does Flipper.features_for_actor(...) but returns a Hash with each feature key as the key and true/false as the value. 

@bkeepers any feelings on this method name or other ideas? If not, let me know and we can just merge as is.

skatkov:
How would this implementation work with active_record adapter? Would it fire off multiple queries to fetch all supported features? or will be done in one request?

jnunemaker:
@skatkov if preloaded it would do minimum network calls. If not preloaded then it would be n+1. 

I think we could move this to adapter level as well and allow each adapter to do this more efficiently (when necessary).

skatkov:
@jnunemaker would be wonderful if you can merge this PR, this can potentially solve performance issues in our app (with changes in adapter). Any chance you're planning to approve/merge this anytime soon?

PR #786, opened 2023-12-11
Fix AR adapter when table doesn't exist

jnunemaker:
This only happens when you haven't migrated yet likely. 

xref https://github.com/flippercloud/flipper/issues/557#issuecomment-1850350022

### DIFF
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -278,7 +278,7 @@ module Flipper
       # See https://github.com/flippercloud/flipper/pull/692
       def value_not_text?
         @gate_class.column_for_attribute(:value).type != :text
-      rescue ::ActiveRecord::AdapterError => error
+      rescue ::ActiveRecord::ActiveRecordError => error
         # If the table doesn't exist, the column doesn't exist either
         warn "#{error.message}. You likely need to run `rails g flipper:active_record` and/or `rails db:migrate`."
       end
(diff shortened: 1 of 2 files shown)


bkeepers (approved):
:bow: thanks for fixing

lib/flipper/adapters/active_record.rb
```diff
@@ -278,6 +278,9 @@ def result_for_gates(feature, gates)
       # See https://github.com/flippercloud/flipper/pull/692
       def value_not_text?
         @gate_class.column_for_attribute(:value).type != :text
+      rescue ::ActiveRecord::StatementInvalid => error
```
billkauf, line 281: We're also seeing `ActiveRecord::ConnectionNotEstablished` when building a docker image. 

(I changed my initializer to just blow off all `ActiveRecord::AdapterError`.)

granth:
This also happens during `db:setup` in a CI run. Not a big deal!

bkeepers:
@granth can you confirm this is fixed in 1.1.2?

granth:
Sorry, I should have been more clear. I see the `flipper_gates` warning in 1.1.2 during `db:setup`.

```
Mysql2::Error: Table 'circle_test.flipper_gates' doesn't exist. You likely need to run `rails g flipper:active_record` and/or `rails db:migrate`.
Already memoized items
Running via Spring preloader in process 871
** Invoke db:setup (first_time)
```

I mean, it's true that the table doesn't exist!

PR #795, opened 2023-12-17
Poll cold start fix

jnunemaker:
It's possible if using an adapter that doesn't retain state that a feature could be checked before the first background thread sync finishes. If this happens flipper thinks no features are enabled. That's not good. So let's avoid that by forcing a sync in the main thread. 

Should we rescue around the sync and warn? Probably?


jnunemaker:
Looks like I have failures to deal with first because the tests don't like doing the sync.

Issue #800, opened 2023-12-19
Sync error with flipper cloud


danielsoto15:
https://github.com/flippercloud/flipper/blob/6c485feab14ccd953e01165c63de0b03ec3524e3/lib/flipper/adapters/sequel.rb#L107C77-L107C77
<img width="1415" alt="Screenshot 2023-12-19 at 8 39 35" src="https://github.com/flippercloud/flipper/assets/117784240/4addc038-e6d6-443a-a941-006e9eddf03c">

bkeepers:
@danielsoto15 Can you give us more information for replicating the error?

What version are you using of Sequel, Flipper, and Ruby?

jmoglesby:
Answering for @danielsoto15 

![CleanShot 2023-12-19 at 12 22 42](https://github.com/flippercloud/flipper/assets/25443574/44224a78-c577-465f-b2eb-48fe1745dd90)

![CleanShot 2023-12-19 at 12 23 06](https://github.com/flippercloud/flipper/assets/25443574/30bc4ee7-2cb3-4ce8-b09c-c0c7aff4f031)

![CleanShot 2023-12-19 at 12 23 49](https://github.com/flippercloud/flipper/assets/25443574/023200f9-f74a-4321-bb7f-5da63a1f7343)

jmoglesby:
@bkeepers,
Is this just as simple as changing:
```ruby
features_sql = @feature_class.select(:key.qualify(feature_table).as(:feature_key))
```
to
```ruby
features_sql = @feature_class.select(&:key.qualify(feature_table).as(:feature_key))
```
??

danielsoto15:
Yes and also on the following two lines could need the same fix

jnunemaker:
Weird. Any idea why this broke? Did sequel change? We haven't changed this part of the adapter in years. 

How does this look for a fix: https://github.com/flippercloud/flipper/pull/802

jmoglesby:
@jnunemaker 
We're trying to make this happen, but... running into complications in our codebase that are slowing it down. I'm probably just an idiot and not doing this right. 😑 

Would you mind helping me understand how this works when you have multiple gemspecs bundled into a single repository?
If I'm referencing your fix branch for the `flipper-sequel` gem, do I also need to do that for the `flipper-cloud` gem?
[...]

PR #802, opened 2023-12-19
Use top level sequel to qualify table and column name

jnunemaker:
xref https://github.com/flippercloud/flipper/issues/800


bkeepers (commented):
Is there any way to replicate the error in tests?

jnunemaker:
@bkeepers yeah I dont know, but my guess is that the symbol extension is more of a hack like I did long ago for plucky whereas this top level stuff is official and will be supported.

jmoglesby (approved):
This fix solved the issue we were having in #800

PR #804, opened 2023-12-30
Setup generator

bkeepers:
```
$ bundle add flipper-active_record
$ rails generate flipper:setup -t <YOUR-TOKEN>
      create  db/migrate/20231230170601_create_flipper_tables.rb
      append  config/credentials.yml.enc
```

- [x] Call `flipper:active_record` if adapter is installed
- [x] `--token xyz` adds to dotenv if it exists
- [x] `--token xyz` adds to Rails credentials if it it's used

### DIFF
--- a/lib/generators/flipper/setup_generator.rb
+++ b/lib/generators/flipper/setup_generator.rb
@@ -28,7 +28,7 @@ module Flipper
       end
 
       def configure_with_credentials
-        return unless exists?("config/master.key") && exists?("config/credentials.yml.enc")
+        return unless exists?("config/credentials.yml.enc") && (ENV["RAILS_MASTER_KEY"] || exists?("config/master.key"))
 
         content = "flipper:\n  cloud_token: #{options[:token]}\n"
         action InjectIntoEncryptedFile.new(self, Rails.application.credentials, content, after: /\z/)
--- a/lib/generators/flipper/setup_generator.rb
+++ b/lib/generators/flipper/setup_generator.rb
@@ -6,7 +6,7 @@ module Flipper
       desc 'Peform any necessary steps to install Flipper'
 
       class_option :token, type: :string, default: nil, aliases: '-t',
-        desc: "Your Flipper Cloud development token"
+        desc: "Your personal environment token for Flipper Cloud"
 
       def generate_active_record
         invoke 'flipper:active_record' if defined?(Flipper::Adapters::ActiveRecord)
--- a/lib/generators/flipper/setup_generator.rb
+++ b/lib/generators/flipper/setup_generator.rb
@@ -5,9 +5,57 @@ module Flipper
     class SetupGenerator < ::Rails::Generators::Base
       desc 'Peform any necessary steps to install Flipper'
 
-      def perform
-        if defined?(Flipper::Adapters::ActiveRecord)
-          invoke 'flipper:active_record'
+      class_option :token, type: :string, default: nil, aliases: '-t',
+        desc: "Your Flipper Cloud development token"
+
+      def generate_active_record
+        invoke 'flipper:active_record' if defined?(Flipper::Adapters::ActiveRecord)
+      end
+
+      def configure_cloud_token
+        return unless options[:token]
+
+        configure_with_dotenv || configure_with_credentials
+      end
+
+      private
+
+      def configure_with_dotenv
+        ['.env.development', '.env.local', '.env'].detect do |file|
+          next unless exists?(file)
+          append_to_file file, "\nFLIPPER_CLOUD_TOKEN=#{options[:token]}\n"
+        end
+      end
+
+      def configure_with_credentials
+        return unless exists?("config/master.key") && exists?("config/credentials.yml.enc")
+
+        content = "flipper:\n  cloud_token: #{options[:token]}\n"
+        action InjectIntoEncryptedFile.new(self, Rails.application.credentials, content, after: /\z/)
+      end
+
+      # Check if a file exists in the destination root
+      def exists?(path)
+        File.exist?(File.expand_path(path, destination_root))
+      end
+
+      # Action to inject content into ActiveSupport::EncryptedFile
+      class InjectIntoEncryptedFile < Thor::Actions::InjectIntoFile
+        def initialize(base, encrypted_file, data, config)
+          @encrypted_file = encrypted_file
+          super(base, encrypted_file.content_path, data, config)
+        end
+
+        def content
+          @content ||= @encrypted_file.read
+        end
+
+        def replace!(regexp, string, force)
+          if force || !replacement_present?
+            success = content.gsub!(regexp, string)
+            @encrypted_file.write content unless pretend?
+            success
+          end
         end
       end
     end
--- a/lib/generators/flipper/setup_generator.rb
+++ b/lib/generators/flipper/setup_generator.rb
@@ -1,0 +1,15 @@
+require 'rails/generators/active_record'
+
+module Flipper
+  module Generators
+    class SetupGenerator < ::Rails::Generators::Base
+      desc 'Peform any necessary steps to install Flipper'
+
+      def perform
+        if defined?(Flipper::Adapters::ActiveRecord)
+          invoke 'flipper:active_record'
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


test_rails/generators/flipper/setup_generator_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    Flipper::Adapters.send(:remove_const, :ActiveRecord) rescue nil
+
+    run_generator
+    assert_no_migration "db/migrate/create_flipper_tables.rb"
+  end
+
+  %w(.env.development .env.local .env).each do |file|
+    test "configures Flipper Cloud token in #{file} if it exists" do
+      File.write("#{ROOT}/#{file}", "")
+      run_generator ["--token", "abc123"]
```
garrettdimon, line 31: Really minor, but using `%w(--token abc123)` makes it so that these arguments and options read more like they would on the command line and can more easily be copy/pasted between the two for manual investigation. The `generator` version needs the args and kwargs separated, but with these, it's just nice to have them mirror how it works on the command line.

lib/generators/flipper/setup_generator.rb
```diff
@@ -0,0 +1,63 @@
...
+module Flipper
+  module Generators
+    class SetupGenerator < ::Rails::Generators::Base
+      desc 'Peform any necessary steps to install Flipper'
+
+      class_option :token, type: :string, default: nil, aliases: '-t',
+        desc: "Your Flipper Cloud development token"
+
+      def generate_active_record
+        invoke 'flipper:active_record' if defined?(Flipper::Adapters::ActiveRecord)
```
garrettdimon, line 12: Any specific reason for not using `generate` here instead of `invoke`? And on the super-minor side, using the conditional as a guard clause instead?

garrettdimon (commented):
Overall, I dig it, but I do think the token should just be an argument since there's never a case where it would be run without it, right?

Also, not important for a first pass, but it might be nice to have an option to automatically run the migrations as well after generating them. It doesn't need to by default, but it reduces the amount of commands to type/copy/paste if we can have a sample command that includes the option and runs the migrations by default. (Although that does blur the line with whether this is generating or installing, so it may be a step too far.)

bkeepers:
> …but I do think the token should just be an argument since there's never a case where it would be run without it, right?

Open source users will run it without the arg, but I guess it could be optional. You and I also talked about allowing signup through the generator, like `rails g flipper:setup email@example.com`, or fetching the token if given credentials. What do you think about waiting until we get some of the other functionality built to decide if this should be an arg instead of an option?

lib/generators/flipper/setup_generator.rb
```diff
@@ -0,0 +1,63 @@
...
+
+      def configure_with_dotenv
+        ['.env.development', '.env.local', '.env'].detect do |file|
+          next unless exists?(file)
+          append_to_file file, "\nFLIPPER_CLOUD_TOKEN=#{options[:token]}\n"
+        end
+      end
+
+      def configure_with_credentials
+        return unless exists?("config/credentials.yml.enc") && (ENV["RAILS_MASTER_KEY"] || exists?("config/master.key"))
```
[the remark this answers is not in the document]
bkeepers: Yeah, good call. I was only thinking about development on this first pass, but I'd like to iterate on this and add multiple environment support.

garrettdimon:
I'm totally down with punting, I was just trying to minimize the churn for the generator arguments/options.

test_rails/generators/flipper/setup_generator_test.rb
bkeepers: Good call!

lib/generators/flipper/setup_generator.rb
bkeepers: The tests fail when I try to use `generate` since they're not running in the context of a real Rails app:

```
/Users/bkeepers/.asdf/installs/ruby/3.3.0-rc1/lib/ruby/3.3.0+0/bundled_gems.rb:68:in `require': cannot load such file -- /Users/bkeepers/projects/flipper/config/boot (LoadError)
	from /Users/bkeepers/.asdf/installs/ruby/3.3.0-rc1/lib/ruby/3.3.0+0/bundled_gems.rb:68:in `block (2 levels) in replace_require'
	from /Users/bkeepers/.asdf/installs/ruby/3.3.0-rc1/lib/ruby/gems/3.3.0+0/gems/railties-7.0.8/lib/rails/app_loader.rb:58:in `block in exec_app'
	from <internal:kernel>:187:in `loop'
	from /Users/bkeepers/.asdf/installs/ruby/3.3.0-rc1/lib/ruby/gems/3.3.0+0/gems/railties-7.0.8/lib/rails/app_loader.rb:48:in `exec_app'
	from /Users/bkeepers/.asdf/installs/ruby/3.3.0-rc1/lib/ruby/gems/3.3.0+0/gems/railties-7.0.8/lib/rails/cli.rb:7:in `<top (required)>'
	from /Users/bkeepers/.asdf/installs/ruby/3.3.0-rc1/lib/ruby/3.3.0+0/bundled_gems.rb:68:in `require'
	from /Users/bkeepers/.asdf/installs/ruby/3.3.0-rc1/lib/ruby/3.3.0+0/bundled_gems.rb:68:in `block (2 levels) in replace_require'
	from /Users/bkeepers/.asdf/installs/ruby/3.3.0-rc1/lib/ruby/gems/3.3.0+0/gems/railties-7.0.8/exe/rails:10:in `<top (required)>'
	from /Users/bkeepers/.asdf/installs/ruby/3.3.0-rc1/bin/rails:25:in `load'
	from /Users/bkeepers/.asdf/installs/ruby/3.3.0-rc1/bin/rails:25:in `<main>'
```

lib/generators/flipper/setup_generator.rb
```diff
@@ -0,0 +1,63 @@
+require 'rails/generators/active_record'
+
+module Flipper
+  module Generators
+    class SetupGenerator < ::Rails::Generators::Base
+      desc 'Peform any necessary steps to install Flipper'
+
+      class_option :token, type: :string, default: nil, aliases: '-t',
+        desc: "Your Flipper Cloud development token"
```
jnunemaker, line 9: I'd maybe swap "development" to "personal", but if you feel strongly either way I'll leave it up to you.

lib/generators/flipper/setup_generator.rb
```diff
@@ -0,0 +1,63 @@
...
+
+      def configure_cloud_token
+        return unless options[:token]
+
+        configure_with_dotenv || configure_with_credentials
+      end
+
+      private
+
+      def configure_with_dotenv
```
jnunemaker, line 23: SICK. 😎

test_rails/generators/flipper/setup_generator_test.rb
```diff
@@ -0,0 +1,64 @@
...
+  %w(.env.development .env.local .env).each do |file|
+    test "configures Flipper Cloud token in #{file} if it exists" do
+      File.write("#{ROOT}/#{file}", "")
+      run_generator %w(--token abc123)
+      assert_file file, /^FLIPPER_CLOUD_TOKEN=abc123$/m
+    end
+  end
+
+  test "configures Flipper Cloud token in .env.development before .env" do
+    File.write("#{ROOT}/.env.development", "")
```
jnunemaker, line 37: Is there some test magic that removes these files (and credentials) in setup or teardown or do they all need to be rm'd?

jnunemaker (approved):
Sweet! Love the reduction in friction. Does it also run the migrations after they are generated?

test_rails/generators/flipper/setup_generator_test.rb
garrettdimon: You know this. https://github.com/flippercloud/flipper/pull/804/files/87d8a2954a1a3464178d4741f4755bf577ed3f0c..4748609b66601a6e94d96dad4ae2e958084a9039#diff-20ac1d85392c5779ec3830ff97c1ed050bc0fb9897b576cb795245320c4853eeL6-L8

lib/generators/flipper/setup_generator.rb
garrettdimon: And probably explicitly state "...your personal _environment_ token"

bkeepers: 👍 Done

bkeepers:
> Does it also run the migrations after they are generated?

It does not. Do other generators often run migrations? We definitely could do it, just not sure I've ever seen another library do that.

In all honesty, the other reason it doesn't is because I couldn't figure out a way to get the tests to still run with it. Adding `rails_command 'db:migrate'` blows up because we don't have a dummy rails app that the test are running against.

PR #808, opened 2024-01-03
Add flipper/test_help to automatically configure Flipper in tests/specs.

bkeepers:
This adds `flipper/test_help` to automatically configure Rails/RSpec the way [we recommend in our testing docs](https://www.flippercloud.io/docs/testing).

Open questions:
1. Should `Flipper::Engine` automatically require this if `RAILS_ENV=test`? I think yes, but maybe we wait until 2.0 to do that?
2. As brought up in #807, what should the default value for `strict` be in test? Currently with `strict = :warn`, you also have to call `Flipper.add` for each feature. In theory, I like the explicitness of this. It will guard against typos in flag names, but also seems like busywork.  One option is to establish a pattern for defining features in an initializer, like this:

```ruby
module Feature
  Hubspot = Flipper.add(:hubspot)
  Seo = Flipper.add(:seo)
end

# app/jobs/hubspot_sync_job.rb
def perform(resource)
  return unless Feature::Hubspot.enabled?(resource)
end
```

### DIFF
--- a/lib/flipper/test_help.rb
+++ b/lib/flipper/test_help.rb
@@ -2,14 +2,17 @@ module Flipper
   module TestHelp
     extend self
 
-    def before_all
+    def flipper_configure
+      # Create a single shared memory adapter instance for each test
+      @flipper_adapter = Flipper::Adapters::Memory.new
+
       Flipper.configure do |config|
-        config.adapter { Flipper::Adapters::Memory.new }
+        config.adapter { @flipper_adapter }
         config.default { Flipper.new(config.adapter) }
       end
     end
 
-    def before_each
+    def flipper_reset
       Flipper.instance = nil # Reset previous flipper instance
     end
   end
@@ -17,13 +20,19 @@ end
 
 if defined?(RSpec)
   RSpec.configure do |config|
-    config.before(:all) { Flipper::TestHelp.before_all }
-    config.before(:each) { Flipper::TestHelp.before_each }
+    config.include Flipper::TestHelp
+    config.before(:all) { flipper_configure }
+    config.before(:each) { flipper_reset }
   end
-elsif defined?(ActiveSupport::TestCase)
-  Flipper::TestHelp.before_all
+end
 
-  ActiveSupport::TestCase.setup do
-    Flipper::TestHelp.before_each
+if defined?(ActiveSupport)
+  ActiveSupport.on_load(:active_support_test_case) do
+    ActiveSupport::TestCase.class_eval do
+      include Flipper::TestHelp
+
+      setup :flipper_configure
+      setup :flipper_reset
+    end
   end
 end
--- a/lib/flipper/test_help.rb
+++ b/lib/flipper/test_help.rb
@@ -1,0 +1,29 @@
+module Flipper
+  module TestHelp
+    extend self
+
+    def before_all
+      Flipper.configure do |config|
+        config.adapter { Flipper::Adapters::Memory.new }
+        config.default { Flipper.new(config.adapter) }
+      end
+    end
+
+    def before_each
+      Flipper.instance = nil # Reset previous flipper instance
+    end
+  end
+end
+
+if defined?(RSpec)
+  RSpec.configure do |config|
+    config.before(:all) { Flipper::TestHelp.before_all }
+    config.before(:each) { Flipper::TestHelp.before_each }
+  end
+elsif defined?(ActiveSupport::TestCase)
+  Flipper::TestHelp.before_all
+
+  ActiveSupport::TestCase.setup do
+    Flipper::TestHelp.before_each
+  end
+end
--- a/lib/flipper/test_help.rb
+++ b/lib/flipper/test_help.rb
@@ -1,7 +1,5 @@
 module Flipper
   module TestHelp
-    extend self
-
     def flipper_configure
       # Create a single shared memory adapter instance for each test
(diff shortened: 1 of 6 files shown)


lib/flipper/test_help.rb
```diff
@@ -0,0 +1,29 @@
+module Flipper
+  module TestHelp
+    extend self
+
+    def before_all
+      Flipper.configure do |config|
+        config.adapter { Flipper::Adapters::Memory.new }
+        config.default { Flipper.new(config.adapter) }
```
jnunemaker, line 8: Does this apply all the `config.use` adapters? It would seem like no. Should it? I know everyone would then have the strict issue but I wonder if it would be confusing if someone uses `use` and doesn't see it in their adapter stack.

lib/flipper/test_help.rb
```diff
@@ -0,0 +1,29 @@
+module Flipper
+  module TestHelp
+    extend self
+
+    def before_all
+      Flipper.configure do |config|
+        config.adapter { Flipper::Adapters::Memory.new }
```
jnunemaker, line 7: Should we do anything different for system tests? I was reminded of [this conversation](https://github.com/flippercloud/flipper/issues/615#issuecomment-1625740455) from a year ago.

jnunemaker (approved):
Looks good as always. Added some questions but I'm not sure on their answers. 

I thought about mentioning Minitest but then I couldn't think of a scenario where it would be used. 

> Should Flipper::Engine automatically require this if RAILS_ENV=test? I think yes, but maybe we wait until 2.0 to do that?

Automatic seems good. We could add an env var or config to disable automatic requiring. 

Why wait until 2.0? Does it seem like a breaking change? I wonder if it would be ok. If anyone has these things configured in rspec or active support test case, those configurations would overwrite these, wouldn't they? 

> As brought up in https://github.com/flippercloud/flipper/discussions/807, what should the default value for strict be in test? Currently with strict = :warn, you also have to call Flipper.add for each feature. In theory, I like the explicitness of this. It will guard against typos in flag names, but also seems like busywork. 

I can see `:raise` as default. In theory, I like the explicitness of that too. 

> One option is to establish a pattern for defining features in an initializer, like this:

I also like establishing a pattern for defining features. I'd probably do it through config, as you mentioned in chat (but I'll copy it in here for everyone else). 

```ruby
Flipper.configure do |config|
  config.add :a, :b, :c
end
```

The only downside of this is that it requires network calls on boot, which I'm usually anti. Or maybe we can avoid that by only doing the network calls if strict is turned on in some fashion? That would limit it to only in dev/test. Thoughts?

lib/flipper/test_help.rb
bkeepers: Yes. `config.adapter` will return the storage adapter wrapped in `use` adapters.

lib/flipper/test_help.rb
bkeepers: Yes. I spent some time investigating this today and posted what I found here: https://github.com/flippercloud/flipper/issues/615#issuecomment-1879449641

Now I just need to try it with RSpec, and then figure out how to configure it so it just works.

Might be worth revisiting in a followup PR, or I'm also happy to leave this set until we figure it out.

jnunemaker: I’m fine with follow up as long as people can override somehow!

davidwessman:
Went down a nice little rabbit hole while investigating our Flipper-test-setup, guess I am waiting for the next release!
Well done 🚀

lib/flipper/test_help.rb
```diff
@@ -0,0 +1,36 @@
...
+      @flipper_adapter = Flipper::Adapters::Memory.new
+
+      Flipper.configure do |config|
+        config.adapter { @flipper_adapter }
+        config.default { Flipper.new(config.adapter) }
+      end
+    end
+
+    def flipper_reset
+      Flipper.instance = nil # Reset previous flipper instance
```
fcheung, line 14: I got some spec failures after this - you get a new flipper instance but using the same memory adapter, so there's  leakage between tests. 

I can make everything green again if I add `Flipper.instance.import(Flipper::Adapters::Memory.new)` to the reset (which was the simplest way I could find to delete all the data from the shared adapter)

lib/flipper/test_help.rb
rnestler: This broke our system specs, so to answer the question from above:

> Why wait until 2.0? Does it seem like a breaking change?

This was a breaking change :slightly_smiling_face:

lib/flipper/test_help.rb
rnestler: I think I hit the same. Maybe it would be better to just disable all feature flags in `reset`?

fcheung: Yes, I hit this. The app thread and spec thread end up using a different flipper instance (fine). The app thread is running for the entire duration of the specs so keeps the same flipper instance.

On the second spec the before each creates a new flipper instance but this only affects the spec thread, the app thread still has the old memory adapter.

To make this work I had to disable the builtin flipper/test_help (with the env var) and then do what I was doing before: create the shared memory adapter once in before(suite) and clear out its contents in before each

lib/flipper/test_help.rb
bkeepers: Not intentionally. Thanks for being helpful.

PR #810, opened 2024-01-05
Use Sanitize on actor_names

bryszard:
Solves issue - https://github.com/flippercloud/flipper/issues/805

### Motivation & context

As pointed out in the issue, there is currently an inconsistency between how we treat on the UI some elements (eg. `Flipper::UI.configuration.banner_text`, `@feature.description`) and how we treat the `@feature.actor_names`.

The real world use case is making the actor names clickable as links to the resources that are the actors.

I've done the simplest possible change, not changing anything about the content itself.

### Changes

- apply `Sanitize.fragment` to the `"#{@feature.actor_names[item]} (#{item})"` on `feature.erb`

### Visual QA

image

### DIFF
--- a/examples/ui/basic.ru
+++ b/examples/ui/basic.ru
@@ -42,8 +42,8 @@ Flipper::UI.configure do |config|
 
   config.actor_names_source = lambda do |_keys|
     {
-      '1' => 'John',
-      '6' => 'Brandon',
+      '1' => '<a href="https://johnnunemaker.com">John</a>',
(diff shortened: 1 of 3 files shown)


bkeepers (approved):
Thanks for the PR. It would be great to have a test for this too.  Let me know if you want help adding that.

bkeepers:
@bryszard 1.2.0 is out with this change. Thanks for your help! https://blog.flippercloud.io/flipper-1-2-0/

bryszard:
Thank you for writing tests!

PR #827, opened 2024-01-18
TestHelp: Fix setup for RSpec, clear adapter before each test

### LINKED ISSUE #821: RSpec/Rails Issues with 1.2.1
Still having issues under 1.2.1 with RSpec/Rails. Not sure I have time tonight to get a PR, so I'm posting some quick notes. 

What I think is happening is: `rspec-rails` uses `ActiveSupport::TestCase`, so both of these blocks are running and clobbering each other:

https://github.com/flippercloud/flipper/blob/2c9386d3b84680d18bb73bc4267f2d82ee3546f0/lib/flipper/test_help.rb#L19-L37

In fact, if I comment out the RSpec chunk, my tests go back to passing.

### LINKED ISSUE #822: Feature flag not being enabled in RSpec specs on 1.2.1
This may be a duplicate of #821, but I don't know enough about the internals of the library to say if it is.

In a spec, my code tries to enable a feature flag, but the flag does not end up enabled:

```ruby
# A helper we use in our specs
module TestUtils
  extend ActiveSupport::Concern

  def with_feature_flag(flag, value)
    old_value = Flipper.enabled?(flag)
    if value == false
      puts "disabling #{flag}"
      Flipper.disable(flag)
    else
      puts "enabling #{flag}"
      Flipper.enable(flag)
    end
    yield
  ensure
    if old_value == false
      puts "setting #{flag} back to disabled"
      Flipper.disable(flag)
    else
      puts "setting #{flag} back to enabled"
      Flipper.enable(flag)
    end
  end

  module ClassMethods
    def with_feature_flag(flag, value)
      around { |example| with_feature_flag(flag, value) { example.run } }
    end
  end
end
```

```ruby
# Excerpt of our spec
context "when the automatic_button feature flag is enabled" do
  with_feature_flag :automatic_button, true

  context "when there is a failure message" do
    before { subject.object.fulfillment.failure_code = "cancelled" }

    it "returns the red banner" do
       puts "Flag symbol enabled? #{Flipper.enabled? :automatic_button}"
      puts "Flag string enabled? #{Flipper.enabled? 'automatic_button'}"
      expect(subject.banner.to_partial_path).to eq "admin/order_banners/automatic_processing_failure"
    end

    it "returns the red banner, even if it can be processed automatically" do
      subject.object.fulfillment.can_process_automatically = true
      expect(subject.banner.to_partial_path).to eq "admin/order_banners/automatic_processing_failure"
    end
  end
end
```

On 1.2.1:
```
$ rspec spec/decorators/order_decorator_spec.rb:114
Run options: include {:locations=>{"./spec/decorators/order_decorator_spec.rb"=>[114]}}
enabling automatic_button
Flag symbol enabled? false
Flag string enabled? false
setting automatic_button back to disabled
```
On 1.1.2:
```
$ rspec spec/decorators/order_decorator_spec.rb:114
Run options: include {:locations=>{"./spec/decorators/order_decorator_spec.rb"=>[114]}}
enabling automatic_button
Flag symbol enabled? true
Flag string enabled? true
setting automatic_button back to disabled
```
In case it's helpful, this is on Rails 7.0.8 and RSpec 3.12.0.

bkeepers:
This should fix #821 and fix #822.

I can't get a failing RSpec spec in this repo right now, but will work on refactoring our test suite to get one similar to `test_rails/system/test_help_test.rb` for rspec.

@danielmorrison @norman @rnestler @davidwessman @fcheung can you test this branch and confirm that it works for you?

cc #808

### DIFF
--- a/lib/flipper/test_help.rb
+++ b/lib/flipper/test_help.rb
@@ -1,17 +1,24 @@
 module Flipper
   module TestHelp
+    extend self
+
     def flipper_configure
-      # Create a single shared memory adapter instance for each test
-      @flipper_adapter = Flipper::Adapters::Memory.new
+      # Use a shared Memory adapter for all tests. This is instantiated outside of the
+      # `configure` block so the same instance is returned in new threads.
+      adapter = Flipper::Adapters::Memory.new
 
       Flipper.configure do |config|
-        config.adapter { @flipper_adapter }
+        config.adapter { adapter }
         config.default { Flipper.new(config.adapter) }
       end
     end
 
     def flipper_reset
-      Flipper.instance = nil # Reset previous flipper instance
+      # Remove all features
+      Flipper.features.each(&:remove) rescue nil
+
+      # Reset previous DSL instance
+      Flipper.instance = nil
     end
   end
 end
@@ -19,19 +26,17 @@ end
 if defined?(RSpec) && RSpec.respond_to?(:configure)
   RSpec.configure do |config|
     config.include Flipper::TestHelp
-    config.before(:each) do
-      flipper_reset
-      flipper_configure
-    end
+    config.before(:suite) { Flipper::TestHelp.flipper_configure }
+    config.before(:each) { flipper_reset }
   end
 end
-
 if defined?(ActiveSupport)
   ActiveSupport.on_load(:active_support_test_case) do
+    Flipper::TestHelp.flipper_configure
+
     ActiveSupport::TestCase.class_eval do
       include Flipper::TestHelp
 
-      setup :flipper_configure
       setup :flipper_reset
     end
   end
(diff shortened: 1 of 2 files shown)


davidwessman:
Not using RSpec, just minitest

bkeepers:
> Not using RSpec, just minitest

Sorry about that. I just CC'd everyone from #808

rnestler (approved):
Works for me :slightly_smiling_face: 

Edit: And slightly improves performance in our specs (2.5% less time) compared to setting `FLIPPER_TEST_HELP` to `'false'`.

norman:
Sorry, still fails for me with the exact same output as in #822.

fcheung:
Works for me (using rspec) - thanks for a quick fix

PR #828, opened 2024-01-19
Support Rack 3

danielmorrison:
Add support for Rack 3. 

Flipper UI's dependency on`rack-protection` was locking the `rack` version to 2x. Now that `rack-protection` 4 is out, we can update `rack`. This brings up a good question: should Flipper UI require `rack-protection`? I like it and all, but might not be Flipper's responsibility. 🤷 

`rack-session` has been extracted from `rack`, but has versions to work under rack 2 or 3.

I have a couple import-related tests failing locally, but haven't figured out exactly why yet. The `request.body` isn't coming through (the body is a `Rackup::Handler::WEBrick::Input` and `.rewind` rails, but trying to `.read` returns `""`).

### DIFF
--- a/flipper-ui.gemspec
+++ b/flipper-ui.gemspec
@@ -21,7 +21,8 @@ Gem::Specification.new do |gem|
   gem.metadata      = Flipper::METADATA
 
   gem.add_dependency 'rack', '>= 1.4', '< 4'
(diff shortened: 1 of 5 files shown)


jnunemaker:
Requiring it has caused some issues in the past. I only use it for csrf. Not sure how that would work without requiring it.

PR #832, opened 2024-01-31
Fix: Respect Rails defaults in an initializer by delaying loading ActiveRecord adapter until framework is loaded

pbstriker38:
The Flipper models are loading too soon and causing weirdness with saving models that have a belongs_to. The belongs_to association is being loaded on save even when there have not been any changes to the association.

BEFORE
```ruby
irb(main):001> token = OrganizationAccessToken.find(1)
...
irb(main):002> token.update(description: 'a token')
DEBUG:   TRANSACTION (0.7ms)  BEGIN
DEBUG:   Organization Load (1.0ms)  SELECT "organizations".* FROM "organizations" WHERE "organizations"."status" != $1 AND "organizations"."id" = $2 LIMIT $3  [["status", "deactivated"], ["id", 3], ["LIMIT", 1]]
DEBUG:   OrganizationAccessToken Update (1.0ms)  UPDATE "organization_access_tokens" SET "description" = $1, "updated_at" = $2 WHERE "organization_access_tokens"."id" = $3  [["description", "a token"], ["updated_at", "2024-01-31 22:19:38.068599"], ["id", 1]]
DEBUG:   TRANSACTION (1.4ms)  COMMIT
=> true
```

AFTER
```ruby
irb(main):001> token = OrganizationAccessToken.find(1)
...
irb(main):002> token.update(description: 'a token')
DEBUG:   TRANSACTION (0.7ms)  BEGIN
DEBUG:   OrganizationAccessToken Update (1.8ms)  UPDATE "organization_access_tokens" SET "description" = $1, "updated_at" = $2 WHERE "organization_access_tokens"."id" = $3  [["description", "a token"], ["updated_at", "2024-01-31 22:21:21.908915"], ["id", 1]]
DEBUG:   TRANSACTION (0.9ms)  COMMIT
=> true
```

### DIFF
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -7,34 +7,36 @@ module Flipper
     class ActiveRecord
       include ::Flipper::Adapter
 
-      # Abstract base class for internal models
-      class Model < ::ActiveRecord::Base
-        self.abstract_class = true
-      end
+      ActiveSupport.on_load(:active_record) do
+        # Abstract base class for internal models
+        class Model < ::ActiveRecord::Base
+          self.abstract_class = true
+        end
 
-      # Private: Do not use outside of this adapter.
-      class Feature < Model
-        self.table_name = [
-          Model.table_name_prefix,
-          "flipper_features",
-          Model.table_name_suffix,
-        ].join
+        # Private: Do not use outside of this adapter.
+        class Feature < Model
+          self.table_name = [
+            Model.table_name_prefix,
+            "flipper_features",
+            Model.table_name_suffix,
+          ].join
 
-        has_many :gates, foreign_key: "feature_key", primary_key: "key"
+          has_many :gates, foreign_key: "feature_key", primary_key: "key"
 
-        validates :key, presence: true
-      end
+          validates :key, presence: true
+        end
 
-      # Private: Do not use outside of this adapter.
-      class Gate < Model
-        self.table_name = [
-          Model.table_name_prefix,
-          "flipper_gates",
-          Model.table_name_suffix,
-        ].join
+        # Private: Do not use outside of this adapter.
+        class Gate < Model
+          self.table_name = [
+            Model.table_name_prefix,
+            "flipper_gates",
+            Model.table_name_suffix,
+          ].join
 
-        validates :feature_key, presence: true
-        validates :key, presence: true
+          validates :feature_key, presence: true
+          validates :key, presence: true
+        end
       end
 
       VALUE_TO_TEXT_WARNING = <<-EOS
--- a/spec/flipper/adapters/active_record_spec.rb
+++ b/spec/flipper/adapters/active_record_spec.rb
@@ -23,6 +23,7 @@ RSpec.describe Flipper::Adapters::ActiveRecord do
     {
       "adapter" => "mysql2",
       "encoding" => "utf8mb4",
+      "host" => ENV["MYSQL_HOST"],
       "username" => ENV["MYSQL_USER"] || "root",
       "password" => ENV["MYSQL_PASSWORD"] || "",
       "database" => ENV["MYSQL_DATABASE"] || "flipper_test",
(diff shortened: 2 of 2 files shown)


pbstriker38:
I moved the 3 commits that fix the tests to this other PR https://github.com/flippercloud/flipper/pull/833

bkeepers:
@pbstriker38 thanks for the PR. I'd love to figure out a test for this. We used to wait to wait to load the AR adaptar with `ActiveSupport.on_load(:active_record)`, but it caused some other weirdness that I can't remember now. Any ideas for a failing test that demonstrates the issue you're running into?

pbstriker38:
I could see if I can get it to reproduce with one of the Rails executable test case examples

https://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#create-an-executable-test-case

pbstriker38:
We've been running this fork for 2 months now because without it many of our tests that use `FactoryBot.build` fail since it attempts to load the association from the DB.

pbstriker38:
@bkeepers I found where it was initially added and removed.

https://github.com/flippercloud/flipper/pull/192

https://github.com/flippercloud/flipper/pull/652

My implementation is a bit different though as it does not wrap the entire adapter but only the definition of the models.

spec/flipper/adapters/active_record_spec.rb
```diff
@@ -215,18 +215,27 @@
         end
       end
 
-      it "works when table doesn't exist" do
```
pbstriker38, line 218: I noticed that this test wasn't actually testing the table not existing, but the fact that it's not even connected to the DB. Without the `silence`, you can see the message:

```
No connection pool for 'ActiveRecord::Base' found.. You likely need to run `rails g flipper:active_record` and/or `rails db:migrate`.
```

pbstriker38:
Ok. So I created a new rails app to create a minimal reproduction. I could not get it to reproduce. The one difference from this new app and ours is the configuration defaults. A new app has `config.load_defaults 7.1`, while our app has `config.load_defaults 7.0` because we have not completed enabling all of the configs in `new_framework_defaults_7_1.rb`. 

image

Flipper is causing ActiveRecord to load too soon and it ignores our `new_framework_defaults_7_1.rb` configuration for `belongs_to_required_validates_foreign_key`.
```ruby
Rails.application.config.active_record.belongs_to_required_validates_foreign_key = false
```

Without my fix
```ruby
irb(main):001> ActiveRecord.belongs_to_required_validates_foreign_key
=> true
```

with my fix
```ruby
irb(main):001> ActiveRecord.belongs_to_required_validates_foreign_key
=> false
```


https://github.com/rails/rails/commit/e5d15140d2dfd25f33e306ab8522f7accd410dc3

pbstriker38:
Using `require: false` in the Gemfile also appears to fix the issue
```ruby
gem 'flipper-active_record', require: false
```

lib/flipper/adapters/active_record.rb
```diff
@@ -7,34 +7,36 @@ module Adapters
     class ActiveRecord
       include ::Flipper::Adapter
 
-      # Abstract base class for internal models
-      class Model < ::ActiveRecord::Base
-        self.abstract_class = true
-      end
+      ActiveSupport.on_load(:active_record) do
```
byroot, line 10: Would be simpler to just move all the things that depends on Active Record into another file and then require it when AR is loaded, or something like that.

pbstriker38: There's already a generator for adding the DB migrations, so another option is to just have the generator add these models to the `app/models` directory.

jnunemaker:
@bkeepers you have more experience with these bits. Any thoughts? Good to merge?

bkeepers:
> @bkeepers you have more experience with these bits. Any thoughts? Good to merge?

Yes, I'm fine moving forward with this as-is, but do want to highlight a concern:

I'm a little nervious that there could be unintended consequences with this specific solution. For example, what happens if this adapter gets loaded in the app boot process before AR is loaded? Previously, it would work fine. Now, it will raise a `NameError` with `uninitialized constant Flipper::Adapters::ActiveRecord::Model` and the app will fail to boot.

It's probably a non-issue since AR gets loaded so early in the app boot process and before any user code runs. I'm just trying to think through scenarios.

If it causes any issues we could explore a couple other options:

1. Moving the models out to separate files and using `autoload` to load them when they are referenced, along with explicitly loading them when ActiveRecord is loaded.
2. A generator as suggested in https://github.com/flippercloud/flipper/pull/832#discussion_r1542407000 (although I would prefer this as a last resort)

we5:
Just a quick heads-up, that this change here seems to break the instructions in the docs to set `connects_to` on the model in an initializer:
https://www.flippercloud.io/docs/adapters/active-record

ryush00:
Note: if your app uses sentry-ruby, upgrading to flipper-active_record ≥ 1.3.1 can raise NameError: uninitialized constant ActiveSupport::ParameterFilter.
Adding `require "active_support/parameter_filter"` in the Sentry initializer resolves it

getsentry/sentry-docs#12082

jnunemaker:
@bonflintstone if you wrap the connects_to with this does it work:

```
ActiveSupport.on_load(:active_record) do
  # connects to stuff here
end
```

PR #833, opened 2024-02-01
Fix tests

pbstriker38:
Rack 3 does not guarantee that the body is rewindable unless you add `Rack::RewindableInput::Middleware`

https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md#rackinput-is-no-longer-required-to-be-rewindable


bkeepers:
Thanks @pbstriker38!

PR #836, opened 2024-02-05
Enable telemetry by default

jnunemaker:
This turns on telemetry by default. 

Also, this turns off telemetry logging by default so its not annoying to people.


bkeepers (approved):
📈

jnunemaker:
Just don't release this until we chat. I need to tweak a couple things on the server side.

PR #839, opened 2024-02-13
added support for redis connection pools

j-castellanos:
To solve this issue https://github.com/flippercloud/flipper/issues/609 in a DRY way I simply wrap all the public methods in the existing redis adapter


j-castellanos:
@jnunemaker could you rerun the failed test?

bkeepers:
@j-castellanos Thanks for the PR!

In https://github.com/flippercloud/flipper/pull/826#pullrequestreview-1832444441 I offered a couple suggestions of how I think we should approach connection pool support. I don't really love having different adapters Redis depending on if you're using the client directly or a connection pool. I'd rather see `Redis` and `RedisCache` updated to work directly with a client or a pool internally, or add a generic ConnectionPool adapter as attempted in #835.

I don't feel strongly between those two approaches, but I would like to make sure we add connection pool support to both `Redis` and `RedisCache` at the same time.

Thoughts?

j-castellanos:
@bkeepers how is this?

lib/flipper/adapters/poolable.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Flipper
+  module Adapters
+    module Poolable
+      def initialize(client_or_pool = nil, key_prefix: nil)
+        @pool = nil
+        @client = nil
+        if client_or_pool.respond_to?(:with)
```
j-castellanos, line 9: Maybe `is_a?(ConnectionPool)` might be better

j-castellanos:
@jnunemaker @bkeepers thoughts on this one?

jnunemaker:
I don't want to add more methods to Flipper::Adapter. I think I'd just make redis and redis cache work with a client or a pool. That seems like easiest way. 

```ruby
def client(&block)
  # if client, then yield that, if connection pool then call with and yield client
end 

# then change all methods that use @client to use client { |c| c.blah... }
```

We could just abstract a method for getting a client and if adapter was initialized with pool it will use with and a block. If initialize with a client the client would just be used. 

I'll make a note to whip something together or anyone else is free to try sooner.

j-castellanos:
@jnunemaker this wouldn't be adding more methods it just overrides existing ones to set the client if a pool was used to initialize. 

However I think solution works as well, are you planning on implementing or would you like me to? The only issue I see is when client gets called multiple times, the pool would continue requesting new clients each time so more refactoring might be needed to ensure that a single method doesn't need multiple connections

jnunemaker:
@j-castellanos ah that is a good point. I'd love a PR if you have time. It's still on my list. I just don't use redis anymore.

PR #846, opened 2024-02-18
Add base class for wrapping another adapter

bkeepers:
This adds `Flipper::Adapters::Wrapper`, which can be used as a base class for an adapter that just wraps another adapter. By default, all methods delegate to the wrapped adapter. Implement `#wrap` to customize the behavior of all delegated methods, or override individual methods as needed.

```ruby
class MyNewAdapter < Flipper::Adapters::Wrapper
  # Override a speciific method to change its behavior
  def enable(feature, gate, resource)
    # …
  end

  # Override this method to customize the behavior of all delegated methods, and just yield to
  # the block to call the wrapped adapter.
  def wrap(method, *args, **kwargs)
     if some_condition?(method)
       yield # call original adapter method
    else
       # do something else
    end
  end
end
```

And then you can configure Flipper to use the new adapter:

```
Flipper.configure do |config|
  config.use MyNewAdapter
end
```

### DIFF
--- a/lib/flipper.rb
+++ b/lib/flipper.rb
@@ -174,6 +174,7 @@ end
 
 require 'flipper/actor'
 require 'flipper/adapter'
+require 'flipper/adapters/wrapper'
 require 'flipper/adapters/memoizable'
 require 'flipper/adapters/memory'
 require 'flipper/adapters/strict'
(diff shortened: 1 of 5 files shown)


jnunemaker (approved):

jnunemaker:
💯 on the branch name.

Merging because I need this in my cache branch.

Issue #837, opened 2024-02-07
Loading all actor identifiers into memory isn't scalable


rstawarz:
Is there a way to configure/modify Flipper so that all gates are not loaded into memory? 

We have several services that were successfully using Flipper.  We began to migrate our monolith away from our custom feature flag implementation / rollout combo.  However we haven't found a way to configure the adapters to NOT load every gate into memory.  It seems all paths lead through loading all [gate_values](https://github.com/flippercloud/flipper/blob/main/lib/flipper/feature.rb#L278-L281).  In our usage, we have multiple flags that are enabled per actor for 100k+ users.  I don't see how this tool can scale in this regard.  I'm hoping I'm just missing something and someone could help shed some light on what I'm missing.

Thanks!

codenamev:
Hey @jnunemaker 👋  Ryan's been trying to make this work for us, and I think the issue may stem from the default handling of the [DSL's feature fetcher](https://github.com/flippercloud/flipper/blob/main/lib/flipper/dsl.rb#L221)?  or... are we holding it wrong? 😂 

I'd be happy to submit a PR if you point me in the right direction 😄

jnunemaker:
@rstawarz @codenamev 👋. You are correct in how flipper loads actors. 

The goal of flipper is to store a smallish data structure of rules that then get applied in your app. We encourage fewer than 100 actors for any given feature and to use groups (and soon expressions) for when the number must be higher than that. 

We have several customers/users in the 1-2k enabled actors range, but, in general, I'd think of that as an anti-pattern. The primary reason is performance. I'll try to explain in detail below but I am also happy to hop on a zoom or tuple and talk more about it. 

## Details 

If you keep a large number of enabled actors per feature, then, like you said, you don't want to pull them all across the wire to see if one is enabled. The reasonable change would be to make it so flipper can check if a single actor is enabled at the adapter level. 

But if we do that, preloading gets a lot more difficult, as instead of preloading the rules you would kind of need to preload the rules that are small or could possibly apply to N actors (user, org, project, etc.). Doing this means you trade large set of individual actors for more network calls or more complicated preloading.

The goal of flipper is to reduce network calls and be as was as possible by default. The way we achieve that is by keeping the rules relatively small. Groups and expressions (in alpha) make it possible to keep the rules small and avoid network calls. They are kind of the best of both worlds. 

[...]

PR #847, opened 2024-02-18
Add ActorLimit adapter

bkeepers:
This adds the `ActorLimit` adapter and configures it to be used by default.

```ruby
101.times { |i| Flipper.enable :some_feature Flipper::Actor.new(i) }
# Error: Actor limit of 100 exceeded for feature some_feature
```

100 has been our [recommended max](https://www.flippercloud.io/docs/features#enablement-actor) for individual actors for a while, so I started with that default.  It can easily be configured with:

```ruby
Rails.application.configure do |config|
  config.flipper.actor_limit = 1000
  
  # or disable
  config.flipper.actor_limit = false
end
```


lib/flipper/engine.rb
```diff
@@ -25,6 +25,7 @@ def self.default_strict_value
         log: ENV.fetch('FLIPPER_LOG', 'true').casecmp('true').zero?,
         cloud_path: "_flipper",
         strict: default_strict_value,
+        actor_limit: ENV["FLIPPER_ACTOR_LIMIT"]&.to_i || 100,
```
jnunemaker, line 28: 100% thought this would return 0 but tested in ira and it was 100. TIL.

lib/flipper/engine.rb
```diff
@@ -65,13 +66,12 @@ def self.default_strict_value
...
+    initializer "flipper.adapters", after: :load_config_initializers do |app|
       flipper = app.config.flipper
 
-      if flipper.strict
-        Flipper.configure do |config|
-          config.use Flipper::Adapters::Strict, flipper.strict
-        end
+      Flipper.configure do |config|
+        config.use Flipper::Adapters::Strict, flipper.strict if flipper.strict
+        config.use Flipper::Adapters::ActorLimit, flipper.actor_limit if flipper.actor_limit
```
jnunemaker, line 74: We'll have to remember to include how to turn it off in the release notes.

lib/flipper/adapters/actor_limit.rb
```diff
@@ -0,0 +1,28 @@
...
+      attr_reader :limit
+
+      def initialize(adapter, limit = 100)
+        super(adapter)
+        @limit = limit
+      end
+
+      def enable(feature, gate, resource)
+        if gate.is_a?(Flipper::Gates::Actor) && over_limit?(feature)
+          raise LimitExceeded, "Actor limit of #{@limit} exceeded for feature #{feature.key}"
```
jnunemaker, line 15: Should we encourage them to use groups or expressions and link to group enablement in the docs? 

Long term I think it would be cool to just have an entire page dedicated to this explaining why and how but a link should be enough for now.

jnunemaker (approved):
Awesome! 

ref #837

lib/flipper/engine.rb
bkeepers: Pretty sure this was a Copilot suggestion

lib/flipper/adapters/actor_limit.rb
bkeepers: Added link to https://www.flippercloud.io/docs/features/actors#limitations (deploying now and should be available in a few minutes)

PR #853, opened 2024-03-01
UI: Replace breadcrumbs with configurable nav

bkeepers:
This replaces breadcrumbs in the Flipper UI with a configurable set of `nav_items`, displayed as tabs with underlines.

You can add items to the nav:

```ruby
  config.nav_items.insert 1, { title: "Custom", href: "/my/custom/path" }
end
```

## Before

image

## After

image

### DIFF
--- a/lib/flipper/ui/action.rb
+++ b/lib/flipper/ui/action.rb
@@ -93,12 +93,6 @@ module Flipper
         @request = request
         @code = 200
         @headers = {Rack::CONTENT_TYPE => 'text/plain'}
-        @breadcrumbs =
-          if Flipper::UI.configuration.application_breadcrumb_href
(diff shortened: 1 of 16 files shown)


lib/flipper/ui/action.rb
```diff
@@ -202,16 +196,6 @@ def active?
...
-      # Public: Add a breadcrumb to the trail.
-      #
-      # text - The String text for the breadcrumb.
-      # href - The String href for the anchor tag (optional). If nil, breadcrumb
-      #        is assumed to be the end of the trail.
-      def breadcrumb(text, href = nil)
-        breadcrumb_href = href.nil? ? href : "#{script_name}#{href}"
-        @breadcrumbs << Breadcrumb.new(text, breadcrumb_href)
-      end
-
```
bkeepers, line 214: This is technically a breaking change, but I'm not sure that this method would be used outside of this code.

PR #854, opened 2024-03-01
Refactor shared adapter test

bkeepers:
This refactors some of the adapter tests related to actors to use the feature interface instead of directly inspecting the adapters data. This change is needed for some work I'm doing to support large actor sets.


jnunemaker (commented):
Do we need to do the specs too?

PR #860, opened 2024-04-08
Introduce Flipper::Adapters::FallbackToCached

jonmagic:
## Why?

For my use case I need to be able to cache values each time `features`, `get`, `get_multi`, or `get_all` but only use those values if subsequent calls to them fail. It should raise an error if no cached values are there to use.

## How?

I introduced a new adapter called `Flipper::Adapters::FallbackToCached` that inherits from the Memoizable adapter and wraps another adapter. It updates the implementation of `features`, `get`, `get_multi`, and `get_all` to call the primary adapter and caches the values. If the primary adapter raises an error, it will return the cached value instead. If there is no cached value, it will raise the error.

Please note you have to disable memoize when configuring Flipper so that it doesn't use the Memoizable adapter like it normally would.


jnunemaker:
I think my only issue with this is the inheriting from memorizable and turning memoize off. 

It might be just because I haven't started from scratch on it to see how much is shared. 

I think I'll take a stab at it from scratch to see the difference.

jonmagic:
> turning memoize off.

I think you have to turn this off or it will add the memoizable adapter as the first in the stack and it defeats the purpose of this change.

jnunemaker:
Memoizable usually just sits in front and falls back to the original adapter. So if your original falls back to in memory cache it should just work. Maybe something else weird was at play. I'll have to kick the tires.

PR #862, opened 2024-04-18
Remove redundant deletion of feature gate in Flipper::Adapters::ActiveRecord#set.

khiet:
Hello, I noticed the duplicate deletions in `Flipper::Adapters::ActiveRecord#set`
The `Flipper::Adapters::ActiveRecord#delete` already does the same `destroy_all`.

### DIFF
--- a/lib/flipper/adapter_builder.rb
+++ b/lib/flipper/adapter_builder.rb
@@ -3,7 +3,7 @@ module Flipper
   #
   #   adapter = Flipper::AdapterBuilder.new do
   #     use Flipper::Adapters::Strict
-  #     use Flipper::Adapters::Memoizer
+  #     use Flipper::Adapters::Memoizable
   #     store Flipper::Adapters::Memory
   #   end.to_adapter
   #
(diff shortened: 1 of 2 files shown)


lib/flipper/adapter_builder.rb
```diff
@@ -3,7 +3,7 @@ module Flipper
   #
   #   adapter = Flipper::AdapterBuilder.new do
   #     use Flipper::Adapters::Strict
-  #     use Flipper::Adapters::Memoizer
+  #     use Flipper::Adapters::Memoizable
```
khiet, line 6: This is unrelated to the PR but I thought I added this small typo fix. I'm happy to create another PR for this if that's more appropriate.

lib/flipper/adapters/active_record.rb
```diff
@@ -221,7 +221,6 @@ def set(feature, gate, thing, options = {})
           @gate_class.transaction do
             clear(feature) if clear_feature
             delete(feature, gate)
-            @gate_class.where(feature_key: feature.key, key: gate.key).destroy_all
```
khiet, line 224: The above `delete` does the same `destroy_all`.

bkeepers (approved):
Thanks!

PR #869, opened 2024-06-12
Address soft deprecation of `ActiveRecord::Base.connection`

joshuay03:
Related: https://github.com/flippercloud/flipper/pull/705

Refs:
- https://github.com/rails/rails/blob/b752c38e81a310c1aaca78c7cdd1784009ea189a/activerecord/lib/active_record/connection_handling.rb#L259
- https://github.com/rails/rails/blob/b752c38e81a310c1aaca78c7cdd1784009ea189a/activerecord/lib/active_record/connection_handling.rb#L293-L294

This will be soft deprecated in Rails 7.2.

The reason for this change is twofold.

1. We're running on Rails Edge with `active_record.permanent_connection_checkout = :deprecated` and this call to `#connection` was logged as a violation. Our goal is to configure this with `:disallowed` some day. Even if this needed to be a permanent checkout, `connection` should ideally be swapped with `lease_connection` (this would require a specific check to ensure the Rails dependency is >= 7.2).

2. Despite being called inside a `with_connection` block, this call actually leases a connection and doesn't return it back to the pool until the end of the request, which seems to have been intention behind #705. This ensures we use the connection temporarily leased by `with_connection`.

### DIFF
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -130,14 +130,14 @@ module Flipper
       end
 
       def get_all
-        with_connection(@feature_class) do
+        with_connection(@feature_class) do |connection|
           # query the gates from the db in a single query
           features = ::Arel::Table.new(@feature_class.table_name.to_sym)
           gates = ::Arel::Table.new(@gate_class.table_name.to_sym)
           rows_query = features.join(gates, ::Arel::Nodes::OuterJoin)
             .on(features[:key].eq(gates[:feature_key]))
             .project(features[:key].as('feature_key'), gates[:key], gates[:value])
-          gates = @feature_class.connection.select_rows(rows_query)
+          gates = connection.select_rows(rows_query)
 
           # group the gates by feature key
           grouped_gates = gates.inject({}) do |hash, (feature_key, key, value)|


jnunemaker:
This is awesome. Thanks for caring and submitting this! Sorry again for the delay in merging. It's been wild over here. I'll be better in the future. :smile:

jnunemaker:

joshuay03:
> If you don't mind, I would love to know what you are using flipper for as well and how it is working out for you (what went well, what was hard other than what you mentioned here).
> 
> If you aren't comfortable leaving that information here in a public comment, feel free to email me directly [john@fewerandfaster.com](mailto:john@fewerandfaster.com). I'm just always curious. :)

I opened this when I was working @TandaHQ. At the time we were trialing Flipper for incrementally rolling out new features and large page rewrites in Hotwire based on region and organisation user count. However, I wasn't very involved with product work and I'm not sure if/how it's currently being utilised. I'm sure either @ghiculescu or @AwolDes would be happy to give you some feedback.

AwolDes:
Thanks for the tag @joshuay03!

@jnunemaker I implemented Flipper a few months ago in our product. The goal was to leverage the groups feature to release features to cohorts like organisations with 100-250 staff, internal accounts, that kind of thing to reduce the blast radius of some of the larger releases we're doing. Flipper is wrapped by a class so we can use deadlines which then raise in `test` and `development` envs. 

CleanShot 2024-08-30 at 08 24 05

If you're keen to talk more I'll shoot you an email.

jnunemaker:
@AwolDes always keen to talk more! We've actually been thinking of adding something like deadlines to flipper.

PR #873, opened 2024-07-03
Load extra JS scripts into flipper-ui

nicastelo:
## Adds the ability to load extra JS scripts into flipper UI.

Added a new field `scripts` to `lib/flipper/ui/configuration.rb` so it allows to load scripts into the layout of flipper-ui.

### Usage:
```
  config.scripts << { src: "https://example.com/script.js" }
end
```

### DIFF
--- a/lib/flipper/ui/configuration.rb
+++ b/lib/flipper/ui/configuration.rb
@@ -77,6 +77,11 @@ module Flipper
       # Default is false.
       attr_accessor :confirm_disable
 
+      # Public: An array of scripts to include in the UI. Each script should be
+      # a hash with a `:src` key and optionally a `:integrity` key. Example:
+      # config.scripts << { src: "https://example.com/script.js", integrity: "sha384-abc123" }
+      attr_accessor :scripts
+
       VALID_BANNER_CLASS_VALUES = %w(
         danger
         dark
@@ -111,6 +116,7 @@ module Flipper
           { title: "Features", href: "features" },
           { title: "Settings", href: "settings" },
         ]
+        @scripts = []
       end
 
       def using_descriptions?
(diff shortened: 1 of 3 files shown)


jnunemaker:
@nicastelo hey! Looks good. I'm curious what the use case is. I'd love to know that prior to merging.

nicastelo:
> @nicastelo hey! Looks good. I'm curious what the use case is. I'd love to know that prior to merging.

Thanks!

In my case, I wanted to load some extra content into the page, like a list of actors to select from. Apart from that, this feature could be useful for adding custom styles or have some company branding, like a logo. It can even be useful for tracking stuff  on the frontend side if you wanted to.

Initially, I thought I could extend the ERB views but didn’t find an easy way to do so, apart from potentially monkey-patching and creating another set of views. With JS, I can simply load whatever content I want onto the page.

jnunemaker:
Ok, sorry for the delay. I'm back cleaning up flipper issues and re-read this. This PR looks great -- clean code and specs.

But I'm struggling with allowing random scripts. My concern is the scripts are out of my control yet I may get issues and have to debug them only to find out it was a custom script. It also means that the html becomes a bit of an "open" api. If I change the html then others scripts will fail. 

I think I'd prefer to make explicit configuration for each part people want to customize. 

For your list of actors, would it just be a select for that and that's all? Or would you also allow people putting in their own actors via text field too?

jnunemaker:

jnunemaker:
Hey @nicastelo, any opinions on this question:

> For your list of actors, would it just be a select for that and that's all? Or would you also allow people putting in their own actors via text field too?

nicastelo:
@jnunemaker I’d have it as a select and create field. It costs nothing to add an option for allowing or disallowing creation in the select field

nicastelo:
>If you don't mind, I would love to know what you are using flipper for as well and how it is working out for you (what went well, what was hard other than what you mentioned here).

We handle feature flags with it, allows us to turn on and off specific features we release, and even lets us pre-release stuff that's still not completed which helps us handing merges and sharing code a bit better.

Tracking flag changes is a bit of a mess, it occasionally happens that we lose track of what’s enabled or disabled in different environments. I saw you’re working on Flipper Cloud, which looks like it has a few features that could help with those issues. We also recently got an email from Sentry about a new feature they released for tracking feature flags. We might try both at some point

jnunemaker:
@nicastelo neat. Thanks! Cloud definitely solves all that. Happy to show you sometime or answer questions. It's super [easy to switch](https://www.flippercloud.io/docs/guides/migration#top) from OSS flipper.

PR #880, opened 2024-08-22
Add NamingStyle adapter to enforce format for added features.

bkeepers:
As suggested by @LandonSchropp in https://github.com/flippercloud/flipper/issues/879, this adds the ability to constrain feature names to a naming style.

```ruby
require 'flipper/adapters/naming_style'

Flipper.configure do |config|
  config.use Flipper::Adapters::NamingStyle, :snake # or :camel, :kebab, :screaming_snake, or a Regexp
end

# This will work because the feature key is in snake_case.
Flipper.enable(:snake_case)

begin
  # This will raise an error because the feature key is in CamelCase.
  Flipper.enable(:CamelCase)
rescue Flipper::Adapters::NamingStyle::InvalidFormat => e
  puts "#{e.class}: #{e.message}"
end
```

Output:
```
Flipper::Adapters::NamingStyle::InvalidFormat: Feature key "CamelCase" does not match format /^[a-z0-9]+(_[a-z0-9]+)*$/
```

### DIFF
--- a/spec/flipper/adapters/naming_style_spec.rb
+++ b/spec/flipper/adapters/naming_style_spec.rb
@@ -1,0 +1,70 @@
+require "flipper/adapters/naming_style"
+
+RSpec.describe Flipper::Adapters::NamingStyle do
+  it_should_behave_like "a flipper adapter" do
+    let(:format) { /.*/ }
+    let(:memory) { Flipper::Adapters::Memory.new }
+    let(:adapter) { described_class.new(memory, format) }
+
+    subject { adapter }
+
+    describe "#initialize" do
+      it "accepts a regex" do
+        expect { described_class.new(memory, format) }.not_to raise_error
+      end
+
+      it "accepts a symbol" do
+        [:camel, :snake, :kebab, :screaming_snake].each do |format|
+          expect { described_class.new(memory, format) }.not_to raise_error
+        end
+      end
+
+      it "raises an error if the format is an unknown symbol" do
+        expect { described_class.new(memory, :Pascal) }.to raise_error(ArgumentError)
+      end
+    end
+
+    describe "#add" do
+      {
+        /\A(breaker|feature)\// => {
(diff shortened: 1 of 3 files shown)


spec/flipper/adapters/naming_style_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        camel: {
+          valid: %w[Camel CamelCase SCREAMINGCamelCase CamelCase1 Camel1Case],
+          invalid: %w[snake_case Camel-Kebab lowercase],
+        },
+        snake: {
+          valid: %w[lower snake_case snake_case_1],
+          invalid: %w[CamelCase cobraCase double__underscore],
+        },
+        kebab: {
+          valid: %w[kebab kebab-case kebab-case-1 htt-party],
```
jnunemaker, line 42: I approve of these kebabs. 🎉

jnunemaker (approved):
This is sick. So clean. Thanks for whipping this together. I'm good as is. Random questions below: 

1. Any thoughts on if/how we should communicate the error raised from the flipper-ui? Should we rescue this error and add validation errors to the form or something? I could see someone being confused about an exception/error page with no details of why it is popping up. If we somehow rescue this error and explain maybe that would be enough? 
2. Any thoughts on what should happen in cloud sync when this error is hit? I think most of the cloud syncing avoids raising anything so I'm wondering if they'd not get feedback. 

I'm good with leaving as is for now but part of me wants to add this by default with snake case in the future.

LandonSchropp:
Awesome!

> Any thoughts on if/how we should communicate the error raised from the flipper-ui? Should we rescue this error and add validation errors to the form or something?

IMO, the best UX would be an inline error message that displays once the user has blurred the input. The second-best would be a banner/inline error if the user tries to submit the form. 😄

joaoserra91:
This feature would be great to ensure consistent naming across all feature flags, could this be revisited @bkeepers @jnunemaker ?

Issue #866, opened 2024-05-13
ActiveRecord adapter: unexpected error when running in a transaction


tjefferson08:
Calling `enable_actor` is not idempotent (from the application perspective)

1st time: 😄 
2nd time: 😡 

```ruby
test "enable_actor is not idempotent" do
  actor = create(:actor)

  run_enable_in_transaction = -> {
    ActiveRecord::Base.transaction do
      Flipper.enable_actor(:foo, actor)
    end
  }

  assert_nothing_raised { run_enable_in_transaction.call }

  assert_raises(ActiveRecord::StatementInvalid) { run_enable_in_transaction.call }
end
```

The observed effect:
Running flipper code in a transaction can unexpectedly "poison" the transaction

As I understand it, this is because the AR adapter is relying solely on DB constraints to enforce uniqueness, whereas most rails apps combine DB constraints for uniqueness with a corresponding validation on the AR model

jnunemaker:
Hi! Can you paste the full statement invalid error? We rescue `ActiveRecord::RecordNotUnique` ([source](https://github.com/flippercloud/flipper/blob/main/lib/flipper/adapters/active_record.rb#L256)), but maybe some other database is raising a statement invalid with a lower level unique error for some reason. 

What version of rails are you using and what database?

tjefferson08:
full error message:

`PG::InFailedSqlTransaction: ERROR:  current transaction is aborted, commands ignored until end of transaction block`

which makes sense to me, the INSERT statement from flipper fails, which aborts the transaction

I'm using rails `7.1.3.1` and Postgres version `14.11` but I suspect it may be reproducible on other databases/versions.

jnunemaker:
@tjefferson08 can you get the failure from flipper? Add a rescue around it and puts the error or something? I want to see what error is failing the transaction, because as I said above, we have a rescue for this so its idempotent, so it must be something different that is being raised that we aren't rescuing. Thanks!

tjefferson08:
>something different that is being raised that we aren't rescuing

I think the "something different" is the `ActiveRecord::StatementInvalid`

I would not expect rescuing `ActiveRecord::RecordNotUnique` to _also_ rescue `ActiveRecord::StatementInvalid`

I tried to get a failing test into the flipper test suite but I couldn't get any of the adapter-specific tests to run (even after following the docker-compose instructions)

jnunemaker:
[...]

PR #881, opened 2024-08-27
Active record poisoned transaction with postgres

jnunemaker:
The issue in #866 is tricky. For some reason when using the pg adapter, active record is poisoning the transaction on unique error even though that is specifically rescued. The error does not occur with sqlite or mysql. 🤷‍♂️

I asked copilot how to let AR and company know that the transaction is not poisoned and they suggested save points. And guess what... it worked! Haha. I don't love these two extra network calls but it does fix the issue for now. 

If anyone smarter or with more time can find a more slick solution or dig into why pg handles this different than sqlite/mysql I'm all ears. 

Initially I thought the problem was because of the connection check out stuff happening before the rescue but even moving the rescue inside with_connection block didn't fix the issue for pg.

refs #867 
fixes #activerecord #866

### DIFF
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -73,14 +73,15 @@ module Flipper
       # Public: Adds a feature to the set of known features.
       def add(feature)
         with_connection(@feature_class) do
-          # race condition, but add is only used by enable/disable which happen
-          # super rarely, so it shouldn't matter in practice
-          @feature_class.transaction do
-            unless @feature_class.where(key: feature.key).exists?
-              begin
+          @feature_class.transaction(requires_new: true) do
+            begin
+              # race condition, but add is only used by enable/disable which happen
+              # super rarely, so it shouldn't matter in practice
+              unless @feature_class.where(key: feature.key).exists?
                 @feature_class.create!(key: feature.key)
-              rescue ::ActiveRecord::RecordNotUnique
               end
+            rescue ::ActiveRecord::RecordNotUnique
+              # already added
             end
           end
         end
@@ -221,7 +222,7 @@ module Flipper
         raise VALUE_TO_TEXT_WARNING if json_feature && value_not_text?
 
         with_connection(@gate_class) do
-          @gate_class.transaction do
+          @gate_class.transaction(requires_new: true) do
             clear(feature) if clear_feature
             delete(feature, gate)
             begin
@@ -255,7 +256,7 @@ module Flipper
               end
             end
           rescue ::ActiveRecord::RecordNotUnique
-            # already added so no need move on with life
+            # already added so move on with life
           end
         end
 
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -1,4 +1,5 @@
 require 'set'
+require 'securerandom'
 require 'flipper'
 require 'active_record'
 
@@ -244,17 +245,26 @@ module Flipper
       end
 
       def enable_multi(feature, gate, thing)
-        with_connection(@gate_class) do
-          @gate_class.create! do |g|
-            g.feature_key = feature.key
-            g.key = gate.key
-            g.value = thing.value.to_s
+        with_connection(@gate_class) do |connection|
+          begin
+            connection.transaction do
+              savepoint_name = "flipper_savepoint_#{SecureRandom.hex(8)}"
+              connection.execute("SAVEPOINT #{savepoint_name}")
+              begin
+                @gate_class.create! do |g|
+                  g.feature_key = feature.key
+                  g.key = gate.key
+                  g.value = thing.value.to_s
+                end
+              rescue ::ActiveRecord::RecordNotUnique
+                # already added so no need move on with life
+                connection.execute("ROLLBACK TO SAVEPOINT #{savepoint_name}")
+              end
+            end
           end
         end
 
         nil
-      rescue ::ActiveRecord::RecordNotUnique
-        # already added so no need move on with life
       end
 
       def result_for_gates(feature, gates)
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -247,20 +247,15 @@ module Flipper
       def enable_multi(feature, gate, thing)
         with_connection(@gate_class) do |connection|
           begin
-            connection.transaction do
-              savepoint_name = "flipper_savepoint_#{SecureRandom.hex(8)}"
-              connection.execute("SAVEPOINT #{savepoint_name}")
-              begin
-                @gate_class.create! do |g|
-                  g.feature_key = feature.key
-                  g.key = gate.key
-                  g.value = thing.value.to_s
-                end
-              rescue ::ActiveRecord::RecordNotUnique
-                # already added so no need move on with life
-                connection.execute("ROLLBACK TO SAVEPOINT #{savepoint_name}")
+            connection.transaction(requires_new: true) do
+              @gate_class.create! do |g|
+                g.feature_key = feature.key
+                g.key = gate.key
+                g.value = thing.value.to_s
               end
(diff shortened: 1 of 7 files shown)


lib/flipper/adapters/active_record.rb
```diff
@@ -244,17 +245,26 @@ def delete(feature, gate)
...
-        with_connection(@gate_class) do
-          @gate_class.create! do |g|
-            g.feature_key = feature.key
-            g.key = gate.key
-            g.value = thing.value.to_s
+        with_connection(@gate_class) do |connection|
+          begin
+            connection.transaction do
+              savepoint_name = "flipper_savepoint_#{SecureRandom.hex(8)}"
+              connection.execute("SAVEPOINT #{savepoint_name}")
```
camallen, line 252: Noting that: pg transaction `savepoints` can cause performance degradation on postgres. 

Gitlab have blogged about this and how they remove them from their app https://about.gitlab.com/blog/2021/09/29/why-we-spent-the-last-month-eliminating-postgresql-subtransactions/

jnunemaker: @camallen yep, I'm aware. This is a very low throughput method. Even at a large company/app I'd be surprised if it was executed more than 100 times in a day. I think that level of throughput this should be fine. Knowing that, are you still worried? 

If anyone knows another way to fix it, I'm all ears but after hacking around on it for 30 minutes this was the first thing that worked.

byroot, line 252: You can use `transaction(require_new: true) do` https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/DatabaseStatements.html#method-i-transaction

byroot: And that's also probably why you are "poisoning" the connection. `conn.transaction { conn.transaction { } }` doesn't do nested transaction. You need `require_new: true` for that.

jnunemaker: Awesome! I'll try that out. Thanks for taking the time to teach!

jnunemaker: https://github.com/flippercloud/flipper/pull/881/commits/1c9e0a295d20de1b4fd58d94616651dbbdcb2b83 did the trick. 

Still odd to me that mysql and sqlite don't raise about the poisoned transaction and pg gem does. But perhaps it is something in the pg gem and not the AR adapter for Postgres? 

`requires_new: true` is definitely cleaner than the save point junk I had so I'll go with that for now.

camallen: > I think that level of throughput this should be fine. Knowing that, are you still worried?

I am still a little worried about this change, perhaps unduly.

From my reading the gitlab investigation showed that any savepoint use on the primary can impact all replicas as the replica has to record all active transactions after the use of a savepoint and this tracking is a problem during any 'long' running queries. On a very active system the duration of a 'long' query will vary as it depends on the number of active transactons. If there are enough they will saturate the pg server cache and spill from memory to disk and then impact query perf due to slower IO on hot code paths. 

My concern is around the the unpredictable nature of this issue due to the interplay of the active transaction count and the duration of any queries that impact the replicas. It'll appear like random performance degradation that (hopefully) eventually self heals after the queries complete. 

Teams that use replicas most likely are already experiencing a high transaction load on the primary and this change then might open them up to this situation.

One of the [proposed solutions](https://postgres.ai/blog/20210831-postgresql-subtransactions-considered-harmful#ideas-for-postgresql-development) (that may have landed in new pg version) was to allow a configurable PG server cache size to avoid spilling over to disk and slow IO on a hot code path. However by default that isn't available on older pg versions and may not be configurable on pg cloud variants. 

Again i may be reading it wrong, I'm wary as I've been bitten in the past from long running queries impacting replica query performance and the random nature of these issues.

> If anyone knows another way to fix it, I'm all ears but after hacking around on it for 30 minutes this was the first thing that worked.

ARs postgres adapter has a private [in_transaction?](https://github.com/rails/rails/blob/93e23dee7a2226e69bf44a5ba18bae445bd36399/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L899-L901) method to detect if we're in a transaction that uses `open_transactions`. We might be able to introspect on the connection adapter's [open_transactions](https://github.com/rails/rails/blob/93e23dee7a2226e69bf44a5ba18bae445bd36399/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb#L367) size to determine if we're in a transaction and if we need to use a savepoint sql directive instead of opting everyone into the use of savepoints by default. 

I'm not privy if there is a better way to introspect the AR connection transaction stack state to determine if we can 'opt' in to the use nested transactions. Maybe @byroot has thoughts.

byroot: > Maybe @byroot has thoughts.

I don't know the code, so hard. But it looks like it's trying to insert a record and ignore if the record already exist? If so then, `ON DUPLICATE KEY IGNORE` wouldn't require a sub transaction.

byroot: Ref to similar discussion when we used a savepoint in find_or_create https://github.com/rails/rails/issues/51052#issuecomment-1965659452.

jnunemaker: @camallen this isn't a hot code path and shouldn't ever be. 

If it is, then you aren't using flipper correctly. Individual actors is for enabling a feature for a few people here and there per feature. Anything that requires 100+ should be in a group or % rollout as detailed in the docs. 

Any write in flipper is a rare thing compared to reads and that is what flipper is optimized for.

camallen: > this isn't a hot code path and shouldn't ever be.

Apologies for the confusion, when I used the 'hot code path' term, I was referring to the PostgreSQL server when a `savepoint` is used on a primary and the replica then has to start tracking all know transactions. I wasn't referring to flipper configuring a feature flag, though flipper could be the trigger when it issues a `savepoint` transaction. 

I still think there is a risk with this change. If the GitLab investigation is correct then this change introduces a ticking bomb that may at some point explode for busy PostgreSQL systems that have replicas enabled. Similar to the concerns raised in https://github.com/rails/rails/issues/51052

Would you be open to a PR that attempts to use the upsert on conflict do nothing approach that byroot mentioned? I'd be up for working on this if it would be considered to land.

A naive implementation would run the model validations and if the model is valid it would then use ARs `insert` / `insert_all` [methods](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-insert_all) combined with :unique_by to skip the duplicate records and thus not poison the outer transaction. 

A caveat being that the `insert on conflict` feature is only available on [PG versions 9.5+](https://www.postgresql.org/docs/9.5/sql-insert.html) and active record [supports PG 9.3+ versions ](https://github.com/rails/rails/blob/0952b2a0a2a741dc6d1b873890d1b8f1210ee9cd/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L661-L662). It appears that flipper supports [active record 4.2+ ](https://github.com/flippercloud/flipper/blob/4b911dd3d185b5ba8d7f0be1323a5ba633565af1/flipper-active_record.gemspec#L35) so any implementation would require branching code based on the servers feature availability [detects server feature availability](https://github.com/rails/rails/blob/0952b2a0a2a741dc6d1b873890d1b8f1210ee9cd/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L268-L273). 

If the server doesn't support `ON CONFLICT DO NOTHING` SQL then the current `requires_new` savepoint would be used.

camallen: Looks like the configurable cache sizes have landed in PG 17 https://pganalyze.com/blog/5mins-postgres-17-configurable-slru-cache

jnunemaker: @camallen happy to take a look at a PR. Again, this is super low throughput so I don't believe it'll be an issue.

PR #887, opened 2024-09-28
Use reject(&:nil?) in enabled? check

stevecrozz:
Rejecting nil has the advantage of allowing use of null object patterns using objects that implement nil? but are not strictly nil. This also works well with the ['actor cannot be nil' guard clause](https://github.com/flippercloud/flipper/blob/main/lib/flipper/types/actor.rb#L12) which uses `nil?` to check nilness.

Just an idea. Would love to hear if there are other ideas. Could handle it differently further down by making Types::Actor.wrap(actor) return nil instead of throw an error. Not sure if that would have unintended affects.

### DIFF
--- a/lib/flipper/feature.rb
+++ b/lib/flipper/feature.rb
@@ -100,6 +100,8 @@ module Flipper
     #
     # Returns true if enabled, false if not.
     def enabled?(*actors)
+      # Allows null object pattern. See PR for more.
+      # https://github.com/flippercloud/flipper/pull/887
       actors = actors.flatten.reject(&:nil?).map { |actor| Types::Actor.wrap(actor) }
       actors = nil if actors.empty?
 
--- a/lib/flipper/feature.rb
+++ b/lib/flipper/feature.rb
@@ -100,7 +100,7 @@ module Flipper
     #
     # Returns true if enabled, false if not.
     def enabled?(*actors)
-      actors = actors.flatten.compact.map { |actor| Types::Actor.wrap(actor) }
+      actors = actors.flatten.reject(&:nil?).map { |actor| Types::Actor.wrap(actor) }
       actors = nil if actors.empty?
 
       # thing is left for backwards compatibility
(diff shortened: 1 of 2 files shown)


jnunemaker:
@stevecrozz hey! I have no thoughts. were you just pursuing code or did you hit a snag that required this?

stevecrozz:
> @stevecrozz hey! I have no thoughts. were you just pursuing code or did you hit a snag that required this?

I was inspired by an old RailsConf talk to try using a null object pattern https://www.youtube.com/watch?v=OMPfEXIlTVE

In my data model we have users that may or may not have a company. So without null object, `user.company == nil` but with null object, `user.company` becomes an instance of `Company::Null`. `Array#compact` won't reject instances of Company::Null.

Probably using compact to reject nils has better performance than `reject(&:nil?)` but using `reject(&:nil?)` allows you to implement `nil?` on your "null object" and it will work.

Ultimately, I don't think I'm going to go through with using this pattern so I'm not actually impacted, but maybe others are, so I wanted to drop this breadcrumb for anyone who might wind up with the same trouble.

I'd be perfectly happy closing this if there's little interest.

jnunemaker:
Maybe just add a code comment with link to this PR so I don't forget and switch it (or someone else)? I'd merge after that.

lib/flipper/feature.rb
```diff
@@ -100,7 +100,7 @@ def clear
     #
     # Returns true if enabled, false if not.
     def enabled?(*actors)
-      actors = actors.flatten.compact.map { |actor| Types::Actor.wrap(actor) }
+      actors = actors.flatten.reject(&:nil?).map { |actor| Types::Actor.wrap(actor) }
```
jnunemaker, line 103:
```suggestion
      # Allows null object pattern. See PR for more.
      # https://github.com/flippercloud/flipper/pull/887
      actors = actors.flatten.reject(&:nil?).map { |actor| Types::Actor.wrap(actor) }
```

stevecrozz:
Sure. I could probably even add a test case so we really don't forget :)

jnunemaker:
@stevecrozz haha yeah that's better.

stevecrozz:
@jnunemaker I added a test case but it looks like something is up with the Rails7+ scenarios. Looks like an existing issue.

PR #888, opened 2024-09-30
Avoid N+1 when loading Flipper UI Index

ericboehs:
Our project has many feature toggles (400+), when accessing the mounted Flipper UI app, it would take 20+ seconds for all the Feature look ups (M2 Max). On slower machines it takes minutes.

This two line change improves load speeds dramatically (22s -> 0.7s).


ericboehs:
Closing as I'm not able to reproduce with a test app.

PR #892, opened 2024-10-29
Sort available groups alphabetically in UI

ghiculescu:
A minor nuisance when enabling a feature for a group is that once you have quite a few groups, they show up in a random* order. IMO the UI would be easier to use if they showed alphabetically. 

| Before | After |
|-------|-------|
| image | image |

* technically they show in the order they are defined in your code.

I only changed this at the view layer, I wasn't sure if you want to sort groups alphabetically everywhere they're interacted with.


jnunemaker:
Seems like a great idea.

PR #896, opened 2024-11-08
Remove usage of OpenStruct

benedikt:
The `ostruct` gem will be a bundled gem in the future and needs to become a properly defined dependency because of that. There seems to be a general trend of just removing the `require "ostruct"` where possible.

It seems like this happened in one of the dependencies of Flipper, causing `OpenStruct` to no longer be available. As it was only used in a single place, I figured it's probably best to just remove its usage.


jnunemaker:
❤️

PR #910, opened 2025-02-19
Make redis and redis cache work with connection pool

jnunemaker:
This makes it so redis and redis cache work with a redis instance or a connection pool instance that generates redis instances.  

@cymen @j-castellanos @maleksiuk @BilalBudhani @tanelsuurhans @gstokkink Any thoughts on this? Seems like it'll just work. 

refs #826, #839, #835 

## To Use

All you need to do is configure your adapter:

```ruby
Flipper.configure do |config|
  config.adapter do
    pool = ConnectionPool.new(size: 5, timeout: 5) { Redis.new(...) }
    Flipper::Adapters::Redis.new(pool)
  end
end
```

If you're using redis cache, it's the same thing but pass connetion pool instance for cache instead of redis instance.


jnunemaker:
Yes. Trying to get a couple more fixes in now.

j-castellanos:
@jnunemaker great news! Keep us posted so we can update and then I can remove this code that I have been using in the meantime:
```
module Flipper
   module Adapters
     module Poolable
       def initialize(client_or_pool = nil, key_prefix: nil)
         @pool = nil
         @client = nil
         if client_or_pool.is_a?(ConnectionPool)
           @pool = client_or_pool
         else
           @client = client_or_pool
         end
         @key_prefix = key_prefix
       end

       def self.included(klass)
         klass.superclass.instance_methods(false).each do |method|
           klass.define_method method do |*args|
             return super(*args) unless @client.nil?

             @pool.with do |client|
               @client = client
               super(*args)
             ensure
               @client = nil
             end
           end
         end
       end
     end
   end
 end
```
Just out of curiosity, have people requested pooling for adapter other than Redis? If so you can probably just make clients for all adapters a pool with a default size of 1 to mimic current behavior.

maleksiuk:
@jnunemaker This looks great! I appreciate the straightforward approach and look forward to using it. Thanks for working on this.

BilalBudhani:
looks nice and clean. I have created a todo in our internally to use this adapter. I will report back once I have enough data. Thank you @jnunemaker!

gstokkink:
About to use this in production 😄

jnunemaker:
🔥 in the 🕳️ !!!

PR #921, opened 2025-03-31
[ruby34] Fix frozen string warning

tjallingvanderwal:
This change fixes the warning below, that appears when running Ruby 3.4.

```
.../flipper/lib/flipper/instrumentation/log_subscriber.rb:60:
warning: literal string will be frozen in the future
(run with --debug-frozen-string-literal for more information)
```


jnunemaker:
Fixed this over in #922 I think. Thanks for alerting!

PR #933, opened 2025-07-25
Add expression support to Flipper UI

blumhardts:
Hey there, apologies in advance for such a large PR. At Underdog, we had a feature flag targeting request come up that was solveable with expressions, but the stakeholders were non-technical and used to interacting with Flipper via the UI.  So, I took a stab at making expressions editable in Flipper UI.

I started with an approach that used partials rendered via a similar tree traversal manner as expression evaluation. I noticed there was hardly any JS in the gem, so I tried to stick with plain Ruby and HTML.

I wanted to make expression editing as consistent as possible with the rest of the gate forms, but found that to be a lot harder than I initially thought. The permissive nature of expressions requires supporting complicated form structures. The fact that an expression can't be added to a feature without enabling the feature for all actors that match that expression means I couldn't store "incomplete" expressions on the server without making changes to features themselves or adding a new data structure. It felt like my only other option would be to add complicated JavaScript for editing expressions on the client, so I threw in the towel and added an option to edit the expression as raw JSON instead.

Everything is gated behind the configuration value `Flipper::UI.configuration.expressions_enabled`, but this still probably isn't in a shippable state. I am happy to continue working on this and get it over the line if y'all find it valuable, but it's reached the point where I felt like feedback from the maintainers was the best way to keep it moving in the right direction (if at all).

### Screenshots

The features index, with two flags enabled for expressions rendering `gates_in_words`:
image

A feature with a simple property-based expression:
image

Expanding the expression editing form for an `All` expression:
image

Editing the same `All` expression as JSON:

image

[An unholy expression that no one would likely ever use](https://github.com/Underdog-Inc/flipper/blob/1378c3d6a4a390e4756288e4b1af77823bd68f37/spec/flipper/ui/actions/feature_spec.rb#L179), but demonstrates the power of expressions (you can see the form styling still needs some tweaks for this type of complexity):

image

### DIFF
--- a/lib/flipper/ui/expression_param_parser.rb
+++ b/lib/flipper/ui/expression_param_parser.rb
@@ -1,0 +1,40 @@
+module Flipper
+  module UI
+    # Internal: Used to parse expressions from the UI into a format that can be
+    # used by Flipper::Expression.build
+    class ExpressionParamParser
+      class InvalidJSONError < StandardError ;end
+
+      def initialize(expression)
+        @expression = expression
+      end
+
+      def parse
+        return {} unless @expression
+
+        begin
+          return JSON.parse(@expression) if @expression.is_a?(String)
+        rescue JSON::ParserError
+          raise InvalidJSONError
+        end
+
+        convert(@expression)
+      end
+
+      private
+
+      def convert(node)
+        return node unless node.is_a?(Hash)
+        return node unless node.key?('type') && node.key?('args')
+
+        type = node['type']
+        args = node['args']
+
+        args_array = []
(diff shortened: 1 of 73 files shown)


lib/flipper/ui/expression_param_parser.rb
```diff
@@ -0,0 +1,40 @@
+module Flipper
+  module UI
+    # Internal: Used to parse expressions from the UI into a format that can be
+    # used by Flipper::Expression.build
+    class ExpressionParamParser
```
blumhardts, line 5: I tried to get the request on the form submission as close to the expression AST as I could to reduce the amount of parsing. This was the best I could do

PR #935, opened 2025-07-30
Do not load ActiveRecord at require time

hugopeixoto:
The last version removed the on_load wrapper, making flipper load ActiveRecord immediately whenever flipper is required. This causes some issues with other gems that setup their own behavior using `on_load`.

This reintroduces the on_load wrapper around the database models, which by inheriting from ActiveRecord::Base were causing ActiveRecord to be loaded at require time.

Papertrail, for example, [loads its activerecord code in an `on_load` hook](https://github.com/paper-trail-gem/paper_trail/blob/master/lib/paper_trail/frameworks/rails/railtie.rb#L25-L27). If the flipper gem is loaded first, AR will be loaded and its hooks will run before papertrail has setup its hook, meaning that papertrail's code will never run.

`on_load` was removed in #924, in a commit that reads "Now that we are lazy loading the column check we don't need this stuff I believe", so I think that the removal was not necessary for any bug fixes or features. It had previously been added in https://github.com/flippercloud/flipper/pull/832.

### DIFF
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -8,33 +8,35 @@ module Flipper
     class ActiveRecord
       include ::Flipper::Adapter
 
-      class Model < ::ActiveRecord::Base
-        self.abstract_class = true
-      end
+      ActiveSupport.on_load(:active_record) do
+        class Model < ::ActiveRecord::Base
+          self.abstract_class = true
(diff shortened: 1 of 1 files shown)


jnunemaker:
Dang. I was hoping my delay of the column warning was enough. I'll merge this and cut a release.

PR #949, opened 2025-10-17
fix sync behavior when remote expression value is set to nil

jdnvn:
### Problem
When a cached/local feature has a non nil expression value and we attempt to sync with the remote, which has a nil expression value, the sync fails with the following error:

```
ArgumentError: nil cannot be converted into an expression (ArgumentError)
from /Users/jdonovan/src/flipper/lib/flipper/expression.rb:27:in 'Flipper::Expression.build'
```

This is because we are attempting to build an expression for the remote value (which is nil), to update our local value.

### Solution
Instead of building an expression, we should just let the local expression be updated to nil.

### DIFF
--- a/lib/flipper/adapters/sync/feature_synchronizer.rb
+++ b/lib/flipper/adapters/sync/feature_synchronizer.rb
@@ -53,7 +53,7 @@ module Flipper
         private
 
         def sync_expression
-          return if local_expression == remote_expression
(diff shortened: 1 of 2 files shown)


jnunemaker:
I merged this and then checked it locally and the spec still passed if i removed the expression nil check. What version of flipper were you on?

PR #952, opened 2025-11-13
Allow explicitly blocking actors and groups

Vicentecorrea:
Hello everyone! 👋

I want to present a possible solution to the issue https://github.com/flippercloud/flipper/issues/514: a good way of explicitly enabling a feature with exceptions.

_(It's just a draft PR to show my idea; it doesn't have any tests or documentation yet. If we decide to go ahead with this implementation, I'll do those stuff.)_

### Idea

My idea is to create blocking gates, and to check if a feature is enabled, there must be at least one open gate (as usual) **AND no blocking gates**.
- This implementation is backward compatible: is independent of 'disable' actions, and if you don't block things, Flipper keeps working just like always.
- No adapter changes are needed.

### Example

We have four users: two men Pepe and Juan, aged 15 and 79, and two women Mari and Cata, aged 18 and 22.
```ruby
user_struct = Struct.new(:age, :sex, :flipper_id)

pepe = user_struct.new(15, 'M', 'pepe')
mari = user_struct.new(18, 'F', 'mari')
cata = user_struct.new(22, 'F', 'cata')
juan = user_struct.new(79, 'M', 'juan')

@users = [pepe, mari, cata, juan]
```

In our `config/initializers/flipper.rb` we have this:
```ruby
Flipper.register(:teenagers) do |actor|
  actor.age.between?(13, 19)
end

Flipper.register(:women) do |actor|
  actor.sex == 'F'
end
```

We will enable and block actors and groups, and after each action, we will execute this method to check whether the `:test` feature is enabled or not for each user:
```ruby
def check_users
  @users.map do |user|
    Flipper.enabled?(:test, user)
  end
end
```

blocking_ff3 drawio

At time 0, the feature is not enabled for any user. Then, at time 1, we enable the feature for the `:teenagers` group, and now the `:test` feature is enabled for Pepe and Mari. And so on.

#### It also works after fully enable a feature

```ruby
Flipper.enable(:feat)
Flipper.block_actor(:feat, some_user)

Flipper.enabled?(:feat, some_user) # false
Flipper.enabled?(:feat, other_user) # true
```

### User interface

I imagine the UI to be something like this:
Captura de pantalla 2025-11-14 a la(s) 11 59 30


Vicentecorrea:
hi @jnunemaker! can you review this? 👀

Vicentecorrea:
> hi @jnunemaker! can you review this? 👀

or maybe @bkeepers?

Vicentecorrea:
Hi @jnunemaker! nice, thank you.

I tested this solution using the ActiveRecord adapter, and it works fine without any changes to it. As far as I know, it should also work fine with the other adapters without any necessary changes.

jnunemaker:
Oh I see how. I forgot I changed the gates hash to be used in gates. I wasn't seeing where you added the gates, but I forgot that gates is powered by gates_hash.

Vicentecorrea:
Yes, I added `blocking_actor` and `blocking_group` gates to the `gates_hash`, defining both gates using only existing classes but with the `blocking` attribute set to true.

Vicentecorrea:
Hi @jnunemaker, any updates?

lib/flipper/gates/actor.rb
```diff
@@ -3,11 +3,15 @@ module Gates
     class Actor < Gate
```
fsateler, line 3: I think this API is a bit awkward in that it would a bug to call `actor.open?` on a blocking actor or group. I think a better approach would be to create a BlockingActor that negates the open actor:

```ruby
class BlockingActor < Actor
  def open?(...)
    !super
  end
end
```

Would that be possible to implement?

regan-wrapbook:
👋 just adding a friendly bump to this PR, we would love this feature

PR #973, opened 2026-02-18
Add config to disable the Fully Enable button

elShiaLabeouf:
Hi!

I'm proposing to add two new configuration options for `Flipper::UI` to allow disabling the `Fully Enable` button.

### Use case

When rolling out features gradually (i.e. test group first, then wider), there are cases where fully enabling for everyone isn't appropriate - for example, when certain users have opted out of receiving new features. This configuration prevents accidental full rollouts.

### Demo

Default behavior (`config.fully_enable_disabled` is `false` by default):
Screenshot From 2026-02-18 11-16-57

Disabled button & default tooltip:
```
  config.fully_enable_disabled = true
end
```
Screenshot From 2026-02-18 11-15-57

Disabled button & custom tooltip:
```
  config.fully_enable_disabled = true
  config.fully_enable_disabled_with = '<Custom disabled message>.'
end
```
Screenshot From 2026-02-18 11-18-58

### DIFF
--- a/lib/flipper/ui/actions/boolean_gate.rb
+++ b/lib/flipper/ui/actions/boolean_gate.rb
@@ -16,6 +16,10 @@ module Flipper
           @feature = Decorators::Feature.new(feature)
 
           if params['action'] == 'Enable'
+            if Flipper::UI.configuration.fully_enable_disabled
+              status 403
+              halt view_response(:fully_enable_disabled)
+            end
             feature.enable
           else
             feature.disable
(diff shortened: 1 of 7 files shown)


jnunemaker:
I'm not sure I understand why. Why would you not eventually fully enable a feature?

elShiaLabeouf:
Hey @jnunemaker, thanks for the quick response! :)

Our project is actively developing, the feature set is expanding, the design is changing; and while most of the users are happy with such active support, a small percentage of them would prefer to keep the product minimalistic. I would very much like to fulfill their wishes since it's easy - we'll just leverage a Flipper group "everyone except these guys". At the same time, I want to minimize the chance of someone accidentally fully enabling a FF for everyone - that's where disabling the button would come in handy.

Please share your thoughts on the proposal! Maybe there's a better way to achieve this goal?

PR #994, opened 2026-04-23
Add disabled? as syntax sugar for !enabled?

loqimean:
## Motivation

When checking whether a feature is **off**, you currently have to write:

```ruby
if !Flipper.enabled?(:my_feature)
  # ...
end
```

or the slightly more idiomatic but still awkward:

```ruby
unless Flipper.enabled?(:my_feature)
  # ...
end
```

Both read backwards — you're asking "is it enabled?" when you really want to express "is it disabled?". This gets noisy in call sites that deal primarily with the disabled state, and the negation adds cognitive overhead on every read.

## Change

Adds `disabled?` as the inverse of `enabled?` on `Feature`, `DSL`, and the top-level `Flipper` module:

```ruby
# Before
if !Flipper.enabled?(:my_feature, current_user)
  redirect_to upgrade_path
end

# After
if Flipper.disabled?(:my_feature, current_user)
  redirect_to upgrade_path
end
```

`disabled?` accepts the same actor arguments as `enabled?` and simply returns `!enabled?(*actors)`, so all existing gate semantics are preserved.

## Test plan

- [x] `Feature#disabled?` specs covering single actor, multiple actors, and enabled/disabled states
- [x] `Flipper.disabled?` delegation spec
- [x] Full suite passes (`bundle exec rspec spec/flipper/feature_spec.rb spec/flipper_spec.rb` — 205 examples, 0 failures)

### DIFF
--- a/lib/flipper.rb
+++ b/lib/flipper.rb
@@ -56,7 +56,7 @@ module Flipper
   # Public: All the methods delegated to instance. These should match the
   # interface of Flipper::DSL.
   def_delegators :instance,
-                 :enabled?, :enable, :disable,
+                 :enabled?, :disabled?, :enable, :disable,
                  :enable_expression, :disable_expression,
                  :expression, :add_expression, :remove_expression,
                  :enable_actor, :disable_actor,
--- a/lib/flipper/dsl.rb
+++ b/lib/flipper/dsl.rb
@@ -36,6 +36,16 @@ module Flipper
       feature(name).enabled?(*args)
     end
 
+    # Public: Check if a feature is disabled.
+    #
+    # name - The String or Symbol name of the feature.
+    # args - The args passed through to the disabled check.
+    #
+    # Returns true if feature is disabled, false if not.
+    def disabled?(name, *args)
+      feature(name).disabled?(*args)
+    end
+
     # Public: Enable a feature.
     #
     # name - The String or Symbol name of the feature.
--- a/lib/flipper/feature.rb
+++ b/lib/flipper/feature.rb
@@ -127,6 +127,13 @@ module Flipper
       end
     end
 
+    # Public: Check if a feature is disabled for zero or more actors.
+    #
+    # Returns true if disabled, false if not.
+    def disabled?(*actors)
+      !enabled?(*actors)
+    end
+
     # Public: Enables an expression_to_add for a feature.
     #
     # expression - an Expression or Hash that can be converted to an expression.
--- a/spec/flipper/feature_spec.rb
+++ b/spec/flipper/feature_spec.rb
@@ -98,6 +98,47 @@ RSpec.describe Flipper::Feature do
     end
   end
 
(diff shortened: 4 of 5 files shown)


jnunemaker:
Hi, appreciate the PR but this is intentional. I talked about it in the past on similar PRs. https://github.com/flippercloud/flipper/pull/94#issuecomment-162917454

Flipper is only about allowing people in, not blocking people but disabled? gives the vibe that someone can be blocked.

loqimean:
> Hi, appreciate the PR but this is intentional. I talked about it in the past on similar PRs. [#94 (comment)](https://github.com/flippercloud/flipper/pull/94#issuecomment-162917454)
> 
> Flipper is only about allowing people in, not blocking people but disabled? gives the vibe that someone can be blocked.

Hey, thanks for the context and linking the previous discussion — I did find #94 after looking more carefully, so I can see this has come up a few times!

I totally get the concern about the mental model. The framing of "flipper only opens gates, never closes them" makes sense architecturally. But I'd argue disabled? doesn't necessarily imply that individual actors can be blocked — it's just a readability convenience for the inverse check, same way Ruby's zero? doesn't mean you set something to zero.

The practical case I keep running into is negative conditionals in real product code. Something like:
```ruby
if !Flipper.enabled?(:gdpr_consent_collected, user)
  show_consent_wall
elsif !Flipper.enabled?(:age_verified, user)
  show_age_gate
else
  show_full_content
end
```
You can't really reach for unless here, and stacking ! negations in elsif chains is considered a code smell in a lot of teams (including mine, where unless is also discouraged). The result is that developers end up writing harder-to-read code not because of their logic, but because the gem nudges them that way.

The fact that this has surfaced in at least 3 independent PRs over the years suggests it's a genuine friction point, not just a personal preference. Maybe a note in the docs that disabled? is intentionally absent and here's why would at least save future contributors the confusion?

Either way, appreciate you taking the time to explain the reasoning!

jnunemaker:
The only thing I can think of is not_enabled? which feels so clunky but still holds the line of the intent/meaning. I tried claude and that is the best they came up with as well.

jnunemaker:
this is also important because i'm working on permit/deny which will actually be checked before enabled? so adding disabled? would be even more confusing in a few weeks. :)

loqimean:
> The only thing I can think of is not_enabled? which feels so clunky but still holds the line of the intent/meaning. I tried claude and that is the best they came up with as well. 

Okay, got it 😄. So, what if I rename it disabled? to 'not_enabled?'? At least that's better than nothing 😆

loqimean:
@jnunemaker friendly reminder

jnunemaker:
Sorry this fell through the cracks. 

Hmm... not_enabled? feels un-pretty. I don't have a better or more logical response than that. Once I get deny and some other things in I'll do some brainstorming and see if I can get over it or come up with something better. Still open to other options but something about it just doesn't feel right.

PR #1031, opened 2026-07-24
Build ActiveRecord gate results in a single pass

connorshea:
AI Disclosure: This was generated using Claude Code, w/ Fable 5. It has been reviewed and tested by me, and I have validated the findings and confirmed their correctness, including by having Claude run a battery of tests to protect against regressions. The PR description was written mostly by Claude (excluding this intro).

`result_for_gates` re-scanned the full list of gate rows once per gate (up to 6 passes), calling `to_sym` on every row key each pass. Features with many enabled actors can have thousands of rows (though hopefully no one does this...), making `get`, `get_multi`, and `get_all` `O(rows * gates)`.

Bucket rows by gate key in one pass instead: `O(rows + gates)` with one `to_sym` per row.

Benchmarks (sqlite3 in-memory, macOS on an M5 Pro, Ruby 3.4.9, arm64):

  | Scenario | Before (i/s) | After (i/s) | Speedup | Before (allocs/call) | After (allocs/call) |
  |---|---:|---:|---:|---:|---:|
  | enabled? - small feature (1 boolean row) | 26,831 | 27,853 | +4% | 137 | 124 |
  | enabled? - feature with 2000 enabled actors | 543 | 819 | +51% | 12,630 | 12,371 |
  | preload_all - 10x200-actor features + the 2000-actor one | 335 | 539 | +61% | 20,570 | 20,409 |

Allocation counts barely move on the big scenarios because they are dominated by `pluck` materializing rows, but the CPU savings from eliminating the repeated scans are the main thing that helps here.

### DIFF
--- a/lib/flipper/adapters/active_record.rb
+++ b/lib/flipper/adapters/active_record.rb
@@ -239,21 +239,28 @@ module Flipper
       end
 
       def result_for_gates(feature, gates)
+        # Bucket rows by gate key in one pass instead of re-scanning all rows
+        # for each of the feature's gates. Features with many enabled actors
+        # can have thousands of rows. Rows with a nil key come from the outer
+        # join in get_all for features with no gate values.
+        values_by_key = nil
+        gates&.each do |key, value|
+          next if key.nil?
+          values_by_key ||= {}
+          (values_by_key[key.to_sym] ||= []) << value
+        end
+
         result = {}
-        gates ||= []
         feature.gates.each do |gate|
+          values = values_by_key && values_by_key[gate.key]
           result[gate.key] =
             case gate.data_type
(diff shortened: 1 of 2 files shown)


connorshea:
Also for context on why I'm opening these PRs in case it's useful: I've been investigating performance improvements in certain Sidekiq jobs at work, and Flipper was a large source of memory allocs/DB queries. This was mostly because some of our code is doing Bad Things and is checking flags far more often than it needs to, and also because there's no memoization for Flipper in Sidekiq by default. The latter problem I have fixed in our Rails app. So I wanted to see if there was anything that would be easy to improve in the Flipper gem using Claude that might help things out :)

PR #1033, opened 2026-07-24
Normalize feature names by removing invisible characters only

jnunemaker:
Follow-up to #1027 (thanks @koddsson for the report and fix!).

That PR fixed the real problem — pasting a feature name with an invisible character (zero-width space, word joiner, etc.) creates a flag that looks identical in the UI but never matches the name in code. However, transliterating to ASCII had two side effects:

- **Silent renames of visible characters**: creating `café` in the UI produced a `cafe` flag, which no longer matches code checking `Flipper.enabled?(:café)` — the same mismatch bug the PR was fixing, just with visible characters.
- **Non-latin names rejected**: `新機能` (Japanese, Cyrillic, Hebrew, etc.) normalized to empty string and hit the invalid-name error, so those names could no longer be created via the UI at all.

This changes the normalization to remove only what you can't see, leaving every visible character untouched:

- Strips unicode format and control characters (`\p{Cf}` / `\p{Cc}`: zero-width spaces, joiners, BOM, soft hyphens)
- Trims unicode whitespace from both ends (non-breaking space, ideographic space — which `String#strip` misses)
- Scrubs invalid UTF-8 bytes
- Names that normalize to empty still fall through to the existing invalid-name error

Since dropped characters are invisible, the result always matches what the user intended to type.

The logic now lives in core as `Flipper::Typecast.to_feature_name` so it can be shared, and is also applied to `POST /features` in flipper-api, which had the same gap (with the same fallthrough: names that normalize to empty return the existing 422 `name_invalid` error instead of creating an empty-key feature).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/flipper/api/v1/actions/features.rb
+++ b/lib/flipper/api/v1/actions/features.rb
@@ -38,7 +38,10 @@ module Flipper
           end
 
           def post
-            feature_name = params.fetch('name') { json_error_response(:name_invalid) }
+            feature_name = Typecast.to_feature_name(
+              params.fetch('name') { json_error_response(:name_invalid) }
+            )
+            json_error_response(:name_invalid) if feature_name.empty?
             feature = flipper[feature_name]
             feature.add
             decorated_feature = Decorators::Feature.new(feature)
--- a/lib/flipper/typecast.rb
+++ b/lib/flipper/typecast.rb
@@ -73,6 +73,20 @@ module Flipper
       end
     end
 
+    # Internal: Convert value to a feature name usable as a feature key.
+    # Removes invisible format/control characters (zero-width spaces,
+    # joiners, BOM, etc.) that make visually identical but distinct names,
+    # and trims unicode whitespace (e.g. non-breaking space) that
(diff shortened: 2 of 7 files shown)


connorshea:
FWIW this missed a few other invisible characters that should potentially also be stripped:

```
U+3164 HANGUL FILLER         kept=true  -> "search\u3164"       distinct_from_'search'=true
U+115F CHOSEONG FILLER       kept=true  -> "search\u115F"       distinct_from_'search'=true
U+1160 JUNGSEONG FILLER      kept=true  -> "search\u1160"       distinct_from_'search'=true
U+FFA0 HW HANGUL FILLER      kept=true  -> "search\uFFA0"       distinct_from_'search'=true
U+2800 BRAILLE BLANK         kept=true  -> "search\u2800"       distinct_from_'search'=true
U+17B4 KHMER VOWEL AQ        kept=true  -> "search\u17B4"       distinct_from_'search'=true
```

jnunemaker:
@connorshea fixed https://github.com/flippercloud/flipper/pull/1043

